PR #67, opened 2023-10-05
Gsw 387 docker image

harryoh:
create docker image for gnoswap contract deployment


.github/workflows/publish-contract-deploy.yml
```diff
@@ -0,0 +1,42 @@
+
+name: Public gnowatp contract deploy
```
r3v4s, line 2: name: Public gnoswap contract deploy

r3v4s (changes requested):

r3v4s (approved):

PR #80, opened 2023-10-31
GSW-516 feat: call grc 20 function with registered interface

r3v4s:
to user function from grc20 token without import statement,
- each token's function must be registered
- to test, added third token (baz)

# BREAKING CHANGE
- every grc20 function call should be called via registered interface
- if not, it must be test case or blunder


notJoon (approved):
LGTM 👍

PR #81, opened 2023-11-06
GSW-530 feat: swap router

r3v4s:
- `BestSwap()` handles all type of swap (exactIn / exactOut), (singlePath, multiPath), (singlePool, multiPool)
- `BestSwapDry()` will return best route path depends on input paramter
- tokenPrice now will be updated with base token `gno.land/r/foo`

# BREAKING CHANGE
- `pool.Swap()` will be deprecated next version, call `router.BestSwap()`


router/find_path.gno
```diff
@@ -0,0 +1,115 @@
...
+				if strings.HasPrefix(fourth, outputTokenPath) {
+					// three times
+					nestedPath := inputTokenPath + "," + secondFee + "," + secondPath + "," + thirdFee + "," + thirdPath + "," + fourthFee + "," + fourthPath
+					swapPaths[len(swapPaths)] = nestedPath
+				}
+			}
+		}
+	}
+	return swapPaths
+}
```
notJoon, line 115: I think we can combine `findTwicePath` and `findThreeTimesPath` into one. There are two ways to do this off the top of my head. One is the parameterize how many times to traverse and then "run it as a recursive function", and the other is to use a "graph algorithm".

The first one is easy to fix and does not involve many changes. For example, we could change like this

```go
func findPath(
    tokenPairs TokenPairs,
    currentTokenPath string,
    outputTokenPath string,
    currentPath string,
    remainingHops int,
    swapPaths *SwapPaths,
) {
    if remainingHops == 0 {
        if strings.HasPrefix(currentTokenPath, outputTokenPath) {
            *swapPaths = append(*swapPaths, currentPath)
        }
        return
    }

    for _, next := range tokenPairs[currentTokenPath] {
        nextPath, nextFee := singlePoolPathWithFeeDivide(next)
        newPath := currentPath
        if currentPath != "" {
            newPath += ","
        }
        newPath += nextFee + "," + nextPath
        findPath(tokenPairs, nextPath, outputTokenPath, newPath, remainingHops-1, swapPaths)
    }
}
```
However, there is a potential for overflow, so you'll need to set a separate maximum value. I prefer the second method because it doesn't improve time complexity much.

You can think of a graph with token pairs as edges and tokens as nodes. Also, if I understand correctly, the fee information in the pool can be used as a weight.

If you use the pool as a weight information, we can use an algorithm like [Dijkstra](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm) or [Bellman-Ford](https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm), but I think that simply using an algorithm like BFS can significantly reduce the time complexity from $O(n^2)$ to $O(V + E)$.

However, implementing a graph algorithm is quite cumbersome, so if you don't mind, you can modify it later.

router/quotation.gno
```diff
@@ -0,0 +1,184 @@
...
+	// it contains outputRatio with 0 which is impossible path
+	// > remove path with 0 ratio
+	var finalTargets []QuoterTarget
+	for _, quote := range quoterTargets {
+		if quote.resultRatioX96 != 0 {
+			finalTargets = append(finalTargets, quote)
+		}
+	}
+
+	sort.Sort(ByOutputRatioDesc(finalTargets))
```
notJoon, line 113: It's not critical, but it would be nice to only sort if the slice is non-empty.

router/quotation.gno
```diff
@@ -0,0 +1,184 @@
...
+			break
+		}
+	}
+
+	return bestSwaps
+}
+
+func removeDuplication(bestSwaps []QuoterTarget) (finalSwaps []QuoterTarget) {
+	finalSwaps = make([]QuoterTarget, 0) // return
+	insertedPath := []string{}           // tmp
```
notJoon, line 150: Here again, I think we could improve the complexity by using a map to check for duplicates, so that when we find the targetPath. By doing so, we can do it in one time. like:

```go
pathIndexMap := make(map[string]int)
```

notJoon (changes requested):
Great work 👍. I've touched on a few things, but just take it as a guide, and please point out any errors.

router/getter_api.gno
```diff
@@ -0,0 +1,291 @@
...
+		fee := poolDatas[5]
+		if token0Path < token1Path {
+			return token0Path + ":" + token1Path + ":" + fee
+		} else {
+			return token1Path + ":" + token0Path + ":" + fee
+		}
+	default:
+		panic(ufmt.Sprintf("[ROUTER] getter_api.gno__makePoolPath() || unknown pool index: %d", poolIndex))
+	}
+}
```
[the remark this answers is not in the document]
r3v4s: I think `baseIndex := poolIndex * 3` should be '2', not '3'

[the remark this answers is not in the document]
r3v4s: I'll commit suggestion anyway, but if test-case fails will set to '2'

router/quotation.gno
r3v4s: lets sort when len finalTagets > 0

router/find_path.gno
```diff
@@ -0,0 +1,115 @@
...
+
+	// check if there is path that starts with input
+	require(len(tokenPairs[inputTokenPath]) != 0, ufmt.Sprintf("[ROUTER] find_path.gno__getSwapPaths() || len(tokenPairs[inputTokenPath]) == 0, inputTokenPath: %s", inputTokenPath))
+
+	// find direct path
+	for _, output := range tokenPairs[inputTokenPath] {
+		if strings.HasPrefix(output, outputTokenPath) {
+			outputPath, outputFee := singlePoolPathWithFeeDivide(output)
+			directPath := inputTokenPath + "," + outputFee + "," + outputPath
+			swapPaths[len(swapPaths)] = directPath
```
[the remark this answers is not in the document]
r3v4s: swapPaths is a type of `map[int]string`, so `len(swapPaths)` doesn't really panic.

tried this and result is `panic: should not happen`

router/find_path.gno
r3v4s: applied cb16f6a9b977b1bead8de7855f4f819141ee85cd

r3v4s:
@notJoon bit of err found which I left in comments, and made function in getter_api recursive too as quotation

notJoon (approved):

r3v4s:
@mconcat branch is merged for further PR, but pls leave your comments :D

PR #83, opened 2023-11-07
GSW-357 feat: change external incentive distribution period 

r3v4s:
- external reward distribution has changed to 90 / 180 / 360 days
- external token now calls with registered interface (to support multi rewards)


notJoon (approved):
LGTM 👍

PR #103, opened 2023-11-29
refator: `router`

notJoon:
WIP: The refactoring is done. But I can't run the test for now, so I'll have to check it later.

For public function, I modified it to throw a panic. But for the fucntions that are not need to be expoesed, ~modified to return an error type.~

- UPDATE 12/04: rollback to return string type as an error


r3v4s:
could you make it draft pr if it isn't ready yet

notJoon:
> @notJoon Do we need left work for this? or is it ready for review?

I'll wrap up today

router/router.gno
```diff
@@ -69,62 +69,129 @@ func SwapRoute(
...
 		quote, _ := strconv.Atoi(quotes[i])
 
-		// check if route hop is up to 3
-		require(numHops <= 3, "[ROUTER] router.gno__SwapRoute() || numHops > 3")
+		if hops > 3 {
+			return errorf(kind, numHopsMustInRange)
+		}
 
-		toSwap := amountSpecified / bigint(100) * bigint(quote)
+		toSwap := amountSpecified / bigInt(100) * bigInt(quote)
```
r3v4s, line 159: don't change type.
`bigInt` doesn't exist, it should be `bigint`

r3v4s (changes requested):
check comment

PR #133, opened 2023-12-15
Refactor getSwapPaths function to use strings.Builder for better performance

notJoon:
Fixed using `strings.Builder` when combining strings, instead of the old `+`.


r3v4s (approved):
code looks good, approving for now.

but please do not merge yet until we finish our demo :D

PR #138, opened 2023-12-17
GSW-712 fix: staker reward calculation

r3v4s:
# Fixed
- only subtract already claimed value when CollectReward() called, not APIReward()

## ETC
- bug was discovered by @jinoosss


r3v4s:
if every thing is ok, just approve ( DO NOT MERGE YET )

notJoon (dismissed):

notJoon (approved):

PR #150, opened 2024-02-05
GSW-838 feat add query metadata in every api function

r3v4s:
# New
- every function for api call should include query metadata
- apply camelCase for json

# Change
- every dry* functions are defined using its own file

# Fix
- test case


notJoon (approved):

jinoosss:
The function to get a single pool has been removed..

r3v4s:
> @r3v4s , The function to get a single pool has been removed.. It was changed to an internal function, is there a reason?

we(@r3v4s @notJoon @mconcat) decided to remove every api related functions except the ones necessary for inserting data to database.

dongwon8247:
@r3v4s Re-implement this feature. We need this because it's better to provide real-time data via RPC when providing liquidity.

dongwon8247:
> we(@r3v4s @notJoon @mconcat) decided to remove every api related functions except the ones necessary for inserting data to database.

@r3v4s Also next time, provide a detailed explanation why it was decided that way and the benefits of doing it. Otherwise It's hard to follow up and understand.

PR #160, opened 2024-02-14
refactor: position/liquidity_amounts

notJoon:
## Description

1. Elimination of Duplicate Code: Implemented a logic to automatically swap the values of `sqrtRatioA` and `sqrtRatioB` using the `swapIfGreater` function without duplication.

2. Function Name Change: Functions that are only used in the `liquidity_amounts.gno` file have been renamed to `computeXXX`, while functions called from other places have been prefixed with `GetXXX`."

### DIFF
--- a/position/liquidity_amounts.gno
+++ b/position/liquidity_amounts.gno
@@ -16,38 +16,38 @@ func toAscendingOrder(a, b bigint) (bigint, bigint) {
 
 // calcIntermediateValue computes the intermediate value
 // used in liquidity calculations.
-func calcIntermediateValue(sqrtRatioA, sqrtRatioB bigint) bigint {
-	return (sqrtRatioA * sqrtRatioB) / consts.Q96
+func calcIntermediateValue(sqrtRatioAX96, sqrtRatioBX96 bigint) bigint {
+	return (sqrtRatioAX96 * sqrtRatioBX96) / consts.Q96
 }
 
-// GetLiquidityForAmount0 calculates liquidity for a given amount of token 0.
-func computeLiquidityForAmount0(sqrtRatioA, sqrtRatioB, amount0 bigint) bigint {
-	sqrtRatioA, sqrtRatioB = toAscendingOrder(sqrtRatioA, sqrtRatioB)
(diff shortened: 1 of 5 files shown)


position/liquidity_amounts.gno
```diff
@@ -1,49 +1,53 @@
...
-	sqrtRatioBX96 bigint,
-	amount0 bigint,
-) bigint {
-	if sqrtRatioAX96 > sqrtRatioBX96 {
-		sqrtRatioAX96, sqrtRatioBX96 = sqrtRatioBX96, sqrtRatioAX96
+import (
+	"gno.land/r/demo/consts"
+)
+
+// swapIfGreater checkes if the first value is greater than
```
r3v4s, line 7: `swapIfGreator` sounds like it `swap` for token swap, not variable swap.
any better word that we cna use?

r3v4s (approved):
1 comment for variable name convension

position/liquidity_amounts.gno
```diff
@@ -1,49 +1,53 @@
...
-	sqrtRatioX96 bigint,
-	sqrtRatioAX96 bigint,
-	sqrtRatioBX96 bigint,
-	amount0 bigint,
-	amount1 bigint,
-) bigint {
-	var liquidity bigint
+// computeLiquidityForAmount1 calculates liquidity for a given amount of token 1.
+func computeLiquidityForAmount1(sqrtRatioA, sqrtRatioB, amount1 bigint) bigint {
+	sqrtRatioA, sqrtRatioB = toAscendingOrder(sqrtRatioA, sqrtRatioB)
```
mconcat, line 35: I think the suffix `X96` is meaningful because it tells us the bigint is actually a 96-bit fixed point integers, but I agree that they should not appear in the variable names. A better way would be defining a new X96 wrapper type so we can enforce the decimal computation at the type level. Once we adapt u256 types we can also work on it.

mconcat (dismissed):
Great! The Get/compute convention here is a great way to demonstrate golang coding style and easier to read. Left a minor comment on `X96` suffixes, could be handled in a future PR.

r3v4s (approved):

mconcat (approved):

PR #161, opened 2024-02-14
refactor: pool/tick_math.gno

notJoon:
## Description

1. Extract tick ratio values into map (`tickRatioMap`)
2. Extracted the individual logic of the `TickMathGetTickAtSqrtRatio` function into functions.

PR #162, opened 2024-02-14
GSW-867 feat common calculation logic to separate package

r3v4s:
# New
- tick_math, liquidity_amounts calculation logic is being used by multiple realms => make it as `common` package

TODO: wait for #160 #161 to merge, then change inner code


mconcat (approved):
Removing redundant functions across packages is a great improvement, thanks for the refactoring!

notJoon:
It's definitely became more concise, thanks for the work 👍

notJoon (dismissed):

r3v4s:
@notJoon @mconcat Had to force push to fix conflicts. Please take a deep look a commit cf010d148e2e8507ab8acfe7e45b4de9d14947c1

common/liquidity_amounts.gno
```diff
@@ -9,12 +13,12 @@ func liquidityAmountsGetLiquidityForAmount0(
...
 	}
 
-	intermediate := (sqrtRatioAX96 * sqrtRatioBX96) / Q96
+	intermediate := (sqrtRatioAX96 * sqrtRatioBX96) / consts.Q96
 
 	return amount0 * intermediate / (sqrtRatioBX96 - sqrtRatioAX96)
 }
 
-func liquidityAmountsGetLiquidityForAmount1(
+func LiquidityAmountsGetLiquidityForAmount1(
```
notJoon, line 21: I think we'll have to rename it something better later.  just like truncate the `LiquidityAmount` prefix 🤔

README.md
```diff
@@ -41,7 +41,8 @@ To run unit tests, follow these steps:
 2. Copy the `gnoswap` contracts into the Gno core:
 
    ```
-   $ cp -R _setup/* gov pool position router staker $WORKDIR/gno-for-swap/examples/gno.land/r/
+   $ cp -R _setup/* consts gov pool position router staker $WORKDIR/gno-for-swap/examples/gno.land/r/demo/
```
notJoon, line 44: 👍

staker/reward_math.gno
```diff
@@ -74,7 +82,7 @@ func rewardMathComputeInternalRewardAmount(tokenId uint64, deposit Deposit) bigi
 	rewardRatio := getRewardRatio(stakedDuration)
 
 	finalRewardQ := myWholeRewardX96 * bigint(rewardRatio)
-	finalReward := finalRewardQ / Q96 / Q96 / Q96 / Q96 / Q96 / 100 / 100
+	finalReward := finalRewardQ / consts.Q96 / consts.Q96 / consts.Q96 / consts.Q96 / consts.Q96 / 100 / 100
```
notJoon, line 85: I'm not familiar with the specific context or behind story of this code, but it appears that replacing `100 / 100` with `1000` might be a straightforward adjustment. 

Additionally, converting `blockRewardInit` to a `bigint` type seems feasible for use in this calculation, but I don't know if that variable can be used in a formula.

notJoon (dismissed):
LGTM. I don't see any particular issues

common/liquidity_amounts.gno
r3v4s: once we refactor our logic and if it works fantastic both on unit tests + integeration tests, then we can have time to rename things (including package path to deploy)

staker/reward_math.gno
r3v4s: >  but it appears that replacing 100 / 100 with 1000 might be a straightforward adjustment.

Actually, it goes same with bunch of `Q96` division. Right below that line there are some comments about why we need this kind of division formula.
https://github.com/gnoswap-labs/gnoswap/blob/64fc5ac2b64fec64e66af6bdcf96a0204b0aa105/staker/reward_math.gno#L87-L93
I kind wanted to comments to be similar as formula and this is exact reason why I didn't use `1000`. However I'm fine with your suggestion (I'll leave it you :D)

r3v4s: > Additionally, converting blockRewardInit to a bigint type 
Suggesting, wait until @mconcat to finish his uint256 then we can do something about it. How about that?

mconcat (dismissed):

notJoon (approved):

mconcat (approved):

PR #168, opened 2024-02-16
Remove unnecessary checks from `liquidityMathAddDelta` function

notJoon:
## Description

When the result(`z`) became negative is generally due to the incorrect input values. For example, if `x` is negative and `y` is a positive number that is less than the absolute value of that negative number (e.g., x=-100, y=50). Here, `z` turn out to be negative value(-50). Another case is when both input values are negative.

So I think it's sufficient to simply check that `z` is negative.

### DIFF
--- a/pool/liquidity_math.gno
+++ b/pool/liquidity_math.gno
@@ -3,15 +3,11 @@ package pool
 import "gno.land/p/demo/ufmt"
 
 func liquidityMathAddDelta(x bigint, y bigint) bigint {
-	if y < 0 {
-		z := x - (-y)
-		requireUnsigned(z, ufmt.Sprintf("[POOL] liquidity_math.gno__liquidityMathAddDelta() #1 || expected z(%d) >= 0", z))
-		require(z < x, ufmt.Sprintf("[POOL] liquidity_math.gno__liquidityMathAddDelta() || expected z(%d) < x(%d)", z, x))
-		return z
-	} else {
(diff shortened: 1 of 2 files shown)


r3v4s (approved):
code is getter better :D

pool/liquidity_math.gno
```diff
@@ -3,15 +3,11 @@ package pool
...
-		requireUnsigned(z, ufmt.Sprintf("[POOL] liquidity_math.gno__liquidityMathAddDelta() #1 || expected z(%d) >= 0", z))
-		require(z < x, ufmt.Sprintf("[POOL] liquidity_math.gno__liquidityMathAddDelta() || expected z(%d) < x(%d)", z, x))
-		return z
-	} else {
-		z := x + y
-		requireUnsigned(z, ufmt.Sprintf("[POOL] liquidity_math.gno__liquidityMathAddDelta() #2 || expected z(%d) >= 0", z))
-		require(z >= x, ufmt.Sprintf("[POOL] liquidity_math.gno__liquidityMathAddDelta() || expected z(%d) >= x(%d)", z, x))
-		return z
-	}
+    z := x + y
```
mconcat, line 6: In the Solidity version of this code, x and y have type of uint and int respectively, which required the negation and applying the seemingly redundant operation. Commenting for the context.

mconcat (approved):

PR #172, opened 2024-02-20
refactor(staker): Optimize the deleteFromDeposits Function

notJoon:
## Description

The previous `deleteFromDeposits` function implemented delete in the `deleteFromIncentives` function by creating a new map and copying the elements from the old map to the new map. 

However, this resulted in unnecessary memory allocation, so I modified it to remove directly from the map using the built-in function `delete` instead of creating a new map. Apply the same method to `deleteFromDeposits` function as well.

### DIFF
--- a/staker/staker.gno
+++ b/staker/staker.gno
@@ -286,7 +286,7 @@ func getTokenPairBalanceFromPosition(tokenId uint64) (bigint, bigint) {
 	lowerX96 := common.TickMathGetSqrtRatioAtTick(pn.PositionGetPositionTickLower(tokenId))
 	upperX96 := common.TickMathGetSqrtRatioAtTick(pn.PositionGetPositionTickUpper(tokenId))
 
-	token0Balance, token1Balance := common.LiquidityAmountsGetAmountsForLiquidity(
+	token0Balance, token1Balance := common.GetAmountsForLiquidity(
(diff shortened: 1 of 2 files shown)


staker/staker.gno
```diff
@@ -128,8 +128,8 @@ func CollectReward(
 	tokenId uint64, // LP TokenID
 ) {
 	deposit, exist := deposits[tokenId]
-	require(exist, ufmt.Sprintf("[STAKER] staker.gno__CollectReward() || tokenId(%d) not staked", tokenId))
 
+	require(exist, ufmt.Sprintf("[STAKER] staker.gno__CollectReward() || tokenId(%d) not staked", tokenId))
```
r3v4s, line 132: what's up with line changes ??

r3v4s (changes requested):
If I remember correctly, deleting item in map via `delete` doesn't work when you actually deploy realm then delete via `gnokey maketx call` while it works great on test cases.

Could you verify `delete` behavior in gno's staging and our gs/base_clean branch ?

staker/staker.gno
notJoon: > what's up with line changes ??

I ran the `gofumpt -s -w` commad in this file, so might be affected.

PR #175, opened 2024-02-22
GSW-792 feat do not burn nft even entire liquidity is removed

r3v4s:
# New
1. even position is clear(no liquidity, no fee left), do not burn nft
> its for interface to show closed position

## NO MERGE YET
wait for #173  #174


pool/_GET_no_receiver.gno
```diff
@@ -0,0 +1,250 @@
+package pool
+
+import (
+	"gno.land/p/demo/ufmt"
+)
+
+// Slot0
+func PoolGetSlot0SqrtPriceX96(poolPath string) bigint {
+	pool, exist := pools[poolPath]
+	requireExist(exist, ufmt.Sprintf("[POOL] _GET_pool.gno__PoolGetSlot0SqrtPriceX96() || pool(%s) does not exist", poolPath))
```
notJoon, line 10: I'm not sure how often the pool getter functions will be called in a row, but if a function checks a `Pool`, storing the result as `boolean` could avoid unnecessary checks. What do you think? (Of course, since `pools` is `map` type, it probably doesn't matter much)

notJoon (dismissed):

pool/_GET_no_receiver.gno
r3v4s: Most of getter functions without receiver is meant be called from sync progress to make API datas, so for now I'cant tell you whether pool getter will be called often or not.

And for the function checking... if we're only talking about pool's getter I like your idea. But in the grand scheme I want to save result as `boolean` since those pattern can be applied to others

notJoon (approved):

PR #176, opened 2024-02-22
GSW-875 feat use compute route logic from uniswap for router to calculate ratio

r3v4s:
# New
1. router use computeRoute logic from uniswap to calculate ratio

## NO MERGE YET
wait for #173 #174 #175


router/_RPC_api.gno
```diff
@@ -47,109 +47,105 @@ func ApiGetRatiosFromBase() string {
...
-	}
-	return result
+	return calculateTokenRatio(currentToken, routes, proceed, priceX96)
 }
 
 func getTokenList() []string {
 	seen := make(map[string]bool)
 	uniqueTokenList := []string{}
-	poolList := p.PoolGetPoolList()
+	poolList := pl.PoolGetPoolList()
```
notJoon, line 148: We don't know the final size of the `uniqueTokenList`, but using the size of the `poolList` seems to allow us to set the initial capacity and this will reduce the re-allocations.

```diff
+ poolList := pl.PoolGetPoolList()
+ uniqueTokenList := make([]string, 0, len(poolList)*2)
- uniqueTokenList := []string{}
- poolList := pl.PoolGetPoolList()
```

r3v4s: looks great, but why does initial length requires `* 2`?

notJoon (approved):

router/_RPC_api.gno
notJoon: > looks great, but why does initial length requires `* 2`?

I simply set it to double because I assumed that since each pool has two tokens (`token0`, `token1`), each pool would add up to two unique tokens to the list. However, since this value was not set after exact measurement, we will need to reevaluate this part later.

PR #186, opened 2024-02-28
GSW-881: Refactor pool swap function (2)

notJoon:
## Description

Refactored the main logic of the Swap function.

Create a mockup type (MockPool) for testing purpose.

Added newXXX functions to create instances of SwapCache and SwapState.

### ETC

Reopen a PR(#182) that was unexpectedly closed.


r3v4s:
Looks Good, however could you divide `transferTokenAndVerify` in two different handlers for `Transfer` and `TransferFrom` ?
> These two functions are crucial to gnoswap, and I think it would be better if we can determine which type of transfer is used without looking into another helpers

notJoon:
> Looks Good, however could you divide `transferTokenAndVerify` in two different handlers for `Transfer` and `TransferFrom` ?

Now I split that function into `transfer`, `transferFrom` and `verifyTransfer` methods.

r3v4s:
Closing due to duplciates #203

PR #188, opened 2024-03-06
GSW-846: feat: uint256, int256 instead of bigint

r3v4s:
# New
1. uint256 (REF: https://github.com/holiman/uint256)
2. int256 (REF: https://github.com/mempooler/int256)

~# WIP~
~u256 * u256 overflow handling~

co worker: @notJoon @mconcat


pool/_RPC_api.gno
```diff
@@ -21,7 +21,7 @@ type RpcPool struct {
 	BalancesToken1 string `json:"balanceToken1"`
 
 	// fee is the fee tier of the pool
-	Fee uint16 `json:"fee"`
+	Fee uint32 `json:"fee"`
```
mconcat, line 24: What is the reason behind using uint32 for fees?

pool/_RPC_dry.gno
```diff
@@ -4,42 +4,59 @@ import (
...
-	pFee uint16,
+	fee uint32,
 	recipient std.Address,
 	zeroForOne bool,
-	amountSpecified_ bigint,
-	sqrtPriceLimitX96_ bigint,
-) (bigint, bigint, bool) {
+	_amountSpecified string, // int256
+	_sqrtPriceLimitX96 string, // uint160
+) (string, string, bool) { // int256 x2
```
mconcat, line 23: I think this a generally a good practice to only use primitive types for public APIs, even if gno vm has no problem with dealing with pointers. :+1:

pool/_TEST_pool_native_swap_test.gn_OK
```diff
@@ -37,19 +36,19 @@ func TestMint(t *testing.T) {
...
+	testBanker.RemoveCoin(std.GetOrigCaller(), "ugnot", 804_032)
 
 	// Deposit(wrap)
 	std.TestSetPrevAddr(test1)
 	wugnot.Deposit()
 
 	std.TestSetPrevAddr(test1)
-	foo.Approve(a2u(consts.POOL_ADDR), 295801)
-	wugnot.Approve(a2u(consts.POOL_ADDR), 804031)
+	foo.Approve(a2u(consts.POOL_ADDR), 295802)
```
mconcat, line 50: Are these differences in the approvals due to some rounding errors?

packages/big/int256/int256.gno
```diff
@@ -0,0 +1,557 @@
...
+
+// Mul sets z to the product x*y and returns z.
+func (z *Int) Mul(x, y *Int) *Int {
+	z.abs = z.abs.Mul(x.abs, y.abs)
+	z.neg = x.neg != y.neg // 0 has no sign
+	return z
+}
+
+// Rsh sets z = x >> n and returns z.
+func (z *Int) Rsh(x *Int, n uint) *Int {
```
mconcat, line 179: In EVM, int256s are represented by signed 256bit integers, whereas int256.Int is using 257bit one. I am curious if shifting operation(and others too) might create any possible out-of-256bit range values. We can either add a range checking logic on operations, or we can first merge this and see if there is any bugs happening on the testnet.

mconcat (approved):
Thanks for making all these work, including int256 and nice integrations! Few comments are left like some questions, but I think this one is good to be merged.

pool/_RPC_api.gno
r3v4s: Position contract uses uint32 as pool's fee data type.
However, it can be still `uint16` I think

pool/_RPC_dry.gno
```diff
@@ -176,43 +195,39 @@ func DrySwap(
...
-			return 0, 0, false
+			return "0", "0", false
 		}
 	}
 
-	// just not enough balance
-	if amount0.IsZero() && amount1.IsZero() {
-		return 0, 0, false
+	// JUST NOT ENOUGH BALANCE
+	if amount0.IsZero() || amount1.IsZero() {
```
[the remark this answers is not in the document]
r3v4s: It's for fail safe when calculation returns 0 for input or output token amount.
Certain test cases barely gets here so I think we need to kee pit

pool/_TEST_pool_native_swap_test.gn_OK
r3v4s: `bigint` really didn't round up anything. As matter of fact it didn't support rounding.
However with u256, i256 we can round up things.

packages/big/int256/int256.gno
r3v4s: I prefer latter. AFAIK solidity <= 0.5.0 doesn't care about overflow/underflow and that's why openzeppelin provide SafeMath

r3v4s:
> Thanks for making all these work, including int256 and nice integrations! Few comments are left like some questions, but I think this one is good to be merged.

Replied all of your comments, and resolved if it was just comment.
If all of replies looks good, please resolve conversation.

notJoon (approved):

pool/_RPC_api.gno
r3v4s: keeping `uint32` since most of other realm(position, staker) uses uint32

PR #189, opened 2024-03-07
GSW-907 remove requires

r3v4s:
# New
1. remove every `require`
> just use `if ~ panic` just like go does

# Fix
2. bit of number change that didn't include last pr


position/sqrt_price_math.gno
```diff
@@ -80,6 +80,34 @@ func sqrtPriceMathGetNextSqrtPriceFromAmount1RoundingDown(
...
 	}
+
+	// quotient mutst be positive when amount and liquidity are positive
+	if add {
+		if amount.Lte(u256.UnsafeFromDecimal(consts.MAX_UINT160)) {
+			quotient = new(u256.Uint).Lsh(amount, 96)
+			quotient = new(u256.Uint).Div(quotient, liquidity)
+		} else {
+			quotient = u256.MulDiv(amount, u256.UnsafeFromDecimal(consts.Q96), liquidity)
+		}
```
notJoon, line 91: It would be better to extract the part that calculates the quotient into a function. Also, since both the if and else branches return some value, the code in the else block could be pulled out.

r3v4s: Oh yes, if my memory serves me right `sqrt_price_math` was handling it with separate function when we're using `bigint`.

But for now we uses `u256/i256` and I had to roll back in old ways if test cases are find

router/_RPC_api.gno
```diff
@@ -71,7 +71,7 @@ func findRatios(maxHops int) []TokenRatio {
 		} else {
 			numRoutes := uint64(len(routes))
 
-			_tokenRatioX96 := u256.Zero()
+			var _tokenRatioX96 bigint
```
notJoon, line 74: After this line, `_tokenRatioX96` is infered as `u256.Uint` type, but I was wondering if there is any specific reason for using bigint?

r3v4s: Blunder on my side.
If you don't mind you're welcome to change it or leave PR RFC then I'll handle it.

router/router.gno
```diff
@@ -46,19 +46,25 @@ func SwapRoute(
 
 	// check route length ( should be 1 ~ 7 )
 	routes := strings.Split(strRouteArr, ",")
-	require(1 <= len(routes) && len(routes) <= 7, ufmt.Sprintf("[ROUTER] router.gno__SwapRoute() || len(routes) should be 1 ~ 7 (len(routes):%d)", len(routes)))
+	if !(1 <= len(routes) && len(routes) <= 7) {
+		panic(ufmt.Sprintf("[ROUTER] router.gno__SwapRoute() || len(routes) should be 1 ~ 7 (len(routes)[%d])", len(routes)))
+	}
```
notJoon, line 51: If the conditional expression is long, I prefer to assign it to a variable. In this case, using a name like `isValidRouteLength` could improve readability.

```go
isValidRouteLength := 1 <= len(routes) && len(routes) <= 7
if !isValidRouteLength {
    // ...
}
```

notJoon (approved):
LGTM. I left some notes to reference for future works.

router/router.gno
r3v4s: https://github.com/gnoswap-labs/gnoswap/blob/e26ff5d08b36b76aefcb3aaf98489f36e6d05eef/router/router.gno#L51-L54

router/_RPC_api.gno
r3v4s: https://github.com/gnoswap-labs/gnoswap/blob/e26ff5d08b36b76aefcb3aaf98489f36e6d05eef/router/_RPC_api.gno#L72-L75

position/sqrt_price_math.gno
r3v4s: https://github.com/gnoswap-labs/gnoswap/pull/195/commits/7ef5241f1d962daf3125008f9206a442d37b2adf

PR #192, opened 2024-03-13
GSW-900 oneclick

r3v4s:
# New
1. Automatically wrap/unwrap native token for Mint, Burn, Swap
2. Automatically stake lpTokenId from mint

# Fix
1. wugnot banker permission
2. test cases


r3v4s:
#194 fixes conflict

PR #198, opened 2024-03-28
GSW-980 test: apply txtar instead of makefile

r3v4s:
# New
1. remove scenario test cases, and apply new txtar testing

# Fix
2. `unexpected object id`

# Change
3. test grc20 admin address

# Refactor
4. apply refactored uint256/int256 
5. folder structure


staker/_TEST_staker_rpc_get_test.gnoa
```diff
@@ -35,18 +35,18 @@ func init() {
...
 }
 
-func TestPoolInitCreatePool(t *testing.T) {
-	std.TestSetPrevAddr(gsa)
-	pl.InitManual()
-	std.TestSkipHeights(1)
-
+func TestPoolCreatePool(t *testing.T) {
 	std.TestSetPrevAddr(test1)
+	for i := 0; i < (5 * 1); i++ {
```
notJoon, line 40: Is there any reason that using `5 * 1` instead of `5`? 🤔🤔🤔

notJoon (approved):
left some question. But LGTM

staker/_TEST_staker_rpc_get_test.gnoa
r3v4s: Basically it uses `5 x {num_of_pool_to_create}` template for every tests. It would be better to keep its format

PR #206, opened 2024-04-09
fix: change internal reward fee tier

r3v4s:
# Hotfix
1. change default (internal incentive) pool tier from 0.3% to 0.05%


dongwon8247 (commented):
@r3v4s Why change it to 0.05%? The 0.3% pool is the default pool for internal incentives.

r3v4s:
OMG.. blunder on me, confused with very old code.

PR #217, opened 2024-05-15
GSW-1037 apply emit event in contract

r3v4s:
## New
1. apply emit & event for pool, position, router, staker
2. test cases to support both unit test and file test

## Update
3. remove Mint, Faucet for all GRC20 tokens
4. no pool creation fee
5. update minDeposit of ugnot to wrap

### FYI
ignore git action failing, it should be run fine when we use gnolang/gno binary, not our own gno binary


notJoon (approved):

notJoon:
LGTM. But weired CI failing occured 

```plain
panic: name Emit not declared [recovered]
	panic: gno.land/r/demo/pool/pool.gno:428: name Emit not declared:
```

PR #247, opened 2024-07-02
GSW-1142 feat: gns (mint with permission) + pre-calculated halving amount

r3v4s:
## feat new gns

1. mint with permission
-> only `emission` contract can call

2. mint with pre-calculated halving amount
2.1 every single year, we (have plan to) mint different number of gns
2.2 from yearly amount, calculate daily amount > then calculate single block amount
2.3 with `lastMintedHeight` variable, calculate amount of gns to mint to `emission`
--> which makes us we don't have to mint actually every block

### DIFF
--- a/_deploy/r/demo/gnoswap/consts_halving/consts.gno
+++ b/_deploy/r/demo/gnoswap/consts_halving/consts.gno
@@ -1,0 +1,95 @@
+package consts
+
+import (
+	"std"
+)
+
+// GNOSWAP SERVICE
+const (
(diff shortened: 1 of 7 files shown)


notJoon (approved):

dongwon8247:
I'm not sure if having the GNS token contract inside the GnoSwap contracts is better than having it outside of the contracts. I find most of the prominent DEXes have its tokens outside of their core contracts, like Uniswap, [Ekubo](https://github.com/EkuboProtocol/token) etc. What is the benefit of each?

mconcat (approved):

mconcat:
We may need to change the name of the gns_halving to something more canonical, as it can confuse the users. TBD

PR #251, opened 2024-07-06
GSW-1149 calculate pool position reward then transfer

r3v4s:
## feat
1. calculate each pool's gns emission amount
2. calculate each pool's -> position's gns emission amount
3. calculate each position's -> each incentive external reward amount


staker/reward_math.gno
```diff
@@ -259,33 +142,60 @@ func getNumPoolTiers() (uint64, uint64, uint64) {
...
+// 	case timestamp >= 5: // 5s ~ 5d (5s is block creation time)
+// 		ratio = 30
+// 		break
+// 	default: // ~ 4s
+// 		ratio = 0
+// 		break
+// 	}
+
+// 	return ratio
+// }
```
notJoon, line 197:
```suggestion
```

notJoon (approved):

staker/reward_math.gno
r3v4s: Still need to keep this.
> if multi node in gno returns same timestamp we can use this function

PR #259, opened 2024-07-08
GSW-1248 feat staker external incentive rpc return with left amount

r3v4s:
## feat
1. staker external incentive rpc return with how much uToken left


notJoon (approved):

notJoon:
LGTM 👍

PR #293, opened 2024-07-29
GSW-1339 feat: router contract set sqrt price limit

r3v4s:
## new
1. when router.SwapRoute calls pool.Swap set min or max sqrt price limit to proper range based on fee


notJoon (approved):

AkydA (approved):

dongwon8247:
Let's get @jinoosss's review on this before merging it

router/swap_inner.gno
```diff
@@ -118,3 +121,33 @@ func i256Max(x, y *i256.Int) *i256.Int {
...
+	case 500:
+		return 887270
+	case 3000:
+		return 887220
+	case 10000:
+		return 887200
+	default:
+		panic(ufmt.Sprintf("[ROUTER] router.gno__getMaxTick() || unknown fee(%d)", fee))
+	}
+}
```
jinoosss, line 153: `gno.land/r/gnoswap/consts` It looks like MAX_TICK is declared in the package here.
What do you think about managing the value of MIN / MAX TICK per fee tier ?

jinoosss (approved):
I I have one question.

PR #295, opened 2024-07-29
GSW-1344 fix: apply warm up period logic to external incentives

r3v4s:
## feat
1. warm up period logic should appleid to external incentive too

### FYI
- DO NOT MERGE ( ONLY REVIEW PLEASE )
- internal(GNS emission) incentive temporary disabled to debug external only


staker/calculate_pool_position_reward.gno
```diff
@@ -235,96 +227,156 @@ func CalcPoolPosition() {
...
+	keys := make(map[string]bool)
+	list := []string{}
+	for _, entry := range strSlice {
+		if _, value := keys[entry]; !value {
+			keys[entry] = true
+			list = append(list, entry)
+		}
+	}
+	return list
 }
```
notJoon, line 382:
```suggestion
func removeDupeStr(strSlice []string) []string {
    seen := make(map[string]struct{}, len(strSlice))
    result := make([]string, 0, len(strSlice))
    
    for _, str := range strSlice {
        if _, ok := seen[str]; !ok {
            seen[str] = struct{}{}
            result = append(result, str)
        }
    }
    
    return result
}
```

Modified it to be faster and less memory allocations. Below are the benchmark results comparing the original implementation and the suggested function.

```benchmark
BenchmarkRemoveDupeStrOriginal-8   	   45552	     26543 ns/op	    2126 B/op	       8 allocs/op
BenchmarkRemoveDupeStrImproved-8   	  100945	     12028 ns/op	   57368 B/op	       3 allocs/op
PASS
```

notJoon (commented):
LGTM. Left some small suggestion, but you don't have to apply this 👍

AkydA (approved):

notJoon (approved):

PR #306, opened 2024-08-06
GSW-1387 fix external remaining amount

r3v4s:
## fix
1. when calling pre-calculated external reward data from on-chain via abci_query, each external incentive's rewardLeft should be updated too.


notJoon (approved):
👍

AkydA (approved):

PR #311, opened 2024-08-08
GSW-1399 fix do not use existing wugnot balance when swapping ugnot

r3v4s:
## fix
1. do not use pre-exising wugnot balance when swapping ugnot
- selecting ugnot => use only ugnot coin
- selecting wugnot => use only wugnot token


notJoon:
test runs well 👍

notJoon (approved):

AkydA (approved):

PR #314, opened 2024-08-08
[WIP] feat: gnoswap run test action

notJoon:
# Description

Make an test action based on the `setup.py`. But I don't know if it works

## TODO

- install gno key


notJoon:
not working

PR #325, opened 2024-08-20
GSW-1488 test: each external duration with 4 different staked position

r3v4s:
## test
1. 3 test case for each external incentive duration (90, 180, 365)
- single pool has 4 position
- all positions are in-range
- all positions are staked
- each position has different liquditiy


notJoon (approved):
In `TestJustPrint`, the test fails because the value of the `time` field is different from the expected, but everything else is fine.

r3v4s:
> In `TestJustPrint`, the test fails because the value of the `time` field is different from the expected, but everything else is fine.

You need to use gno binary that skips 2s for single block, by default its 5s

AkydA (approved):

PR #329, opened 2024-08-23
chore: Remove boolean conversion

notJoon:
# Description

Since the gno's boolean value comparison has been fixed, removed the boolean conversions.

note: this must be merge after merged [#2725](https://github.com/gnolang/gno/pull/2725). So, until then this PR will drafted. So, I'll leave it in draft until then.


r3v4s:
Let's keep this draft for now.
We need to see if next test4 is going to use latest branch or not.

r3v4s:
@notJoon I think it can be closed, all of type conversion has been fixed 
https://github.com/gnoswap-labs/gnoswap/pull/349#discussion_r1798660614

PR #333, opened 2024-09-02
GSW-1553 refactor: divide staker.GetNecessary() data to return by number of functions

r3v4s:
## refactor
1. divide `staker.GetNecessary()` data to be return by different functions


r3v4s:
dups with #334

PR #335, opened 2024-09-05
GSW-1575 feat: proposal quorum

r3v4s:
## feat
1. proposal quorum should be 50% of xgns total supply at the time of its creation
2. add orig caller, prevRealm in event to to emit
3. to create proposal, proposer needs to have at least 1_000 xGNS (1_000_000_000u)

## fix
4. testcase
5. undelegation should decrease amount from very first

### DIFF
--- a/gov/governance/_RPC_api_proposal.gno
+++ b/gov/governance/_RPC_api_proposal.gno
@@ -257,6 +257,7 @@ func getProposalVotes(id uint64) string {
 
 	max := gs.GetTotalDelegatedUntil(weightUntilTimestamp)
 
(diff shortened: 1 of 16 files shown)


onlyhyde (approved):
I suggest that we need to support to check total value of account have xgns and minimum xgns value.

when one account have some xgns, he can create propsal. but only one account can also vote.

So, we need to discuss it.

notJoon (dismissed):

notJoon (approved):

onlyhyde (approved):
LGTM

PR #337, opened 2024-09-09
GSW-1579 feat: proposer can propose several proposals

r3v4s:
## feat
1. same address can propose several proposals


notJoon (approved):

onlyhyde (approved):
LGTM

PR #339, opened 2024-09-11
GSW-1610 fix: collect xGns reward when undelegation

r3v4s:
## fix
1. when undelegation happens, xgns reward should be collected automatically

## chore
2. emit config version when Reconfigure triggered


gov/staker/staker.gno
```diff
@@ -67,6 +67,7 @@ func Delegate(to std.Address, amount uint64) {
 func Redelegate(from, to std.Address, amount uint64) {
 	en.MintAndDistributeGns()
 	calculateReward()
+	CollectReward()
```
onlyhyde, line 70: When CollectReward function is called, calculateReward function is also called.
So, i suggest that calculateReward function should remove when CollectReward function is called to avoid duplicated function call

onlyhyde (changes requested):

gov/staker/staker.gno
r3v4s: b56b63a950d6c6d9059f582a2bb722305cdb57e2

notJoon (dismissed):

onlyhyde (approved):

notJoon (dismissed):

notJoon (approved):

onlyhyde (approved):

PR #342, opened 2024-09-12
GSW-1620 feat: support multiple execution with single proposal

r3v4s:
## feat
1. single param change proposal can change multiple contract's param.


gov/governance/parameter_change.gno
```diff
@@ -0,0 +1,371 @@
...
+
+func handleSingleBool(params []string) bool {
+	if params[0] == "true" {
+		return true
+	} else if params[0] == "false" {
+		return false
+	} else {
+		panic(ufmt.Sprintf("invalid bool(%s)", params[0]))
+	}
+}
```
notJoon, line 371: minor suggestion

```suggestion
func handleSingleBool(params []string) bool {
	switch params[0] {
	case "true":
		return true
	case "false":
		return false
	default:
		panic(ufmt.Sprintf("invalid bool(%s)", params[0]))
	}
}
```

gov/governance/governance_proposal.gno
```diff
@@ -174,6 +174,37 @@ func ProposeParameterChange(
...
+
+		var pkgPath, funcName, params string
+		switch len(splitExe) {
+		case 3:
+			pkgPath = splitExe[0]
+			funcName = splitExe[1]
+			params = splitExe[2]
+		case 2:
+			pkgPath = splitExe[0]
+			funcName = splitExe[1]
```
notJoon, line 199: It seems like it would be good to manage the extracted results as structures. I'll check on this later

notJoon (approved):

onlyhyde (approved):

notJoon (approved):

PR #359, opened 2024-10-21
refactor: remove test value

r3v4s:
## refactor
1. removed value for testing with testcase and on-chain tests.
> each testcase overwrites from testing logic
> on-chain test value will be overwrite when we actually deploy those.

### DIFF
--- a/gov/governance/config.gno
+++ b/gov/governance/config.gno
@@ -18,13 +18,15 @@ var (
 
 func init() {
 	// XXX: TODO, REMOVE AFTER QA ( smaller value for testing )
-	VotingStartDelay:              uint64(180),           // 3m
-	VotingPeriod:                  uint64(600),           // 10m
-	VotingWeightSmoothingDuration: uint64(600),           // 10m
-	Quorum:                        uint64(50),            // 50% of total xGNS supply
-	ProposalCreationThreshold:     uint64(1_000_000_000), // 1_000_000_000
-	ExecutionDelay:                uint64(60),            // 1m
-	ExecutionWindow:               uint64(3600),          // 1h
+	config = Config{
+		VotingStartDelay:              uint64(180),           // 3m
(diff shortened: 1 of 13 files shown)


notJoon (dismissed):

gov/governance/config.gno
```diff
@@ -17,40 +17,17 @@ var (
...
-		// XXX: TODO, REMOVE AFTER QA
-		// - config for live qa
-		// VotingStartDelay:              uint64(180),           // 3m
-		// VotingPeriod:                  uint64(600),           // 10m
-		// VotingWeightSmoothingDuration: uint64(600),           // 10m
-		// Quorum:                        uint64(50),            // 50% of total xGNS supply
-		// ProposalCreationThreshold:     uint64(1_000_000_000), // 1_000_000_000
-		// ExecutionDelay:                uint64(60),            // 1m
-		// ExecutionWindow:               uint64(3600),          // 1h
+		VotingStartDelay:              uint64(86400),         // 1d
```
onlyhyde, line 22: as i know, dongwon have request to use governanve config for QA. need to check it.

onlyhyde (changes requested):

gov/governance/config.gno
dongwon8247: yes, let's set all configs with QA time for now, and when we're done with the final QA before `beta v3`, we can roll all back.

r3v4s: FYI, current dev chain deployed with qa timing

r3v4s: 9bb9416667c213fff1860a24ff3686b9aaef4c69

onlyhyde (approved):

onlyhyde (dismissed):

notJoon (approved):

onlyhyde (approved):

PR #365, opened 2024-11-01
GSW-1735 test: fix testcase related to machine reset

r3v4s:
## fix
1. fix testcases which relies on gno's machine context.
> due to https://github.com/gnolang/gno/pull/2975, machine gets reset between tests

### refactor
2. use `gno.land/p/demo/uassert` to compare test results.


onlyhyde (approved):

notJoon (approved):
👍

PR #370, opened 2024-11-04
GSW-1814 refactor: reduce iteration to calculate governance voting weight

r3v4s:
## refactor
1. reduce number of iteration to calculate governance proposal's voting power
> by using new variable `delegationStatHistory`
> which keeps each users's latest voing power with timestamp

2. clean up logic for keeping `delegationStatHistory` as small as possible


gov/staker/clean_delegation_stat_history.gno
```diff
@@ -0,0 +1,66 @@
...
+
+const BLOCK_PER_DAY = 86400 // based on block time 1s
+
+var (
+	lastCleanedHeight uint64 = 0
+)
+
+var (
+	running bool = true
+)
```
notJoon, line 17:
```suggestion
var (
	lastCleanedHeight uint64 = 0
	running bool = true
)
```

notJoon (approved):

onlyhyde (approved):

PR #372, opened 2024-11-08
GSW-1816 refactor: calculate block height based on const value

r3v4s:
## refactor
1. rather then using pre-calculated value as hardcoded, calculate them dynamically based on consts value

> chainging block interval time in consts.gno will do rest jobs


notJoon (approved):

r3v4s:
closing due to no file change

PR #379, opened 2024-11-13
refactor(pool): pool manager

notJoon:
# Description

Refactored `pool/pool_manager.gno` file


pool/pool_manager.gno
```diff
@@ -30,6 +51,87 @@ func init() {
 	feeAmountTickSpacing[10000] = 200 // 1%
```
r3v4s, line 51: It seems like tickSpacingMap has getter, why don't you also implement setter and use it instead of setting value directly `feeAmountTickSpacing[10000] = 200 // 1%`

pool/pool_manager.gno
```diff
@@ -30,6 +51,87 @@ func init() {
...
 }
 
+type poolInput struct {
```
r3v4s, line 54: variable name kinda confuses, wdyt about something like `createPoolParams` ?

pool/pool_manager.gno
```diff
@@ -30,6 +51,87 @@ func init() {
...
+		tickSpacing: tickSpacing,
+	}
+}
+
+func (p *poolInput) isSameTokenPath() bool {
+	return p.token0Path == p.token1Path
+}
+
+func (p *poolInput) isInOrder() bool {
+	if strings.Compare(p.token0Path, p.token1Path) > 0 {
```
r3v4s, line 83: ` > 0` vs `== 1`, and rule for this ?

pool/pool_manager.gno
```diff
@@ -39,82 +141,56 @@ func CreatePool(
...
-	// then check if token0Path == token1Path
-	if token0Path == token1Path {
-		panic(addDetailToError(
-			errDuplicateTokenInPool,
-			ufmt.Sprintf("pool_manager.gno__CreatePool() || expected token0Path(%s) != token1Path(%s)", token0Path, token1Path),
-		))
-	}
-
-	sqrtPriceX96 := u256.MustFromDecimal(_sqrtPriceX96)
+	poolInfo := newPoolInput(token0Path, token1Path, fee, _sqrtPriceX96)
```
r3v4s, line 149: poolInfo needs initialized after wrap/unwrap check

pool/pool_manager.gno
```diff
@@ -30,6 +51,87 @@ func init() {
...
+	}
+
+	token0Path, token1Path := p.wrap()
+
+	if !p.isInOrder() {
+		return "", "", errTokenSortOrder
+	}
+
+	return token0Path, token1Path, nil
+}
```
r3v4s, line 113: seems like `filterValidTokenPaths()` will pass if 
 - token0 = ugnot
 - token1 = wunogt 
or vice versa.

you also need to check if tokens are same after wrapping

pool/pool_manager.gno
```diff
@@ -39,82 +141,56 @@ func CreatePool(
...
-			"prevAddr", prevAddr,
-			"prevRealm", prevRealm,
-			"token0Path", token0Path,
-			"token1Path", token1Path,
-			"fee", ufmt.Sprintf("%d", fee),
-			"sqrtPriceX96", _sqrtPriceX96,
-			"internal_poolPath", poolPath,
-		)
-	}
+	gns.TransferFrom(a2u(std.PrevRealm().Addr()), a2u(consts.PROTOCOL_FEE_ADDR), poolCreationFee)
```
r3v4s, line 170: `poolCreationFee` is var, not consts.
Do you think we need to check if amount > 0 execute transfer, if not don't

pool/pool_manager_test.gno
```diff
@@ -132,6 +132,7 @@ func TestApiGetPools(t *testing.T) {
 	uassert.Equal(t, response.Size(), 2)
 }
 
-func gnsBalance(addr std.Address) uint64 {
+func gnsBalance(t *testing.T, addr std.Address) uint64 {
```
r3v4s, line 135: what is meaning of using T here, it's really unnecessary I think.

pool/pool_manager_test.gno
```diff
@@ -28,17 +28,17 @@ func TestCreateFooBarPool(t *testing.T) {
...
+	uassert.Equal(t, gnsBalance(t, consts.DEV_OPS), uint64(0))
 
 	std.TestSetRealm(adminRealm)
 	gns.Approve(a2u(consts.POOL_ADDR), poolCreationFee)
 
 	t.Run("wrong pool path", func(t *testing.T) {
 		uassert.PanicsWithMessage(
 			t,
-			`[GNOSWAP-POOL-012] tokens must be in lexicographical order || pool_manager.gno__CreatePool() || expected token0Path(gno.land/r/onbloc/foo) < token1Path(gno.land/r/onbloc/bar)`,
+			`[GNOSWAP-POOL-012] tokens must be in lexicographical order`,
```
r3v4s, line 41: add detail error messsage

r3v4s (changes requested):
Please check about filtering tokenPath(s) condition.
- and left few questions.

pool/pool_manager_test.gno
notJoon: just for calling the `t.Helper`, to indicate that it is a helper function for testing.

pool/pool_manager.gno
notJoon: I thought the current implementation was clearer, because I couldn't make a proper name to this constant because I don't know exactly what value it refers to at the moment, and I felt that using a hardcoded value in the `Set` function would only add to the confusion. So, I let it them.

pool/pool_manager.gno
notJoon: There is no strict rule for this, but I prefer `> 0` because this function determines the value relative to 0.

pool/pool_manager.gno
notJoon: fixed https://github.com/gnoswap-labs/gnoswap/pull/379/commits/d7ed7d1f3d4f3a5211dfa870b7ba537b156abbe9

pool/pool_manager.gno
notJoon: I Agree. that can be prevent unnecessary transfers.

notJoon: https://github.com/gnoswap-labs/gnoswap/pull/379/commits/8aade0fbf02500a7a86c5e86a1d33a483b8ac873

pool/pool_manager_test.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/379/commits/90ca23ce2aca81a8783d05c5c0ef3953a8894c92

pool/pool_manager.gno
notJoon: maybe like? https://github.com/gnoswap-labs/gnoswap/pull/379/commits/eff5fa9445c03078ace3aeff0f6cd5802faef913

r3v4s: Do we have to re-initialize? Can't we just initialize once?

notJoon: I think it wouldn’t have been necessary to add additional initialization before the modification since the value of `tokenPathXX` inside `poolInfo` (now `newPoolParams`) is directly updated its fields in the `wrap` function. Also, I didn't found relevant code from the original. 🫠

Could you clarify exactly what needs to be done? For now, I’ve added this line to update it uniformly.

r3v4s (dismissed):

r3v4s (approved):
👍

onlyhyde (approved):

PR #384, opened 2024-11-19
GSW-1959 fix: use clone liquidity to calculate emission

onlyhyde:
- when generate emission reward recipient, use clone value for liquidity


r3v4s (approved):
LGTM 👍

notJoon (approved):

PR #385, opened 2024-11-19
chore(int256): Prevent Passing `nil` Value When Compare Values

notJoon:
# Description

Added `nil` check when comparing two `int256` values


onlyhyde (approved):

notJoon:
@r3v4s The same guard logic has been added to the newly written int256 package

PR #387, opened 2024-11-19
GSW-1959 fix: liquidity pointer nil bug after unstake

onlyhyde:
- After unstake token, liquidity pointer become nil issue
- Fix this bug to copy value to calculation and then set result


notJoon (approved):

r3v4s (approved):
LGTM 👍

PR #394, opened 2024-11-20
chore: Remove String Concat With `+` Operator

notJoon:
# Description

* modified the part where strings were concatenated using the `+` operator to use a formatting function instead.
    - This change was made for readability rather than performance concerns (generally, using the `+` operator for string concatenation can result in minor performance degradation). 

* extracted the logic for splitting strings with a specific separator into a common utility.

### DIFF
--- a/gov/governance/_GET_vote.gno
+++ b/gov/governance/_GET_vote.gno
@@ -2,11 +2,11 @@ package governance
 
 import (
 	"std"
+	"strconv"
 
 	"gno.land/p/demo/ufmt"
 
-	"strconv"
-	"strings"
+	"gno.land/r/gnoswap/v1/common"
 )
 
 func GetVoteByVoteKey(voteKey string) bool {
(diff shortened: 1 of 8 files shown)


router/comptue_routes.gno
```diff
@@ -180,7 +181,7 @@ func findCandidatePools() []PoolWithMeta {
 			token0Path,
 			token1Path,
 			pFee,
-			token0Path + ":" + token1Path,
+			ufmt.Sprintf("%s:%s", token0Path, token1Path),
```
r3v4s, line 184: why not use string builder here?

gov/governance/_GET_vote.gno
```diff
@@ -110,12 +110,9 @@ func GetVotedAtByVoteKey(voteKey string) uint64 {
 }
 
 func divideVoteKeyToProposalIdAndUser(voteKey string) (uint64, std.Address) {
-	parts := strings.Split(voteKey, ":")
-	if len(parts) != 2 {
-		panic(addDetailToError(
-			errInvalidInput,
-			ufmt.Sprintf("_GET_vote.gno__divideVoteKeyToProposalIdAndUser() || voteKey(%s) is invalid", voteKey),
```
r3v4s, line 117: I really don't think removing detail error message is good idea.
It makes us hard to why it failed.
@onlyhyde Any thoughts?

r3v4s (changes requested):
IMHO, removing detail error message should not be happened.
@onlyhyde WDYT

gov/governance/_GET_vote.gno
onlyhyde: yes i agree with @r3v4s . error message is already confirmed our policy. so, we have to use previous error message

router/comptue_routes.gno
notJoon: in this case, these is no significant difference between using strings.Builder and formatting function. so I use latter one.

gov/governance/_GET_vote.gno
notJoon: reverted for panic messages

https://github.com/gnoswap-labs/gnoswap/pull/394/commits/1b17b6c4312417f01c59f084130cce786b482c45

r3v4s (approved):

onlyhyde (approved):

PR #396, opened 2024-11-21
GSW-1699 feat: Update Tokens to Follow New GRC20 Specs

notJoon:
# Description

unable to run the Gnoswap test on local env due to a type error, so I fixed it.

updated the token files to align with the latest GRC20 specification.


r3v4s (approved):
Changing changes seems fine.
But our dev chain uses test5 image, which uses previous grc20 spec which means spec in this pr wont' work.

@onlyhyde What's your call for this issue?

onlyhyde (approved):

PR #400, opened 2024-11-22
refactor, test(pool): Change `updatePosition` param

notJoon:
# Description

- Change `updatePosition` function's parameter to only take `ModifyPositionParams` type
- Add unit test for `modifyPosition`. this test covers both `update` and `modify` function

PR #402, opened 2024-11-24
refactor: modifyPosition

notJoon:
# Description

**Merge This PR After #400 Is Completely Merged**

Refactored `modifyPosition` function
- removed `NilToZero` converter, because those lines are meaningless if we defined `amount0`, `amount1` value with `i256.Zero`.


notJoon:
PR #400 was merged. So, changed it to review status.

pool/type.gno
```diff
@@ -171,7 +172,7 @@ func (t *TickInfo) init() {
 
 type Ticks map[int32]TickInfo          // tick => TickInfo
 type TickBitmaps map[int16]*u256.Uint  // tick(wordPos) => bitmap(tickWord ^ mask)
-type Positions map[string]PositionInfo // positionKey => PositionInfo
+type Positions map[string]PositionInfo // positionKey => PositionInfo
```
r3v4s, line 175: Aren't we supposed to replace all map to avl?
Since you imported avl from #L8 but doesn't using it. Maybe its WIP??

r3v4s (commented):
code looks good, but left one comment for map -> avl

pool/type.gno
notJoon: All the PRs are tangled together, I think avl must be apply last.

onlyhyde (approved):

PR #407, opened 2024-11-29
feat: add getter to query pool's position's data

r3v4s:
## feat
1. (current for debug) add getter to query pool's position's data


onlyhyde (approved):

onlyhyde:
@notJoon This function adds a getter function to debug the issue of not finding a value by key when using map. This is needed for test validation, so please review it quickly.

notJoon (approved):

PR #403, opened 2024-11-28
refactor: pool.gno

notJoon:
- mainly for the Swap function
- add test for validating entry functions

PR #410, opened 2024-12-01
GSW-1838 refactor: type, api and getter

r3v4s:
## refactor
1. rename filenames
2. add missing test for getter

###
wait for #403
> struct field type might be change (smaller type for numeric, avl rather than map...)


dongwon8247:
Don't mix with other files. Work on each file and open a PR separately.
This PR to be closed.

PR #414, opened 2024-12-02
GSW-1838 refactor: use grc20reg

r3v4s:
## refactor
1. use gno's `grc20reg` realm to support dynamic token in pool.
> it completely replaces previous `token_register.gno`

closes #409 

### FYI
this pr SHOULD BE MERGE AFTER #413

PR #412, opened 2024-12-02
GSW-1838 feat: grc20reg

r3v4s:
## feat
1. feat grc20reg from (gnolang/gno#3135)
- replace previous token_register to latest grc20reg

PR #422, opened 2024-12-06
GSW-1838 fix: test errors are fixed

onlyhyde:
- Integrate helper functions for tests
- Change file extensions to prevent test code in the test folder from being executed
- Fixing failure errors due to code integration
- Known issue : Fixed additional test failure case related to getter


onlyhyde:
@r3v4s 
There are still test failures related to the getter. There are a lot of functions that need to be fixed, so I'll ask the existing person in charge. Also, the pool_manager related test code is all outdated. Please fix it according to our rule.

onlyhyde:
With the additional commit from @r3v4s , all test code will pass when the current PR is applied.

notJoon (approved):

r3v4s (approved):
LGTM 👍 

Just simple question,
Can we make `users.Resolve(admin)` it as var or const ??
I really don't see pros for keep resolving AddressOrName type to address.

r3v4s:
FYI, @onlyhyde @notJoon 

to finish #414 (which also includes work from #412 #413) this pr needs to be merge first, so I can update some test cases without many conflicts.

PR #432, opened 2024-12-10
GSW-1839 refactor: integrated helper and test code

onlyhyde:
- integrated helper with nft helper
- add test helper code
- add test code for helper
- change file filename


position/helper.gno
```diff
@@ -28,9 +37,95 @@ func tid(tokenId interface{}) grc721.TokenID {
 	case grc721.TokenID:
 		return tokenId.(grc721.TokenID)
 	default:
-		panic(addDetailToError(
-			errInvalidInput,
-			"helper.gno__tid() || unsupported tokenId type",
-		))
+		panic(newErrorWithDetail(errInvalidInput, "unsupported tokenId type"))
```
r3v4s, line 40: just curious, can we add what kind of type it had??
something like `unsupported token id type{string/uint64/int/.../}`

r3v4s (approved):

notJoon (approved):

PR #437, opened 2024-12-12
refactor(staker): staker

notJoon:
# Description

- Remove unused internal functions and imports
- Moved external incentive related functions from the `staker` file to a separate `staker_external_incentive` file
- Extract condition check as functions
- Change the parameters of the `getTokenPairBalanceFromPosition` function
    - The pool path is already obtained before this function is called with `pn.PositionGetPositionPoolKey(tokenId)` method, it is unnecessary to process it internally.
```diff
- func getTokenPairBalanceFromPosition(tokenId uint64) (string, string) {
+ func getTokenPairBalanceFromPosition(poolPath string, tokenId uint64) (string, string) {
- 	poolKey := pn.PositionGetPositionPoolKey(tokenId)
        // ...
}
```

## Newly Added Test Files

- `manage_pool_tiers`
- `reward_internal_emission`
- `reward_manager`
- `reward_recipient_store`
- `staker`

## Test Result

Screenshot 2024-12-16 at 7 32 14 PM


r3v4s (changes requested):
- tc doesn't run with current master branch as base
- do not inline `get -> data check -> if nil panic -> if not nil type assertion` for avl data
  - make it as separate function
- for testcase screenshot, run it from gnoswap's repo(with current br branch), not from gno

notJoon:
>     * for testcase screenshot, run it from gnoswap's repo(with current br branch), not from gno

I don't think this will be a problem, since I just moved the entire main branch over to the gno side.

staker/calculate_pool_position_reward.gno
```diff
@@ -518,3 +517,13 @@ func externalDepositGnsAmount() uint64 {
 
 	return amount
 }
+
+
+func getPoolTiers() map[string]InternalTier {
```
r3v4s, line 522: should be avl

staker/staker.gno
```diff
@@ -71,310 +86,142 @@ func StakeToken(tokenId uint64) (string, string, string) {
...
-	deposit.numberOfStakes++
-	deposit.stakeTimestamp = time.Now().Unix()
-	deposit.stakeHeight = std.GetHeight()
-	deposit.targetPoolPath = poolPath
-	deposits[tokenId] = deposit
-
-	if callerIsOwner { // if caller is owner, transfer NFT ownership to staker contract
-		transferDeposit(tokenId, consts.STAKER_ADDR)
+	if err := applyStake(result); err != nil {
+		panic(err)
```
r3v4s, line 98: no need err codes?

staker/_RPC_api_stake.gno
```diff
@@ -98,30 +99,38 @@ func ApiGetRewards() string {
...
-			stakedOrCreatedAt := max(deposit.stakeTimestamp, incentive.startTimestamp)
+			stakedOrCreatedAt := max(deposit.stakeTimestamp, ictv.startTimestamp)
 			now := time.Now().Unix()
 			if now < stakedOrCreatedAt {
-				continue
+				return
 			}
 
-			externalWarmUpAmount, exist := positionsExternalWarmUpAmount[tokenId][incentiveId]
+			externalWarmUpAmount, exist := positionsExternalWarmUpAmount[tokenId][ictvId]
```
r3v4s, line 119: Separate getter(and setter) will be much better for nested avl datas.

staker/reward_recipient_store.gno
```diff
@@ -195,6 +195,33 @@ func NewPoolLiquidity() *PoolLiquidity {
...
+
+	// nil check for totalLiquidity
+	if p.totalLiquidity == nil {
+		p.totalLiquidity = u256.Zero()
+	}
+
+	// Add position's liquidity to total liquidity
+	positionLiquidity := position.GetLiquidity()
+	if positionLiquidity != nil {
+		p.totalLiquidity.Add(p.totalLiquidity, positionLiquidity)
```
r3v4s, line 218: liquidity is already huge number (that's why we use uint256 for it).
adding u256 + u256 does need overflow check.

staker/calculate_pool_position_reward.gno
notJoon: calc_pool_position_XXX files will be removed. don't need to change

staker/_RPC_api_stake.gno
notJoon: same as here: https://github.com/gnoswap-labs/gnoswap/pull/437#discussion_r1897793290

that global variable will be removed

staker/reward_recipient_store.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/437/commits/9a1d977976448874c848a1a4d1a2ff63a03357a3

staker/staker.gno
notJoon: The `applyStake` already returns `errNoPermission` in the `transferDeposit` function, so it does not have to handle it again.

notJoon:
branch has merged with PR #438

PR #439, opened 2024-12-13
GSW-1845 refactor: gns (halving related)

r3v4s:
# Description
1. replace `map` to []array
> turns out key-value (map or avl) really doesn't need for gns.
2. make unit test as simple as possible (complete test goes into `test` folder as scenario test)
3. use const as much as possible rather than hard coded value
4. add more halving related testcase

image


_deploy/r/gnoswap/gns/halving.gno
```diff
@@ -22,6 +23,26 @@ import (
...
+		9_375_000_000_000 * 12,  // Year 4:  112500000000000
+		4_687_500_000_000 * 12,  // Year 5:  56250000000000
+		4_687_500_000_000 * 12,  // Year 6:  56250000000000
+		2_343_750_000_000 * 12,  // Year 7:  28125000000000
+		2_343_750_000_000 * 12,  // Year 8:  28125000000000
+		1_171_875_000_000 * 12,  // Year 9:  14062500000000
+		1_171_875_000_000 * 12,  // Year 10: 14062500000000
+		1_171_875_000_000 * 12,  // Year 11: 14062500000000
+		1_171_875_000_000 * 12,  // Year 12: 14062500000000
+	}
```
notJoon, line 43: `HALVING_AMOUNTS_PER_YEAR` has a fixed side, I think it would be better to change it to a fixed-size array.

_deploy/r/gnoswap/gns/halving.gno
```diff
@@ -30,67 +51,48 @@ var (
...
 
 func init() {
-	height := std.GetHeight()
-
-	startHeight = height
+	startHeight = std.GetHeight()
 	startTimestamp = time.Now().Unix()
 
-	initialized = true
+	for year := HALVING_START_YEAR; year <= HALVING_END_YEAR; year++ {
```
notJoon, line 75: It would be better to extract the loop's internal logic. Because the `init` function cannot be tested, but we can verify the extracted function.

_deploy/r/gnoswap/gns/halving.gno
```diff
@@ -143,64 +145,55 @@ func setAvgBlockTimeInMs(ms int64) {
...
 	adjustedAmountPerBlock := amountLeft / uint64(blockLeft)
 
 	// update it
-	amountPerBlockPerHalvingYear[year] = adjustedAmountPerBlock
+	setAmountPerBlockPerHalvingYear(year, adjustedAmountPerBlock)
 
-	// adjust halving block
-	for keyYear, _ := range halvingYearBlock {
-		yearEnd := halvingYearTimestamp[keyYear]
+	for year := int64(1); year < 13; year++ {
```
notJoon, line 166:
```suggestion
	for year := int64(1); year < HALVING_END_YEAR + 1; year++ {
```

notJoon (changes requested):
I've left some comments. Also, I think it would be better if the declarations of constant and variables were more structed.

_deploy/r/gnoswap/gns/halving.gno
r3v4s: I don't quite get it, can you enlighten me with more detail? (small peace of suggestion  would be perfect)

notJoon: theres no other details for this line. I don't prefer hard coded value

_deploy/r/gnoswap/gns/halving.gno
r3v4s: https://github.com/gnoswap-labs/gnoswap/blob/b98a424f2714129b7900860392f2f9905753efae/_deploy/r/gnoswap/gns/halving.gno#L30

_deploy/r/gnoswap/gns/halving.gno
r3v4s: https://github.com/gnoswap-labs/gnoswap/blob/b98a424f2714129b7900860392f2f9905753efae/_deploy/r/gnoswap/gns/halving.gno#L188

notJoon (dismissed):

onlyhyde (approved):

notJoon (approved):

PR #449, opened 2024-12-23
GSW-1844 refactor: query distributed amount for each targets

r3v4s:
## Description

1. getter(and setter) for each emission distribution target's distributed and accumualted amount
2. return `distributed` amount in MintAndDistributeGns
3. make unnecessary public setter to private
4. if amount left due to rounding errors, save it to distribute in next time
5. replace to map to avl


notJoon (dismissed):

notJoon (dismissed):

onlyhyde (approved):

notJoon (approved):
👍

PR #452, opened 2024-12-26
GSW-1668 feat: use grc20reg to support multiple grc20 tokens

r3v4s:
## Description
- use `grc20reg` realm instead previous token_register pattern to support grc20 tokens
- since grc20reg doesn't provide registered token lists, we need to parse it by calling Render()


_deploy/r/gnoswap/common/grc20reg_helper.gno
```diff
@@ -58,6 +60,27 @@ func MustRegistered(path string) {
...
+// NOTE:
+// - Unfortunate, grc20reg doesn't support this.
+// - We need to parse the rendered grc20reg page to get the list of registered tokens.
+func ListRegisteredTokens() []string {
+	render := grc20reg.Render("")
+	return extractTokenPathsFromRender(render)
+}
+
+func extractTokenPathsFromRender(render string) []string {
+	re := regexp.MustCompile(`\[gno\.land/r/[^\]]+\]`)
```
notJoon, line 73: this regex must be a global variable

_deploy/r/gnoswap/common/grc20reg_helper.gno
```diff
@@ -58,6 +60,27 @@ func MustRegistered(path string) {
...
+	matches := re.FindAllString(render, -1)
+
+	tokenPaths := []string{}
```
notJoon, line 76: here, you can pre-allocate capacity.

```suggestion
	tokenPaths := make([]string, 0, len(matches))
```

notJoon (changes requested):
LGTM. I left some comments

notJoon (approved):
👍 👍

onlyhyde (approved):

PR #456, opened 2025-01-03
feat: `uint256` single value overflow checker

notJoon:
# Description

The `IsOverflow()` function checks if a uint256 number has overflowed by examining the 255th bit (the most significant bit). In uint256, the valid range is from 0 to 2^255-1, so any number with the 255th bit set is considered an overflow.


onlyhyde (approved):

r3v4s (approved):
nice catch

PR #458, opened 2025-01-08
refactor: event message

onlyhyde:


dongwon8247:
@onlyhyde Let's merge this one first if it has no further changes, and finish the left ones in another PR.

onlyhyde:
This pr only has event updates, and there is a conflict because emission's modification is a big change in the staker pr. I think it's better for code safety to rework and update the staker pr after applying it. I'll close the current pr and rewrite it after merging the staker pr (#461).

PR #459, opened 2025-01-13
feat: referral

notJoon:
# Description

This PR introduces a referral system that allows authorized contracts to manage referral relationships between addresses. The system enables tracking and managing referrer-referee relationships.

## Structure

```mermaid
classDiagram
	class ReferralKeeper {
		<<interface>>
		+register(addr, refAddr std.Address)
		+update(addr, newAddr std.Address)
		+remove(addr std.Address)
		+has(addr std.Address)
		+get(addr std.Address)
	}
	class Keeper {
		-store *avl.Tree
		+register(addr, refAddr std.Address)
		+update(addr, newRefAddr std.Address)
		+remove(addr std.Address)
		+has(addr std.Address)
		+get(addr std.Address)
	}
	class Referral {
		-keeper ReferralKeeper
		+Register(addr, refAddr std.Address)
		+Update(addr, newRefAddr std.Address)
		+Remove(addr std.Address)
		+Has(addr std.Address)
		+Get(addr std.Address)
	}
	
	Referral --> ReferralKeeper : uses
	Keeper ..|> ReferralKeeper : implements
```

### Implementation Details

The referral system consists of the following components:

1. Core Components:
   - ReferralKeeper interface: Defines the fundamental operations for referral management
   - AVL Tree-based storage: Efficiently stores referral relationships

2. Main Operations:
   - `Register`: Create new referral relationships
   - `Update`: Modify existing referrals
   - `Remove`: Delete referral relationships
   - Query (`Get`, `Has`): Check existence and get referral information

3. Security:
   - Strict caller validation through a lookup table
   - Address format validation
   - Zero address handling for removal operations

## Permissions

Only the following addresses can modify referral data:

   - Governance contracts
   - Router
   - Position
   - Staker

### DIFF
--- a/referral/keeper.gno
+++ b/referral/keeper.gno
@@ -37,31 +37,42 @@ func (k *keeper) update(addr, newRefAddr std.Address) error {
 // setReferral handles the common logic for registering and updating referrals.
 // It validates the addresses and caller, then stores the referral relationship.
 //
+// Note: The current implementation allows circular references, but since it only manages
+// simple reference relationships, cycles are not a significant issue. However, when introducing
+// a referral-based reward system in the future or adding business logic where cycles could cause problems,
+// it will be necessary to implement validation checks.
+//
 // TODO: need to discuss what values to emit as event
 func (k *keeper) setReferral(addr, refAddr std.Address, eventType string) error {
 	if err := isValidCaller(std.GetOrigCaller()); err != nil {
 		return err
 	}
 
-	if !addr.IsValid() {
-		return NewError(ErrInvalidAddress)
+	if err := k.validateAddresses(addr, refAddr); err != nil {
+		return err
 	}
 
 	if refAddr == zeroAddress {
-		k.remove(addr)
-	}
-
-	if !refAddr.IsValid() {
-		return NewError(ErrInvalidAddress)
+		return k.remove(addr)
 	}
 
 	k.store.Set(addr.String(), refAddr.String())
-
 	std.Emit(eventType,
 		"my_address", addr.String(),
 		"ref_address", refAddr.String(),
 	)
 
+	return nil
+}
+
+// validateAddresses checks if the given addresses are valid for referral
+func (k *keeper) validateAddresses(addr, refAddr std.Address) error {
+	if !addr.IsValid() || (refAddr != zeroAddress && !refAddr.IsValid()) {
+		return NewError(ErrInvalidAddress)
+	}
+	if addr == refAddr {
+		return NewError(ErrSelfReferral)
+	}
 	return nil
 }
 
--- a/referral/keeper.gno
+++ b/referral/keeper.gno
@@ -2,13 +2,20 @@ package referral
 
 import (
 	"std"
+	"time"
 
 	"gno.land/p/demo/avl"
+)
+
+const (
+	// MinTimeBetweenUpdates represents minimum duration between operations
+    MinTimeBetweenUpdates int64 = 24 * 60 * 60
 )
 
 // keeper implements the `ReferralKeeper` interface using an AVL tree for storage.
 type keeper struct {
 	store *avl.Tree
+	lastOps map[string]int64
 }
 
 // check interface implementation at compile time
@@ -19,6 +26,7 @@ var _ ReferralKeeper = &keeper{}
 func NewKeeper() ReferralKeeper {
 	return &keeper{
 		store: avl.NewTree(),
+		lastOps: make(map[string]int64),
 	}
 }
 
@@ -45,6 +53,10 @@ func (k *keeper) update(addr, newRefAddr std.Address) error {
 // TODO: need to discuss what values to emit as event
 func (k *keeper) setReferral(addr, refAddr std.Address, eventType string) error {
 	if err := isValidCaller(std.PrevRealm().Addr()); err != nil {
+		return err
+	}
+
+	if err := k.checkRateLimit(addr.String()); err != nil {
 		return err
 	}
 
@@ -139,4 +151,20 @@ func (k *keeper) isEmpty() bool {
         return true // stop iteration on first item
     })
     return empty
+}
+
+// checkRateLimit verifies if enough time has passed since the last operation
+func (k *keeper) checkRateLimit(addr string) error {
+    now := time.Now().Unix()
+    
+    if lastOpTime, exists := k.lastOps[addr]; exists {
+        timeSinceLastOp := now - lastOpTime
+        
+        if timeSinceLastOp < MinTimeBetweenUpdates {
+            return NewError(ErrTooManyRequests)
+        }
+    }
+    
+    k.lastOps[addr] = now
+    return nil
 }
--- a/referral/keeper.gno
+++ b/referral/keeper.gno
@@ -63,8 +63,8 @@ func (k *keeper) setReferral(addr, refAddr std.Address, eventType string) error 
 	k.store.Set(addr.String(), refAddr.String())
 	std.Emit(
 		eventType,
-		"my_address", addr.String(),
-		"ref_address", refAddr.String(),
+		"myAddress", addr.String(),
+		"refAddress", refAddr.String(),
 	)
 
 	return nil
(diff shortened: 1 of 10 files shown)


referral/keeper.gno
```diff
@@ -0,0 +1,128 @@
...
+	if err := k.validateAddresses(addr, refAddr); err != nil {
+		return err
+	}
+
+	if refAddr == zeroAddress {
+		return k.remove(addr)
+	}
+
+	k.store.Set(addr.String(), refAddr.String())
+	std.Emit(eventType,
```
dongwon8247, line 60: We need to think of ways to sanitize and validate all event data before emitting to avoid spam or DoS attacks.

referral/keeper.gno
```diff
@@ -0,0 +1,128 @@
...
+// The keeper is initialized with an empty AVL tree for storing referral relationships.
+func NewKeeper() ReferralKeeper {
+	return &keeper{
+		store: avl.NewTree(),
+	}
+}
+
+// register implements the `register` method of the `ReferralKeeper` interface.
+// It sets a new referral relationship between the given address and referral address.
+func (k *keeper) register(addr, refAddr std.Address) error {
```
dongwon8247, line 27: I have a concern that since there's no rate limiting on key functions like `register`, `update`, or `remove`, there may be spam to bloat the AVL tree. Or break the referral system by frequently updating the addresses.

referral/keeper.gno
```diff
@@ -0,0 +1,128 @@
...
+	std.Emit(eventType,
+		"my_address", addr.String(),
+		"ref_address", refAddr.String(),
+	)
+
+	return nil
+}
+
+// validateAddresses checks if the given addresses are valid for referral
+func (k *keeper) validateAddresses(addr, refAddr std.Address) error {
```
dongwon8247, line 69: This validation must be added to `register` and `update` too.

notJoon: Since this validation is being checked through the `setReferral` function, there is no need to call it separately.

onlyhyde (commented):
Referral Contract requires a book that manages Caller Address -> Referral Address.

referral/keeper.gno
```diff
@@ -0,0 +1,128 @@
...
+// It validates the addresses and caller, then stores the referral relationship.
+//
+// Note: The current implementation allows circular references, but since it only manages
+// simple reference relationships, cycles are not a significant issue. However, when introducing
+// a referral-based reward system in the future or adding business logic where cycles could cause problems,
+// it will be necessary to implement validation checks.
+//
+// TODO: need to discuss what values to emit as event
+func (k *keeper) setReferral(addr, refAddr std.Address, eventType string) error {
+	if err := isValidCaller(std.GetOrigCaller()); err != nil {
```
onlyhyde, line 47: You should get the Addr via prevRealm instead of GetOrigCaller.

referral/keeper.gno
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return nil
+}
+
+// remove implements the `remove` method of the `ReferralKeeper` interface.
+// It validates the caller and address before removing the referral relationship.
+//
+// TODO: need to discuss what values to emit as event
+func (k *keeper) remove(target std.Address) error {
+	if err := isValidCaller(std.GetOrigCaller()); err != nil {
```
onlyhyde, line 84: You should get the Addr via prevRealm instead of GetOrigCaller.

referral/utils.gno
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// validCallers is a lookup table of addresses that are authorized to modify referral data.
+// This includes governance contracts, router, position manager, and staker contracts.
+var validCallers = map[std.Address]bool{
+	consts.GOV_GOVERNANCE_ADDR: true,
+	consts.GOV_STAKER_ADDR:     true,
+	consts.ROUTER_ADDR:         true,
+	consts.POSITION_ADDR:       true,
+	consts.STAKER_ADDR:         true,
```
onlyhyde, line 16: should include launchpad

referral/keeper.gno
```diff
@@ -0,0 +1,128 @@
...
+		return err
+	}
+
+	if err := k.validateAddresses(addr, refAddr); err != nil {
+		return err
+	}
+
+	if refAddr == zeroAddress {
+		return k.remove(addr)
```
onlyhyde, line 56: EVENT EMIT is required even if the ADDR is deleted.

onlyhyde (changes requested):

referral/keeper.gno
notJoon: How about set a time limit once a referral is established, like 1 day?

However, since calculating and tracking within inside the contract may cause issues like increased data or computational load, perhaps we could block it on the frontend side instead?

dongwon8247: There's no meaning blocking from the frontend side.

If we set like 1day of blocks for re-registering it, would it be a big overload?

dongwon8247:
One question: Thinking about preventing people from registering their other wallets as a referral, to manipulate the system (getting more points). We only use this referral contract to know the referral-referee relationships and point calculation/prevention will be on our back-end side, correct?

notJoon:
> One question: Thinking about preventing people from registering their other wallets as a referral, to manipulate the system (getting more points). We only use this referral contract to know the referral-referee relationships and point calculation/prevention will be on our back-end side, correct?

@dongwon8247 Yes, that's correct. This package only stores referral-referee relationships and I believe that's how it should be.

referral/keeper.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/459/commits/cb7ae69f69b1bd81e86a34cb279f1f75e1a30579

referral/utils.gno
notJoon: update address map
https://github.com/gnoswap-labs/gnoswap/pull/459/commits/cb7ae69f69b1bd81e86a34cb279f1f75e1a30579

referral/keeper.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/459/commits/cb7ae69f69b1bd81e86a34cb279f1f75e1a30579

referral/keeper.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/459/commits/cb7ae69f69b1bd81e86a34cb279f1f75e1a30579

notJoon:
@onlyhyde 

> Referral Contract requires a book that manages Caller Address -> Referral Address.

The store field of the keeper handles that role.

onlyhyde (approved):

referral/keeper.gno
onlyhyde: @dongwon8247 
I think it's good to have a delay time for resetting after setting referrals, maybe a day, but it would be good to add guard logic and update the delay time with more thought.

@notJoon 
Please add a guard to check the previous registration time when trying to re-register after registration and register when the registration time is available.

notJoon: https://github.com/gnoswap-labs/gnoswap/pull/459/commits/dbe7d684ae8c250b450d1d25289a5482dab7fd0c

referral/keeper_test.gno
```diff
@@ -73,13 +73,6 @@ func TestRegister(t *testing.T) {
 			wantErr:     true,
 			errCode:     ErrSelfReferral,
 		},
-		{
-			name:        "zero address referral",
```
dongwon8247: Why delete? This will be needed to reset the existing referral address.

referral/keeper.gno
```diff
@@ -102,6 +102,10 @@ func (k *keeper) remove(target std.Address) error {
 		return err
 	}
 
+	if err := k.checkRateLimit(target.String()); err != nil {
```
dongwon8247, line 105: Just to make sure, this limit only triggers when a valid address is registered as a referral, right?

notJoon: The order is incorrect. However, since the map only contains valid address values, if an invalid value is entered, it won't be found in the search. Anyway, fixed.

https://github.com/gnoswap-labs/gnoswap/pull/459/commits/8556fe44adc8434e54170352396916ef4c3d9b60

referral/keeper_test.gno
notJoon: There was a strange type-related error so I had deleted it, but now I've found a workaround and restored it.

https://github.com/gnoswap-labs/gnoswap/pull/459/commits/8556fe44adc8434e54170352396916ef4c3d9b60

dongwon8247 (approved):
Well done. LGTM

onlyhyde (approved):

PR #464, opened 2025-01-17
feat/manage protocol fee list

r3v4s:
## Description
1. `protocol_fee` manages accumulated token path with its amount
> to avoid iterating (entire) registered token list

2. each contract which send certain amount to protocol_fee contract has to set which token and how much has been sent

### DIFF
--- a/protocol_fee/token_list_with_amount.gno
+++ b/protocol_fee/token_list_with_amount.gno
@@ -14,7 +14,7 @@ var (
 	tokenListWithAmount = make(map[string]uint64) // tokenPath -> amount
 )
 
-// TokenList returns only the token path.
+// TokenList returns only the list of token path.
 // If positive is true, it returns only the token path with amount > 0.
 // If positive is false, it returns all the token path.
 func TokenList(positive bool) []string {
@@ -47,7 +47,7 @@ func AddToProtocolFee(tokenPath string, amount uint64) {
 // only `gov/staker` can execute this function.
 func ClearTokenListWithAmount() {
 	assertOnlyGovStaker()
-	tokenListWithAmount = map[string]uint64{}
+	clearTokenListWithAmount()
 }
 
 // TransferProtocolFee transfers the protocol fee to devOps and gov/staker.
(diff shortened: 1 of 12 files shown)


protocol_fee/token_list_with_amount.gno
```diff
@@ -0,0 +1,126 @@
...
+
+	sentToDevOps := []string{}
+	sentToGovStaker := []string{}
+	toReturn := map[string]uint64{}
+
+	for token, amount := range tokenListWithAmount {
+		balance := common.BalanceOf(token, consts.PROTOCOL_FEE_ADDR)
+
+		// amount in `tokenListWithAmount` is corrupted, it should be equal to balance
+		if amount != balance {
```
onlyhyde, line 66: Anyone can transfer assets to a PROTOCOL_FEE contract. The guard on that line is panicking if the assets on the books don't match the balance in the actual token contract. If someone maliciously sends assets to the PROTOCOL_FEE Contract, PROTOCOL_FEE will never receive them back. We recommend removing the guard on that line as it can cause problems.

protocol_fee/token_list_with_amount.gno
```diff
@@ -0,0 +1,126 @@
...
+			panic(errInvalidAmount.Error())
+		}
+
+		if balance > 0 {
```
onlyhyde, line 70: It looks like the condition should enter when the amount, not balance, is greater than or equal to 0.

onlyhyde (changes requested):

protocol_fee/token_list_with_amount.gno
r3v4s: ad464fca7922b2f00b5db4aa0a06efe3d8b2e383

protocol_fee/token_list_with_amount.gno
r3v4s: ad464fca7922b2f00b5db4aa0a06efe3d8b2e383

protocol_fee/token_list_with_amount_test.gno
```diff
@@ -0,0 +1,237 @@
...
+	"std"
+	"testing"
+
+	"gno.land/p/demo/uassert"
+
+	"gno.land/r/gnoswap/v1/common"
+	"gno.land/r/gnoswap/v1/consts"
+
+	_ "gno.land/r/onbloc/baz"
+	_ "gno.land/r/onbloc/foo"
```
notJoon, line 13: remove unused packages

```suggestion
```

dongwon8247 (approved):

onlyhyde (approved):

PR #476, opened 2025-01-21
chore: Use `strconv` in event when formatting values

### LINKED ISSUE #472: Using `strconv` Instead of `ufmt` When Emitting Events
# Description

Currently, most of the codes uses `ufmt.Sprintf` function to format numeric values when emitting events. Since IO operations generally use more memory than `strconv`, we need to modify all these cases.

notJoon:
closes #472 

Note: While I could move the `parseUint` and `parseInt` functions to the `common` package, I added them to the utils file to prevent an increase in cycles due to package calls.


contract/r/gnoswap/emission/utils.gno
```diff
@@ -90,3 +91,29 @@ func assertSumDistributionPct(pct01, pct02, pct03, pct04 uint64) {
...
+	case int32:
+		return strconv.FormatInt(int64(v), 10)
+	case int64:
+		return strconv.FormatInt(v, 10)
+	case int:
+		return strconv.Itoa(v)
+	default:
+		panic(ufmt.Sprintf("invalid type: %T", v))
+	}
+}
```
r3v4s, line 119: Seems like you're willing to format numeric values, in this case please handle all size
`8` `16` `32` `64`

r3v4s (approved):
LGTM

contract/r/gnoswap/emission/utils.gno
notJoon: I have only registered the types that are currently being used in the gnoswap repository

r3v4s:
What do you think about moving all conversion in to p/gnoswap/XXX ?

onlyhyde (approved):

notJoon:
> What do you think about moving all conversion in to p/gnoswap/XXX ?

@r3v4s I will consider it when dividing the packages later.

PR #481, opened 2025-01-21
feat: use std.Address (rather then users.AddressOrName)

r3v4s:
## Description
use `std.Address` type(rather than user.AddressOrName) for grc20 tokens in gnoswap contract.

### reference
- https://github.com/gnoswap-labs/gnoswap/pull/465#discussion_r1922981910


contract/r/gnoswap/gov/governance/proposal_test.gno
```diff
@@ -284,11 +283,11 @@ func TestUpdateProposalsState(t *testing.T) {
 			proposals = avl.NewTree()
 
 			proposal := tt.setupProposal(tt.currentTime)
-			proposals.Set(fomatUint(1), proposal)
+			proposals.Set(formatUint(uint32(1)), proposal)
```
notJoon, line 286: I think we don't need to covert type here
```suggestion
			proposals.Set(formatUint(1), proposal)
```

contract/r/gnoswap/gov/governance/proposal_test.gno
```diff
@@ -284,11 +283,11 @@ func TestUpdateProposalsState(t *testing.T) {
...
 
 			updateProposalsState()
 
-			updatedProposal, exists := proposals.Get(fomatUint(1))
+			updatedProposal, exists := proposals.Get(formatUint(uint32(1)))
```
notJoon, line 290: also here
```suggestion
			updatedProposal, exists := proposals.Get(formatUint(1))
```

contract/r/gnoswap/gov/governance/proposal_test.gno
r3v4s: It needs to be, if you run test file you can see it panics without converting

contract/r/gnoswap/gov/governance/proposal_test.gno
r3v4s: It needs to be, if you run test file you can see it panics without converting

notJoon (dismissed):

contract/r/gnoswap/position/position.gno
```diff
@@ -828,7 +828,7 @@ func decreaseLiquidity(params DecreaseLiquidityParams) (uint64, *u256.Uint, *u25
 	}
 
 	caller := getPrevAddr()
-	beforeWugnotBalance := wugnot.BalanceOf(a2u(caller)) // before unwrap
+	beforeWugnotBalance := wugnot.BalanceOf(common.AddrToUser(caller)) // before unwrap
```
dongwon8247, line 831: can you add a comment saying this will be modified once the `wugnot` package spec is changed? It would be great if you open a fix in the gno repo.

dongwon8247 (approved):

contract/r/gnoswap/position/position.gno
r3v4s: aed9cbc41761ab582684d81c7e06c6258e1f4f31

notJoon (approved):

contract/r/gnoswap/position/position.gno
r3v4s: and make pr to fix wugnot(and foo20) to user std.address
https://github.com/gnolang/gno/pull/3580

dongwon8247 (approved):

PR #483, opened 2025-01-22
refactor: event message change

onlyhyde:


notJoon (dismissed):
LGTM

contract/r/gnoswap/emission/distribution.gno
```diff
@@ -287,3 +287,18 @@ func GetHalvingBlocksInRange(start, end int64) ([]int64, []uint64) {
...
+	case LIQUIDITY_STAKER:
+		return "LIQUIDITY_STAKER"
+	case DEVOPS:
+		return "DEVOPS"
+	case COMMUNITY_POOL:
+		return "COMMUNITY_POOL"
+	case GOV_STAKER:
+		return "GOV_STAKER"
+	default:
+		return "UNKNOWN"
```
r3v4s, line 302: should panic

contract/r/gnoswap/position/position.gno
```diff
@@ -237,8 +237,8 @@ func Mint(
 		"Mint",
 		"prevAddr", prevAddr,
 		"prevRealm", prevPkgPath,
-		"tickLower", formatInt(processedInput.tickLower),
-		"tickUpper", formatInt(processedInput.tickUpper),
+		"tickLower", formatInt(int64(processedInput.tickLower)),
```
r3v4s, line 240: @onlyhyde @notJoon 
I think `formatInt` function needs to more type than current does, otherwise we'll have convert every time just like this file does.

r3v4s (changes requested):
event seems be more detailed than before, LGTM.
However, found some minor issues.

contract/r/gnoswap/emission/distribution.gno
onlyhyde: @r3v4s 
The validation for the target is already checked by the assert function and you will call that function. It is not a default and it doesn't seem right to cause panic here because it is a function to convert to a string for the event message.

contract/r/gnoswap/position/position.gno
onlyhyde: Supporting more types will increase the case checking of the switch, so I think it's better to support only the types we currently use. Do you have any additional types you need besides the ones we currently use? I would appreciate it if you could share which types we should use in the future.

r3v4s (dismissed):

contract/r/gnoswap/position/position.gno
r3v4s: Nah, I can't think of any more type that we're willing to use.

r3v4s (dismissed):

dongwon8247 (dismissed):

notJoon (dismissed):

dongwon8247 (approved):

r3v4s (approved):

PR #484, opened 2025-01-22
build: git action CI to test contract

r3v4s:
## Description
1. adds `git action` to run all test file
> each contract tests run parallelly to reduce CI time.
> it always pull latest master branch from gno.
> bit of gnoVM modification due to change block time.

2. gnoswap no longer handles users.AddressOrName for logic
> for tests, there are conversion function that requires this type.

3. fix lots of bug


.github/workflows/run_test.yml
```diff
@@ -0,0 +1,611 @@
...
+    strategy:
+      fail-fast: false
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22
```
notJoon, line 19: to enabling go module cache
```suggestion
          go-version: 1.22
          cache: true
```

onlyhyde (approved):

PR #480, opened 2025-01-21
feat(reward): Make `reward` calculation logics into separate package

notJoon:
# Description

closes #478

PR #485, opened 2025-01-22
feat: Split `Launchpad` Contract into `p`/`r` Pacakges

notJoon:
# Description

WIP


notJoon:
@moul 
I am currently working on separating the contract into p and r packages based on their status. As you mentioned earlier, I've structured the two packages to have identical names and am adding files as needed.

Although this PR still in WIP, could you please check if this is the structure you had in mind? Thank you.

_**cc: @dongwon8247**_

moul:
if you think that the r/launchpad/v2 will be able to reuse most of the p/launchpad, then it could make sense to make this split, but if not, it's probably better to keep most of the logic in a single r/launchpad.

dongwon8247:
@notJoon A reminder: let's extract what's needed for `r/launchpad` from #485 #480, and we don't need to work on separating p/r. Please close the other PR when you start working on it.

PR #488, opened 2025-01-31
GSW-2080 docs: update comments

r3v4s:
## Description
More comments so code can be more easily understand.
Made time related variable name to be more clear.
Clean up non used code.


contract/r/gnoswap/position/position.gno
```diff
@@ -1037,6 +1037,12 @@ func getCurrentFeeGrowth(postion Position, token0, token1 string, fee uint32) (F
 
 // Repositiomn adjusts the position of an existing liquidity token
 // by changing its price range and liquidity amount.
+// It only increase liquidity to existing position without
+// decreasing liquidity even function does reposition.
+// This is because, amount of each token for new position can be
+// different from the amount of each token for previous position.
+// Therefore as describe in the reference docs, we use multi msg to
+// decrease liquidity -> swap -> add liquidity to reposition.
```
dongwon8247, line 1045:
```suggestion
// Reposition adjusts the price range and liquidity of an existing position.
// Reposition is executed via multiple messages as follows:
//
// 1. Decrease the liquidity of the existing position.
// 2. Swap tokens as needed to reach the desired token ratio.
// 3. Add new liquidity with the desired price range and amount.
//
```

notJoon (approved):

contract/p/gnoswap/consts/consts.gno
```diff
@@ -117,17 +110,18 @@ const (
...
+)
 
-	DAY_PER_YEAR = 365
+// BLOCK TIME
+const (
+	// Using this value contract does calculate
+	// how many block will be minted in a certain time range.
+	// However due to chain status, it can be different.
+	// So we use this value as initial value, and update by calling setter through governance.
+	BLOCK_GENERATION_INTERVAL int64 = 2 // seconds
```
dongwon8247, line 124: How about we unify this to milliseconds? I see we're using milliseconds in [SetAvgBlockTimeInMs](https://github.com/gnoswap-labs/gnoswap/blob/a29e0f994466430618548ae992cca11a52f5102a/contract/r/gnoswap/gns/halving.gno#L359)

contract/p/gnoswap/consts/consts.gno
```diff
@@ -117,17 +110,18 @@ const (
...
 )
 
 // ETCs
```
dongwon8247, line 127: I don't think the zero address defined here is used in Gno. The PR is closed. Maybe we should not use this. Where is it needed? any alternatives?

contract/p/gnoswap/consts/consts.gno
dongwon8247: @onlyhyde @notJoon What do you think?

contract/p/gnoswap/consts/consts.gno
```diff
@@ -117,17 +110,18 @@ const (
...
+	SECONDS_PER_DAY = 86400
+)
 
-	DAY_PER_YEAR = 365
+// BLOCK TIME
+const (
+	// Using this value contract does calculate
+	// how many block will be minted in a certain time range.
+	// However due to chain status, it can be different.
+	// So we use this value as initial value, and update by calling setter through governance.
```
dongwon8247, line 123:
```suggestion
	// Default block generation interval, in seconds, used to estimate
	// how many blocks will be produced over a given timeframe.
	// GnoSwap’s emission logic relies on this as an initial value,
	// but actual block times may vary (e.g. due to network conditions).
	// Governance or an admin can adjust it dynamically by calling
	// [SetAvgBlockTimeInMs](https://github.com/gnoswap-labs/gnoswap/blob/a29e0f994466430618548ae992cca11a52f5102a/contract/r/gnoswap/gns/halving.gno#L359) to keep emission schedules accurate.
```

contract/p/gnoswap/consts/consts.gno
```diff
@@ -4,15 +4,10 @@ import (
...
-	TOKEN_REGISTER std.Address = "g1er355fkjksqpdtwmhf5penwa82p0rhqxkkyhk5"
-
-	TOKEN_REGISTER_NAMESPACE string = "gno.land/r/g1er355fkjksqpdtwmhf5penwa82p0rhqxkkyhk5"
-
-	BLOCK_GENERATION_INTERVAL int64 = 2 // seconds
+	ADMIN   std.Address = "g17290cwvmrapvp869xfnhhawa8sm9edpufzat7d"
+	DEV_OPS std.Address = "g1mjvd83nnjee3z2g7683er55me9f09688pd4mj9"
 )
 
 // WRAP & UNWRAP
```
dongwon8247, line 13: Do you think we can use the whitelist system for native tokens as well? @onlyhyde @notJoon @mconcat 
Same as making the contract path/address modifiable with a whitelist system by governance or admin, if we can add/delete native token support, it would be helpful when gno.land enables IBC.

dongwon8247: Discussion only. This is for another PR.

contract/p/gnoswap/consts/consts.gno
r3v4s: We can just use `std.Address("")` as an zero(or empty) address.
Currently we use this when staked position has been unstaked -> we need to set operator of that position to zero address. ( It was idea from @onlyhyde )

contract/p/gnoswap/consts/consts.gno
r3v4s: Perhaps good to unify base maybe, leave your thoughts @onlyhyde @notJoon

contract/p/gnoswap/consts/consts.gno
r3v4s: I think whitelisting not only native token, but also corresponding wrapped token might be needed.. However don't have not sure how we can integrate this.

contract/p/gnoswap/consts/consts.gno
onlyhyde: Unifying the units seems like a better choice.

contract/p/gnoswap/consts/consts.gno
onlyhyde: @r3v4s @dongwon8247 
Values that are valid as an address type, but not actually used, such as a Burn address or a Zero address, are required in the code.  If you use an empty value, the logic breaks down, with errors in some places and meaningful values in others. Our service has the following complexities 
1. once a position is set, it is fixed forever by the owner. (It is not burned, it can be used again)
2. assets can be transferred by operator in addition to owner, so operator management is required. 
3. When staking, the LP of the position belongs to the staker contract, so it can only be transferred by the contract, and the staked owner must be able to collect rewards.
4. When an owner is registered as an operator in GNFT, there is a vulnerability that can ignore the stake contract and steal assets.
5. A management ledger is required to satisfy 3 and 4 above at the same time, and it is necessary to initialize the value registered in this ledger. 
6. The initialization value in 5 is valid as an address type, and it should be an address that is difficult to be created by other users.

This requires something like a zero address or burn address, which also needs to be defined in the protocol for initializing the referral address, so that it is a meaningful address and is necessary to satisfy the condition that it is unusable. If we decide not to use this, we need to completely rethink the logic of the entire contract, and I can't think of any other alternative, so I need additional ideas. Empty values like “” cannot be used as options in Referral contracts, and I don't think it's a good idea to use them in a separate transaction that doesn't use the gnoswap interface because of the difficulty in error validation if the parameter is not entered correctly.

Translated with www.DeepL.com/Translator (free version)

contract/p/gnoswap/consts/consts.gno
onlyhyde: Currently, the logic for handling native coins is hardcoded, but it seems possible to modify it to register, manage, and use denom in the whitelist managed part. However, this would need to be done over a long enough period of time.

dongwon8247 (approved):

dongwon8247 (approved):

contract/r/gnoswap/gns/utils.gno
```diff
@@ -57,7 +58,7 @@ func i64Min(x, y int64) int64 {
 }
 
 func secToMs(sec int64) int64 {
-	return sec * MILLISECONDS_PER_SECOND
+	return sec * consts.MILLISECONDS_PER_SECOND
 }
```
onlyhyde, line 63: Since the default unit is milliseconds, it would be nice to annotate variables that are converted to seconds so that we know whether they are in seconds or milliseconds. 

It would be more readable if a milliToSec function was added and enabled.

onlyhyde (approved):

notJoon (approved):

Issue #496, opened 2025-02-03
Central memory bus architecture


mconcat:
## 1. Introduction and Motivation

To facilitate smoother migrations, emergency bug fixes, and security patches in gnoswap v1, we propose introducing a **central communication bus**. Currently, contracts call one another directly—for example, `staker.MintAndStake()` directly invokes `position.Mint()`—which creates tight coupling and makes coordinated upgrades difficult. By decoupling these interactions via a central bus, we can:

- **Reduce interdependency:** Allow individual contracts to be halted, upgraded, or replaced without breaking the overall system.
- **Enable dynamic upgrades:** Support on-the-fly version transitions (mounted via governance proposals) without the need for full state or asset migration.
- **Improve security and resilience:** Quickly mitigate discovered vulnerabilities while ensuring continued cross-contract communication.

**Note:** This approach is intended for operational continuity (e.g., mitigating bugs or security vulnerabilities) rather than full state or asset migrations.

---

### 2.1 Overview

The central communication bus acts as an intermediary between contracts. Instead of contracts invoking each other’s methods directly, they will register their external entrypoints with the bus. When a contract needs to perform an operation, it will query the bus, which then routes the request to the appropriate contract’s current implementation. This indirection enables seamless upgrades by updating registry entries rather than modifying individual contract logic.

### 2.2 Architectural Components

- **Registry:**  
  The bus maintains a registry mapping entrypoint identifiers (and, potentially, version tags) to the corresponding contract addresses. This registry is updated via a controlled governance process.

- **Routing Mechanism:**  
  When a contract makes a call, it does so by querying the bus. The bus then routes the call to the registered contract. This design encourages stateless, pure function calls that pass all necessary context explicitly, reducing risks associated with stateful interactions.

- **Versioning:**  
  The bus allows for versioned endpoints. When a new version of a contract is approved via a governance proposal, the registry is updated to point to the new version. For the duration of the upgrade, calls are seamlessly redirected without requiring changes to the calling contract.

- **Permissions and Access Controls:**  
  The bus will assume responsibility for enforcing permissions (e.g., ensuring that only authorized callers access certain functionalities), eliminating the need for contracts to reference prior realms or hard-coded permissions.

### 2.3 Operational Flow

1. **Registration:**  
   Upon deployment, each contract registers its entrypoints with the bus. For example, the `position` contract registers its `Mint()` function.

2. **Invocation:**  
[...]

PR #506, opened 2025-02-06
feat(p/rbac): RBAC Pacakge Implementation

notJoon:
# Description

Implemented the RBAC package, but need to discussion whether it should be located in `p` or `r`.

paritally related with #496

### DIFF
--- a/contract/p/gnoswap/rbac/doc.gno
+++ b/contract/p/gnoswap/rbac/doc.gno
@@ -55,9 +55,8 @@
 // import (
 //	"std"
 //
-//	"gno.land/p/demo/ufmt"
 //	"gno.land/p/gnoswap/rbac"
-//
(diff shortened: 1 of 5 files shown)


onlyhyde:
If you organize RBAC as a package, there is no ledger to manage roles and addresses for roles. How do you think it will be used?

notJoon:
> If you organize RBAC as a package, there is no ledger to manage roles and addresses for roles. How do you think it will be used?

I considered a structure where `p` contains only pure implementations, while `r` contains global variable declarations.

For example:

```go
package xxx

import (
    "gno.land/p/demo/rbac"
)

var Manager = rbac.New()

// ...
```

If we move the package to the `r` side, I just need to add the code above directly to the `rbac` package.

onlyhyde (approved):
LGTM

PR #509, opened 2025-02-07
Registry based cross contract communication

mconcat:


notJoon:
implemented this feature (not exactly same) in a different way, so we'll close this PR.

PR #544, opened 2025-02-25
feat: Remove Address Dependencies within Tests

notJoon:
# Description

Modified to get the address of the contract registered in `access` (via `rbac` package) from directly using the address defined in the `consts` package.

Expected to be able to automatically recognize when a contract is upgraded to a different version in the future.

## Minor Changes

- Removed unnecessary (mostly duplicated) constants and variables from test code
- Fixed tests that didn't inspect output values
- Added a new list of addresses that were missing from `access`

NOTE: If paths (e.g. `“gno.land/r/gnoswap/xxx”`) can also be stored via `rbac`, it seems possible to remove dependencies in the same way.


contract/r/gnoswap/access/consts.gno
```diff
@@ -17,8 +19,9 @@ const (
 	ROLE_STAKER     = "staker"
 	ROLE_LAUNCHPAD  = "launchpad"
 	ROLE_EMISSION   = "emission"
-	ROLE_USER       = "user"
+	ROLE_PROTOCOL_FEE = "protocol_fee"
 
+	ROLE_USER       = "user"
```
onlyhyde, line 24: Is a user role necessary?

notJoon: this was a placeholder, so it does not necessary. https://github.com/gnoswap-labs/gnoswap/pull/544/commits/1b5e555c5f9e3dafddbbd5c305df972c0c177ff0

onlyhyde (approved):

onlyhyde (approved):

PR #553, opened 2025-03-01
test: router test

notJoon:
# Description

- Added tests for exact out
- Added validation tests for swap single input values
    - Modified to throw a panic when input and output tokens are the same


contract/r/gnoswap/router/errors.gno
```diff
@@ -20,6 +20,7 @@ var (
 	errWugnotMinimum          = errors.New("[GNOSWAP-ROUTER-011] less than minimum amount	")
 	errQuoteParser            = errors.New("[GNOSWAP-ROUTER-012] quote parse failed")
 	errHopsOutOfRange         = errors.New("[GNOSWAP-ROUTER-013] number of hops must be 1~3")
+	errSameTokenSwap          = errors.New("[GNOSWAP-ROUTER-014] cannot swap same token")
```
dongwon8247, line 23: @notJoon Just to make sure - This doesn't affect the swap (wrapping essentially) between ugnot (native) and wugnot (grc20), right? it's done within the same Swap interface.

dongwon8247 (approved):
One comment. Please check.

contract/r/gnoswap/router/errors.gno
notJoon: need to add test for this case

notJoon: Following the code flow:

1. When creating a pool, the `createPool` function calls the `updateWithWrapping` method
2. Inside `updateWithWrapping`, the `wrap` function automatically converts the native token path to the wrapped token path.
https://github.com/gnoswap-labs/gnoswap/blob/d0a5749e8f60b7751b5e9663f300998b55eaeb8c/contract/r/gnoswap/pool/pool_factory_param.gno#L67-L68
3. In the `wrap` function, if the token path `GNOT` (native), it automatically changes to `WUGNOT` (wrapped)
https://github.com/gnoswap-labs/gnoswap/blob/d0a5749e8f60b7751b5e9663f300998b55eaeb8c/contract/r/gnoswap/pool/pool_factory_param.gno#L104-L112

Therefore, the swap between ugnot and wugnot should not be a problem because the wrapping process is automated internally.

dongwon8247 (approved):

PR #558, opened 2025-03-04
feat: Emission Callback

### LINKED ISSUE #542: Import Address Recognition Issue in Staker Contract
# Description

The staker contract currently recognizes emission addresses through imports, which means it only recognizes the previous emission contract even when the contract changes.
https://github.com/gnoswap-labs/gnoswap/blob/cb8f66329434c6a4ec39726ef19d5a1d62b36883/contract/r/gnoswap/staker/staker.gno#L196-L199
This creates a potential issue where updates to the emission contract may not be properly reflected.

## Current Implementation
- The staker contract imports emission addresses directly
- Contract changes are not dynamically recognized
- Creates tight coupling between staker and emission contracts

## Proposed Solution
Consider implementing a callback-based approach to:
1. Make the contract relationship more dynamic
2. Avoid direct dependencies through imports
3. Allow for better contract upgradability

## Additional Considerations
Need to find a better architecture that:
- Prevents circular dependencies
- Maintains loose coupling between contracts
- Allows for dynamic contract updates

notJoon:
# Description

closes #542


contract/r/gnoswap/position/emission.gno
```diff
@@ -0,0 +1,20 @@
+package position
+
+import (
+	br "gno.land/r/gnoswap/v1/brigde"
+)
+
```
onlyhyde, line 6: When redeploying due to contract upgrade, please add a comment on the structure of registering and calling the callback function in a bridge structure to avoid import issues.

notJoon: https://github.com/gnoswap-labs/gnoswap/pull/558/commits/e96b66b43263c2c7ac936269d11c288ba41f248b

onlyhyde (approved):

PR #560, opened 2025-03-06
feat(access): Applying Permission Checks when Editing Roles

notJoon:
# Description

Adds admin permission checks to role management functions to ensure that only admin can modify role configurations. This is a security enhancement to prevent unauthorized role modifications.

Additionally, I've also added a function to remove RBAC-registered roles.

## Changes

- `SetRole`
- `UpdateRoleAddress`
- `CreateRole`
- `RemoveRole` (RBAC)

## Testing

All tests have been updated and verified to ensure:
- Admin can successfully modify roles
- Non-admin cannot modify roles
- Appropriate error messages are returned for unauthorized attempts
- Existing functionality remains unchanged when called by admin


contract/p/gnoswap/rbac/rbac.gno
```diff
@@ -34,6 +34,52 @@ func (rb *RBAC) RegisterRole(roleName string) error {
 	return nil
 }
 
+func (rb *RBAC) RemoveRole(roleName string) error {
```
onlyhyde, line 37: The current implementation of access is such that only one address is applied to the admin role. If you set the admin role to the wrong address here, you will lose control of the system. We need a way to prevent the admin role address from being set incorrectly, and a function that removes only those addresses where the role exists but the role has permissions. 
Another way is to manage the owner address that owns the access contract separately and allow the owner address to recover when the owner address is incorrectly registered for the admin role. 
The management of roles and the management of addresses with roles should be separated.

onlyhyde (dismissed):

onlyhyde (approved):

PR #587, opened 2025-03-27
fix: when input token is native and exact-out type, swap fail

onlyhyde:
Fix for the case where the input token is gnot and the swap is of the exact-out type, and the swap fails due to an asset shortage error because the native cannot be wrapped


notJoon (approved):
👍

PR #589, opened 2025-03-28
fix: Early withdrawals in `CollectDepositGnsByDepositId`

notJoon:
# Description

The `CollectDepositGnsByDepositId` function was allowing users to withdraw their deposits prematurely by only checking `claimableHeight`. This was problematic because claimable duration (for collecting rewards) is much shorter than the fixed tier duration.

Changes:
- Added validation in `validateDepositCollection` to ensure deposits can only be withdrawn after the full tier period (30/90/180 days) has ended
- Added `BLOCKS_PER_DAY` constant based on block generation interval
- Added test cases to verify both successful withdrawal after tier period and failed withdrawal before tier period ends

This ensures deposits remain locked for the entire tier duration as intended, preventing early withdrawals.

### DIFF
--- a/contract/r/gnoswap/launchpad/launchpad_deposit.gno
+++ b/contract/r/gnoswap/launchpad/launchpad_deposit.gno
@@ -18,6 +18,12 @@ import (
 	"gno.land/r/gnoswap/v1/referral"
 )
 
+const (
+	// BLOCKS_PER_DAY = SECONDS_PER_DAY * (MILLISECONDS_PER_SECOND / BLOCK_GENERATION_INTERVAL)
+	// = 86400 * (1000 / 2000) = 43200
+	BLOCKS_PER_DAY uint64 = uint64(consts.SECONDS_PER_DAY * consts.MILLISECONDS_PER_SECOND / consts.BLOCK_GENERATION_INTERVAL)
+)
+
 var launchpadAddr, _ = access.GetAddress(access.ROLE_LAUNCHPAD)
 
 var (
@@ -47,8 +53,6 @@ type DepositState struct {
 	DepositsByUserProject map[std.Address]map[string][]string
 }
 
-// TODO:
-// 1. 주석 추가
 // ProjectTierInfo contains information about a project tier
 type ProjectTierInfo struct {
 	Project       Project
@@ -883,6 +887,20 @@ func validateDepositCollection(depositId string, caller std.Address) (Deposit, P
 	}
 
 	currentHeight := uint64(std.ChainHeight())
+
+	tierDuration, err := strconv.ParseUint(deposit.tier, 10, 64)
+	if err != nil {
+		return Deposit{}, Project{}, errors.New("invalid tier duration")
+	}
+
+	// calculate tier end height (deposit height + tier duration in blocks)
+	tierEndHeight := deposit.depositHeight + (tierDuration * BLOCKS_PER_DAY)
+
+	// to prevent the withdrawal of deposits before the actual tier period ends,
+	// we need to verify that deposits can only be withdrawn after the tier period has completely ended.
+	if currentHeight < tierEndHeight {
+		return Deposit{}, Project{}, errors.New("tier lock period has not ended yet")
+	}
 
 	if !isPassedClaimableHeight(deposit, currentHeight) {
 		return Deposit{}, Project{}, errors.New("not passed claimable block height yet")
--- a/contract/r/gnoswap/launchpad/launchpad_deposit_test.gno
+++ b/contract/r/gnoswap/launchpad/launchpad_deposit_test.gno
@@ -2,6 +2,7 @@ package launchpad
 
 import (
 	"std"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


contract/r/gnoswap/launchpad/launchpad_deposit.gno
```diff
@@ -868,39 +875,70 @@ func validateDepositCollection(depositId string, caller std.Address) (Deposit, P
...
+// validateTierPeriod checks if the tier lock period has ended
+func validateTierPeriod(deposit Deposit, currentHeight uint64) error {
+	tierDuration, err := strconv.ParseUint(deposit.tier, 10, 64)
+	if err != nil {
+		return errors.New("invalid tier duration")
+	}
+
+	// To prevent the withdrawal of deposits before the actual tier period ends,
+	// we need to verify that deposits can only be withdrawn after the tier period has completely ended.
+	tierEndHeight := deposit.depositHeight + (tierDuration * BLOCKS_PER_DAY)
```
onlyhyde, line 932: There is an end time for each tier. It would be more accurate to use that value.

contract/r/gnoswap/launchpad/launchpad_deposit_test.gno
```diff
@@ -765,12 +766,65 @@ func TestCollectDepositGnsByDepositId_ExactClaimable(t *testing.T) {
 		1000,
 	)
 
+	// Skip to the end of tier period (30 days)
+	tierDuration := uint64(30)
+	blocksToSkip := (tierDuration * BLOCKS_PER_DAY) + 1 // +1 to ensure we're past the tier period
```
onlyhyde, line 771: I suggest that you test it using the endBlockHeight of the tier rather than the duration.

onlyhyde (changes requested):

contract/r/gnoswap/launchpad/launchpad_deposit.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/589/commits/cd31201b17e5b04ed30fe18314fa84821e806410

contract/r/gnoswap/launchpad/launchpad_deposit_test.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/589/commits/cd31201b17e5b04ed30fe18314fa84821e806410

contract/r/gnoswap/launchpad/launchpad_deposit.gno
```diff
@@ -18,6 +18,12 @@ import (
 	"gno.land/r/gnoswap/v1/referral"
 )
 
+const (
```
dongwon8247, line 21: You can probably drop this const since not being used?

dongwon8247, line 21: You can probably drop this const since not being used?

notJoon: https://github.com/gnoswap-labs/gnoswap/pull/589/commits/ecfcc1fb08a4cd239671b55e56433137a995aca2

dongwon8247 (approved):

jinoosss:
The `CollectDepositGnsByProjectId` function should also be validated.

contract/r/gnoswap/launchpad/launchpad_deposit.gno
```diff
@@ -291,6 +279,14 @@ func CollectDepositGnsByProjectId(projectId string) uint64 {
 			errNotExistDeposit, ufmt.Sprintf("no deposit list for project(%s)", projectId)))
 	}
 
+	// Validate each deposit to prevent early withdrawal
+	for _, depositId := range depositList {
```
jinoosss, line 283: We are already iterate the deposit list in the `processDepositCollection` function. 
Do we need to iterate over the deposit list here?

contract/r/gnoswap/launchpad/launchpad_deposit.gno
```diff
@@ -291,6 +279,14 @@ func CollectDepositGnsByProjectId(projectId string) uint64 {
...
+		_, _, err := validateDepositCollection(depositId, caller)
+		if err != nil {
+			panic(addDetailToError(errInvalidInput, err.Error()))
```
jinoosss, line 286: Items that validate successfully should collect.
If there is any item that cannot be collected, a panic will occur and it will fail.

jinoosss (commented):
I left a few comments.

contract/r/gnoswap/launchpad/launchpad_deposit.gno
notJoon: When conducting checks within the loop of the `processDepositCollection` function, it allows for a kind of partial success where only failed deposits are skipped while the rest are processed.

This structure was used because when some deposits are processed, their status changes, which can cause atomicity-related issues.

notJoon: https://github.com/gnoswap-labs/gnoswap/pull/589/commits/1661a17610fcf2fc9780673fff3dd3d98f9875c9

notJoon:
no longer valid for current structure. closing now

PR #596, opened 2025-04-04
GSW-2214 feat: Add Overflow Protection to `DrySwapRoute`

notJoon:
## Description

In the dry swap, may overflow in the following two cases:

1. During quote calculation in `calculateSwapAmount`
2. When accumulating results in `resultAmountIn` and `resultAmountOut`

## Changes

- Added `IsOverflow` function to `int256` package
- Added overflow check in `calculateSwapAmount` and `DrySwapRoute` functions in the `router_dry.gno` file
- Added related tests
    - Need to run separately because it creates the pool within the test
- Refactored `router.gno` file


dongwon8247 (approved):

jinoosss (approved):
👍

PR #597, opened 2025-04-07
GSW-2216 fix: Unable to create Full Range position in 0.01% fee pool

notJoon:
## Description

Fixes an issue where users cannot create Full Range positions in pools with 0.01% fee tier. The problem was in the `assertValidTickUpper` function in `tick.gno`, which incorrectly rejected valid tick upper values equal to `MAX_TICK`.

## Changes

- `tickUpper >= consts.MAX_TICK` -> `tickUpper > consts.MAX_TICK`


jinoosss (approved):
👍

PR #598, opened 2025-04-07
GSW-2213 fix: Slippage error in ExactOut

notJoon:
# Description

The `ExactOutSwap` operation is failing with a slippage check error. The error occurs when there is a difference of just `1` unit between the requested output amount and the actual output amount.

```plain
[GNOSWAP-ROUTER-002] slippage check failed || Received more than requested in [EXACT_OUT] requested=600901352, actual=600901351
```

## Root Cause

The issue occurs from the precision handling in the `finalizeSwap` function. In the `ExactOut` swap type, the function checks if `resultAmountOut` is less than `amountSpecified`:

```go
if resultAmountOut.Lt(amountSpecified) {
    panic(addDetailToError(
        errSlippage,
        ufmt.Sprintf("Received more than requested in [EXACT_OUT] requested=%s, actual=%s", amountSpecified.ToString(), resultAmountOut.ToString()),
    ))
}
```

This check fails when there is even a 1-unit difference between the requested and actual amounts. Normally, in big number arithmetic, can lead to small differences due to rounding or truncation during calculations.


jinoosss (approved):
LGTM !

PR #599, opened 2025-04-08
refactor: router contract

notJoon:
# Description

- Modified the `DrySwapRouter` implementation to reuse existing functions.
- Optimized to maximize object reuse within safe boundaries.
- Updated `MustRegistered` function to accept multiple parameters at once, reducing the number of function calls.
- Extracted interfaces from `singleSwap` and `singleDrySwap` functions to reduce duplication.
- Added an `Unknown` type to `SwapType` for error handling. This type is only used conceptually and is not actually implemented.
- Applied "Interface Consolidation" pattern to `SingleSwapParams` and `SwapParams` types
- Added a `SwapParamsI` interface to modify the recognition of `SwapParams` and `SingleSwapParams` types when creating a `SwapCallbackData` object

### DIFF
--- a/contract/p/gnoswap/int256/cmp.gno
+++ b/contract/p/gnoswap/int256/cmp.gno
@@ -99,3 +99,7 @@ func (z *Int) Gt(x *Int) bool {
 func (z *Int) Clone() *Int {
 	return &Int{z.abs.Clone(), z.neg}
 }
+
+func (z *Int) IsOverflow() bool {
+	return z.abs.IsOverflow()
+}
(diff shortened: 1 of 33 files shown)


jinoosss (approved):
The logic looks good, so I approve it 👍 

However, I have a question about your refactoring work.
Is there a reason for the distinction between SwapExecutor (single swap) and MultiSwapExecutor?
Both of them use swapInner or swapDryInner.

notJoon:
> However, I have a question about your refactoring work. Is there a reason for the distinction between SwapExecutor (single swap) and MultiSwapExecutor? Both of them use swapInner or swapDryInner.

@jinoosss Thanks for the review. I thought it would be more explicit to divide it into separate types. Also, it was difficult to abstract the required parameters and flow into one.

PR #602, opened 2025-04-15
fix: Correctly Emit Event Parameters in Referral Registration Function

notJoon:
# Description


jinoosss (approved):
It looks good.

I have one comment. When using the `TryRegister` function, the following pattern is repeated.
```go
success := referral.TryRegister(caller, referrer)
actualReferrer := referrer
if !success {
	actualReferrer = referral.GetReferral(caller.String())
}
```

Why not respond with the Referrer address registered in `TryRegister`?
Semantically, it doesn't violate the single responsibility principle.

notJoon:
> Why not respond with the Referrer address registered in `TryRegister`? Semantically, it doesn't violate the single responsibility principle.

I did consider that approach, but I decided to handle it explicitly because the TryRegister function would be doing too many things and might appear to be operating like a black box.

PR #605, opened 2025-04-18
feat: Emit ticks information from Swap events

notJoon:
### Enhancements to tick data handling:

* **Added `Ticks` method to `Pool` struct**: This method iterates over the `ticks` AVL tree, converts each tick's data into a JSON-like string format, and returns a concatenated string representation of all ticks. (`contract/r/gnoswap/pool/pool_type.gno`, [contract/r/gnoswap/pool/pool_type.gnoR79-R102](diffhunk://#diff-aff118d48fc271a85312e9706162cf06f2b8fd3a4e2b687f2a8286700c078b4bR79-R102))

* **Updated `Swap` function to include tick data**: The `Swap` function now calls the new `Ticks` method and includes the resulting tick data in its output. (`contract/r/gnoswap/pool/swap.gno`, [contract/r/gnoswap/pool/swap.gnoR133](diffhunk://#diff-2e80ee12d605a9afabc1424c74775545c4dfa7e0c7b7999a368069aa65b6a771R133))

### DIFF
--- a/contract/r/gnoswap/pool/pool_type.gno
+++ b/contract/r/gnoswap/pool/pool_type.gno
@@ -1,7 +1,11 @@
 package pool
 
 import (
+	"strconv"
+	"strings"
+
 	"gno.land/p/demo/avl"
+	"gno.land/p/demo/ufmt"
 	u256 "gno.land/p/gnoswap/uint256"
 )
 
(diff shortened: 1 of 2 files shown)


jinoosss (approved):
LGTM 👍 

Right now, we're giving down all ticks that branch, but you may want to configure it for past ticks only.

I'll check further.

PR #606, opened 2025-04-18
GSW-2258 fix: Referral does not propoerly removed

notJoon:


contract/r/gnoswap/referral/keeper.gno
```diff
@@ -68,11 +66,14 @@ func (k *keeper) setReferral(addr, refAddr std.Address, eventType string) error
...
-			EventTypeRemove,
-			"removedAddress", addr.String(),
-		)
-		return k.remove(addr)
+		if k.has(addr) {
+			err := k.remove(addr)
+			if err != nil {
+				return err
+			}
+			return nil
```
jinoosss, line 74: This is the same as this.
`return k.remove(addr)`

contract/r/gnoswap/referral/keeper.gno
```diff
@@ -109,16 +108,20 @@ func (k *keeper) remove(target std.Address) error {
 		return ErrInvalidAddress
 	}
 
-	if err := k.checkRateLimit(target.String()); err != nil {
+	tgt := target.String()
+
+	if err := k.checkRateLimit(tgt); err != nil {
```
jinoosss, line 113: Why are the checks different for set and remove?

jinoosss (commented):
I left a few comments.

contract/r/gnoswap/referral/keeper.gno
notJoon: The `remove` function is only used in the `setReferral`, and since checks are already done in the set function, there was no need for additional checks, which is why `validateAddresses` was excluded. If I modify it for consistency, I could add a line like this:

```go
if err := k.validateAddresses(target, zeroAddress); err != nil {
	return err
}
```

How do you think?

contract/r/gnoswap/referral/keeper.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/606/commits/97b425103e6d0387ebaac02d05c47568eefb0768

contract/r/gnoswap/referral/keeper.gno
jinoosss: I didn't mean `validateAddress`.

The `checkRateLimit` function is inside the remove function.
If you look at the contents of the function, it checks for the minimum time to change the referral and updates the last change time.

This logic is not being executed in the update logic for non-zero addresses.

notJoon: Oh I missed this one. thanks
https://github.com/gnoswap-labs/gnoswap/pull/606/commits/c5eefe56697cfe8512d85a0151b779537500ddc6

jinoosss (approved):

PR #607, opened 2025-04-18
test(swap): swap scenarios for tick changes

jinoosss:
### Add swap test scenarios for tick boundary crossing

This PR adds comprehensive test scenarios for swap operations when crossing tick boundaries. The implementation includes:

- Testing swap functionality with positions at all possible tick levels (-887272 to 0)
- Setting up test infrastructure for verifying pool liquidity changes during tick crossings
- Adding validation points for fee growth calculations at tick boundaries
- Implementing test cases for tracking claimable fees across positions after swaps


jinoosss:
When the tick of the pool is -887272, 
an error occurs in creating positions in the range of -887272 to 0 ticks.

A zero liquidity error is thrown and it needs to be verified that this is normal behavior.

cc @dongwon8247 @notJoon

contract/r/gnoswap/pool/swap_cross_ticks_test.gno
```diff
@@ -0,0 +1,107 @@
...
+	tick := int32(-887272)
+	token0Path := barPath
+	token1Path := bazPath
+
+	SetPoolCreationFeeByAdmin(0)
+
+	TokenApprove(t, token0Path, adminAddr, poolAddr, uint64(100_000_000_000_000))
+	TokenApprove(t, token1Path, adminAddr, poolAddr, uint64(100_000_000_000_000))
+
+	CreatePool(token0Path, token1Path, FeeTier100, "4295128740")
```
notJoon, line 67: This must be an expected behaviour in concentrated liquidity model, where posiitions must contain the current price to be active.

1. When a pool created at tick `-887272`, it's at the absolute minimum prioce
2. creatinf positions from `-887272` to `0` would not include the current price point

So, we need to ensures positions include the current proce and have liquidity.

```suggestion
	CreatePool(token0Path, token1Path, FeeTier100, common.TickMathGetSqrtRatioAtTick(0).ToString())
```

This fix would work.

dongwon8247:
@jinoosss please check if it's resolved. We can add it as a test.

PR #612, opened 2025-04-22
GSW-2254 fix(pool): Improve Price Calculation Precision in Pool Creation

notJoon:
# Description

Replaces the tick-based price inversion method with direct mathematical inversion in the pool factory parameter calculation to eliminate precision loss.

## Problem

The previous implementation used a tick-based approach to calculate the inverse price:
1. Convert `sqrtPriceX96` to tick
2. Negate the tick
3. Convert back to `sqrtPriceX96`

This method introduced precision loss due to tick discretization, as ticks correspond to specific price points defined by the formula $Price = 1.0001^{tick}$. The rounding that occurs during these conversions could lead to deviations from the true inverse price, particularly significant at extreme price levels.

## Solution

Implemented a direct mathematical inversion approach using the formula:

$$
new = \frac{2^{192}}{old}
$$

Where, $new$ and $old$ follows `sqrtPriceX96` forms.

## Note

The value `4295128738` is the minimum square root price allowed in the Uniswap V3 protocol[^1]. Therefore, the max range check test has been changed to this value.

[^1]: https://www.rareskills.io/post/uniswap-v3-min-max-tick


notJoon:
TODO: add precision loss case

contract/r/gnoswap/pool/pool_factory_param.gno
```diff
@@ -44,47 +44,48 @@ func newPoolParams(
...
-}
-
-func (p *createPoolParams) TickSpacing() int32 {
-	return p.tickSpacing
-}
-
-func (p *createPoolParams) SqrtPriceX96() *u256.Uint {
-	return p.sqrtPriceX96
-}
+func (p *createPoolParams) SqrtPriceX96() *u256.Uint { return p.sqrtPriceX96 }
```
kjeom, line 47: How do you think naming `createPoolParams` as poolParams. It looks like creat function in factory, not struct.

notJoon: hmm.. still ambiguous. how about `poolCreateConfig` or `poolSpecification`?

contract/r/gnoswap/pool/pool_factory_param.gno
```diff
@@ -44,47 +44,48 @@ func newPoolParams(
...
+				errOutOfRange,
+				ufmt.Sprintf("sqrtPriceX96(%s) is out of range", p.sqrtPriceX96.ToString()),
 			))
 		}
-		newPrice := common.TickMathGetSqrtRatioAtTick(oldTick * int32(-1))
+
+		// newPrice = 2^192 / oldPrice
+		newPrice := u256.Zero().Div(Q192, p.sqrtPriceX96)
+
 		if newPrice.IsZero() {
```
kjeom, line 73: Doesn't it need anymore? because we are checking the range more than least value, `minSqrtRatio`.

kjeom: It couldn't be zero after the range checking.

contract/r/gnoswap/pool/pool_factory_param.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/612/commits/abb45ef9c1e5e3b3b0f3449586ede6958bc64254

contract/r/gnoswap/pool/pool_factory_param.gno
notJoon: removed https://github.com/gnoswap-labs/gnoswap/pull/612/commits/6cb6ff42cab5721f8181dc4a813de98e25d68d27

kjeom (approved):

PR #616, opened 2025-04-25
GSW-2258 fix: referral time checks

notJoon:
# Description

The existing implementation had a condition that required waiting for 24 hours before attempting to register a referral again, even when the previous attempt had failed.

So, modified the code to store the success status along with the referral information for each address. Now, the time check is only performed when the previous referral registration was successful; otherwise, the check is skipped.

Additionally, I added an error field to the events emitted when the `TryRegister` function fails, so that the specific reason for failure is also output.

For example:

```json
[
  {
    "type": "ReferralRegistrationFailed",
    "attrs": [
      {
        "key": "address",
        "value": "g1weskc6tyx9047h6lta047h6lta047h6lwtjkk9"
      },
      {
        "key": "error",
        "value": "invalid address format"
      }
    ],
    "pkg_path": "gno.land/r/gnoswap/v1/referral",
    "func": "TryRegister"
  },
  {
    "type": "RegisterReferral",
    "attrs": [
      {
        "key": "myAddress",
        "value": "g1weskc6tyx9047h6lta047h6lta047h6lwtjkk9"
      },
      {
        "key": "refAddress",
        "value": "g1weskc6tyxf047h6lta047h6lta047h6lxjwum9"
      }
    ],
    "pkg_path": "gno.land/r/gnoswap/v1/referral",
    "func": "setReferral"
  },
  {
    "type": "ReferralRegistrationSuccess",
    "attrs": [
      {
        "key": "address",
        "value": "g1weskc6tyx9047h6lta047h6lta047h6lwtjkk9"
      },
      {
        "key": "referral",
        "value": "g1weskc6tyxf047h6lta047h6lta047h6lxjwum9"
      }
    ],
    "pkg_path": "gno.land/r/gnoswap/v1/referral",
    "func": "TryRegister"
  }
]
```

### DIFF
--- a/contract/r/gnoswap/referral/global_keeper.gno
+++ b/contract/r/gnoswap/referral/global_keeper.gno
@@ -58,6 +58,7 @@ func TryRegister(addr std.Address, referral string) bool {
 		std.Emit(
 			EventRegisterFailed,
 			"address", addr.String(),
+			"error", err.Error(),
 		)
 		return false
 	}
--- a/contract/r/gnoswap/referral/keeper.gno
+++ b/contract/r/gnoswap/referral/keeper.gno
@@ -63,17 +63,27 @@ func (k *keeper) setReferral(addr, refAddr std.Address, eventType string) error 
 			return k.remove(addr)
 		}
 		refAddr = std.Address("")
-	} else {
-		if err := k.checkRateLimit(addr.String()); err != nil {
-			return err
-		}
 	}
 
-	k.store.Set(addr.String(), refAddr.String())
+	addrStr := addr.String()
+	refAddrStr := refAddr.String()
+	prevVal, prevExists := k.store.Get(addrStr)
+
+	k.store.Set(addrStr, refAddrStr)
+	if err := k.checkRateLimit(addrStr); err != nil {
+		// rollback: if there is a previous value, restore it, otherwise delete it
+		if prevExists {
+			k.store.Set(addrStr, prevVal)
+		} else {
+			k.store.Remove(addrStr)
+		}
+		return err
+	}
+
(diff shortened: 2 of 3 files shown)


jinoosss (commented):
It doesn't make sense to manage lastSuccess, we're just checking for the first run.

In my opinion, the only meaningful fixes are the two below.
I think two lines should be enough to fix the issue.
- https://github.com/gnoswap-labs/gnoswap/pull/616/files#diff-6cf89664f4ed4a5f2aa0395fe09e78e8cfc138723e319ed513f6cedad5630e65R82
- https://github.com/gnoswap-labs/gnoswap/pull/616/files#diff-6cf89664f4ed4a5f2aa0395fe09e78e8cfc138723e319ed513f6cedad5630e65L171

notJoon:
> It doesn't make sense to manage lastSuccess, we're just checking for the first run.
> 
> In my opinion, the only meaningful fixes are the two below. I think two lines should be enough to fix the issue.
> 
>     * https://github.com/gnoswap-labs/gnoswap/pull/616/files#diff-6cf89664f4ed4a5f2aa0395fe09e78e8cfc138723e319ed513f6cedad5630e65R82
> 
>     * https://github.com/gnoswap-labs/gnoswap/pull/616/files#diff-6cf89664f4ed4a5f2aa0395fe09e78e8cfc138723e319ed513f6cedad5630e65L171

@jinoosss 
fixed https://github.com/gnoswap-labs/gnoswap/pull/616/commits/3afd6a4f96151795c45bc687ade5cfba0f692140

dongwon8247 (approved):

jinoosss (approved):

PR #624, opened 2025-05-02
fix(governance): Proposal Archiving

notJoon:
# Description

Archive proposals that are not in an active state in a separate tree


contract/r/gnoswap/gov/governance/vote_test.gno
```diff
@@ -182,7 +183,14 @@ func TestVote(t *testing.T) {
 			func() {
 				defer func() {
 					if r := recover(); r != nil {
-						err = r.(error)
+						switch v := r.(type) {
```
kjeom, line 186: Could you let me know why thus logic was added?

kjeom:
@notJoon If proposalState doesn't be change, shouldn't gnoA be maintained?

contract/r/gnoswap/gov/governance/vote_test.gno
notJoon: @kjeom This change was to handle a type error in the `error` type by ensuring it's checked properly. Therefore, this change is necessary.

notJoon:
> @notJoon If proposalState doesn't be change, shouldn't gnoA be maintained?

@kjeom 
Yep. I modified it to check for archived proposals. https://github.com/gnoswap-labs/gnoswap/pull/624/commits/38d4d51b6ee00e471a1de66612fcf76d2809c85b

contract/r/gnoswap/gov/governance/proposal.gno
```diff
@@ -314,27 +315,61 @@ type proposalStateUpdater struct {
...
+		// We will keep the active proposals.
+		// If a proposal is terminated, it will be moved to archivedProposals.
+		if isProposalTerminated(*updater.proposal) {
+			archivedProposals.Set(key, *updater.proposal)
+			proposals.Remove(key)
+		} else {
+			// Keep active proposals
 			proposals.Set(key, *updater.proposal)
 		}
 		return false
```
dongwon8247, line 344:
```suggestion
		return true
```

In `updateProposalsState`, the `Iterate` callback returns `false`, which tells the AVL tree to stop iterating—so only the very first proposal ever gets processed. Change the callback to return `true` so the iteration continues through all entries.

dongwon8247, line 344: Is it possible that calling `proposals.Remove` or `proposals.Set` inside the same iteration can lead to skipped nodes or unpredictable behavior? if so, you can collect all keys to archive, then perform `Remove`/`Set` outside of the iterator.

contract/r/gnoswap/gov/governance/proposal.gno
```diff
@@ -314,27 +315,61 @@ type proposalStateUpdater struct {
...
 			updater.updateExecutionState()
+		}
 
+		// Handle terminated proposals
+		//
+		// We will keep the active proposals.
+		// If a proposal is terminated, it will be moved to archivedProposals.
+		if isProposalTerminated(*updater.proposal) {
+			archivedProposals.Set(key, *updater.proposal)
+			proposals.Remove(key)
```
dongwon8247, line 339:
```suggestion
			proposals.Remove(key)
			// also remove from proposer index
			archivedVal, _ := archivedProposals.Get(key)
			proposer := archivedVal.(ProposalInfo).Proposer.String()
			proposalIdListByProposer.Remove(proposer)
```

When a proposal is archived, its entry in `proposalIdListByProposer` remains, so proposers may be blocked from creating new proposals indefinitely. After moving to `archivedProposals`, also remove the proposer-to-ID mapping.

dongwon8247 (changes requested):

dongwon8247:
@notJoon Please apply suggestions which only make sense.

Additionally,
1. `shouldUpdate` only checks `Created`, `Canceled`, and `Executed`, so proposals that are `Rejected` or `Expired` (but not Executed) still get re-entered into the update logic. (https://github.com/gnoswap-labs/gnoswap/pull/624/files#diff-a39e1c07100b3b9f9227cec71ecd8f72daa81a283829ea15cfc24195e76f2cc3R384-R388) 

```go
func (u *proposalStateUpdater) shouldUpdate() bool {
-   return u.proposal.State.Created &&
-       !u.proposal.State.Canceled &&
-       !u.proposal.State.Executed
+   // only update if not in a terminal state
+   return !isProposalTerminated(*u.proposal) &&
+       (u.proposal.State.Upcoming || u.proposal.State.Active || u.proposal.State.Passed)
```

2. Each proposer can have only one active proposal, but there’s no guard against duplicates in `ProposeText`, etc. Add a checker before mint and updateState?

```go
func ProposeText(...) uint64 {
-   en.MintAndDistributeGns()
+   // prevent duplicate active proposal
+   if existingAny, exists := proposalIdListByProposer.Get(proposer.String()); exists {
+       existingId := existingAny.(uint64)
+       if _, active := proposals.Get(formatUint(existingId)); active {
+           panic("proposer already has an active proposal: " + formatUint(existingId))
+       }
+   }
+   en.MintAndDistributeGns()
```

3. Both p/gnoswap/halt and r/gnoswap/v1/halt are imported. Can we unify?

```go
-import phalt "gno.land/p/gnoswap/halt"
-import "gno.land/r/gnoswap/v1/halt"
+import halt "gno.land/r/gnoswap/v1/halt"
```

contract/r/gnoswap/gov/governance/proposal.gno
notJoon: to ensuring the atomicity, this line must not be changed.

notJoon:
> Please apply suggestions which only make sense.
@dongwon8247 
1. applied. https://github.com/gnoswap-labs/gnoswap/pull/624/commits/c91e49598fbdc349145906b2a7c97d491a9073fb
2. agree. but, if we add them this'll need to modify many other tests (to isolate states), which would increase the size of the PR. so it seems better to create a separate PR for this.
3. not possible.

dongwon8247 (approved):

PR #625, opened 2025-05-02
fix(position): `PositionIsInRange` range check

notJoon:
# Description

In Uniswap V3, a position should no longer be considered active when the price moves to or beyond its upper tick boundary. However, the current implementation uses `<= position.tickUpper` for range checking, which incorrectly includes the upper boundary tick in the active range.

- Modified the range check condition from `<= position.tickUpper` to `< position.tickUpper`
- This ensures that positions are correctly marked as inactive once the current tick reaches the upper boundary

### DIFF
--- a/contract/r/gnoswap/position/getter.gno
+++ b/contract/r/gnoswap/position/getter.gno
@@ -63,10 +63,7 @@ func PositionIsInRange(positionId uint64) bool {
 	poolPath := position.poolKey
 	poolCurrentTick := pl.PoolGetSlot0Tick(poolPath)
 
-	if position.tickLower <= poolCurrentTick && poolCurrentTick <= position.tickUpper {
-		return true
-	}
-	return false
+	return position.tickLower <= poolCurrentTick && poolCurrentTick < position.tickUpper
 }
 
 func PositionGetPositionOwner(positionId uint64) std.Address {


contract/r/gnoswap/position/getter.gno
```diff
@@ -63,10 +63,7 @@ func PositionIsInRange(positionId uint64) bool {
 	poolPath := position.poolKey
 	poolCurrentTick := pl.PoolGetSlot0Tick(poolPath)
 
-	if position.tickLower <= poolCurrentTick && poolCurrentTick <= position.tickUpper {
-		return true
-	}
-	return false
```
dongwon8247, line 69: Is it okay to remove this?

dongwon8247 (approved):

dongwon8247:
@notJoon can you please double check if any of the existing logic includes `currentTick == tickUpper` ? if so, they need to be changed according to this change.

notJoon:
> can you please double check if any of the existing logic includes `currentTick == tickUpper` ? if so, they need to be changed according to this change.

@dongwon8247 
I confirmed that `tickUpper` is only used as condtion in here. All other parts of the usage, it's just used as a value. Therefore, there are no additional changes needed.

PR #626, opened 2025-05-02
fix(staker): Add owner verification in `UnStakeToken`

notJoon:
# Description

The `UnStakerToken` allowed any caller to unstake tokens without verifying if they are the actual owner of the deposit. This could potentially lead to unauthorized withdrawal of staked tokens.

Thus, Added explicit owner verification in the `UnStakeToken` function.

### DIFF
--- a/contract/r/gnoswap/staker/staker.gno
+++ b/contract/r/gnoswap/staker/staker.gno
@@ -1,7 +1,6 @@
 package staker
 
 import (
-	"errors"
 	"std"
 	"time"
 
--- a/contract/r/gnoswap/staker/staker.gno
+++ b/contract/r/gnoswap/staker/staker.gno
@@ -1,6 +1,7 @@
 package staker
 
 import (
+	"errors"
 	"std"
 	"time"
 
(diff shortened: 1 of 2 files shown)


contract/r/gnoswap/staker/staker.gno
```diff
@@ -552,8 +552,19 @@ func CollectReward(positionId uint64, unwrapResult bool) (string, string, map[st
 func UnStakeToken(positionId uint64, unwrapResult bool) (string, string, string) { // poolPath, token0Amount, token1Amount
 	assertOnlyNotHalted()
 
-	// unStaked status
+	// onwer check
 	deposit := deposits.Get(positionId)
+	caller := getPrevAddr()
+	if deposit.owner != caller {
```
kjeom, line 558: Please add a unit test or leave a comment if not needed.

notJoon: left the comment because the checking condition is clear.

Also, this function is difficult to write unit tests for, it seems more appropriate to test it in a live environment.
https://github.com/gnoswap-labs/gnoswap/pull/626/commits/9cc8da30d45945aff3bb831fa5a71efebf7249de

kjeom (approved):

PR #629, opened 2025-05-14
feat: Apply interrealm specs (1/n)

notJoon:
# Description

- `halt` package p/r integration: After applying the interrealm spec, we encountered an issue where values initialized in `p` could not be accessed in `r`, so we merged them into a single package. Of course, since this approach doesn't seem ideal, it would be good to find a way to separate them again.
- Reflected the modified `grc20` spec: `uint64` -> `int64`


notJoon:
Note: It appears that there is a type stored in access that conflicts with the interrealm specification. We need to resolve this.

dongwon8247 (approved):

dongwon8247 (approved):

PR #630, opened 2025-05-19
external incentive minimum amount paramter change

dongwon8247:


notJoon (approved):

dongwon8247:
@notJoon do you think we should change all uint64 to int64 in this PR? or merge it and fix it in other PR

notJoon:
@dongwon8247 merge it first and fix in take2 PR would be better

PR #644, opened 2025-06-02
fix: Change staker types to follow new grc20 specs

notJoon:
# Description

- remove unused helpers from `_helper_test` file
- change types to follow new grc20 specs


dongwon8247:
There are several areas (especially in tier, pool, reward, and emission logic) that still use `uint64`. I understand this better be handled step by step, but do you agree that, eventually, we should change all uint64 to int64 for consistency, even if negative values are not expected (e.g., for emission rates and balances)?

dongwon8247 (approved):

PR #651, opened 2025-06-12
fix(staker): token amount type from uint64 to int64

mconcat:


notJoon (approved):
Still I have no idea what's going on, but I confirmed that everything works perfect. happy to approve 👍

PR #662, opened 2025-06-17
feat: apply crossing in token teller helper

jinoosss:
## Descriptions

WIP


notJoon (approved):
We were able to confirm that at least the basic functions work on the local chain. It would be good to address future improvements by fixing unit tests as we go. glad to approve this 👍

PR #666, opened 2025-06-19
fix: Fix remaining pool tests

notJoon:
WIP

### DIFF
--- a/contract/r/gnoswap/pool/pool_test.gno
+++ b/contract/r/gnoswap/pool/pool_test.gno
@@ -5,17 +5,34 @@ import (
 	"testing"
 
 	"gno.land/p/demo/testutils"
-	// "gno.land/p/demo/uassert"
+	"gno.land/p/demo/uassert"
 	"gno.land/p/gnoswap/consts"
+
 	// u256 "gno.land/p/gnoswap/uint256"
-	// "gno.land/r/demo/wugnot"
-	// "gno.land/r/gnoswap/v1/gns"
+	"gno.land/r/demo/wugnot"
+	"gno.land/r/gnoswap/v1/gns"
+
 	// pn "gno.land/r/gnoswap/v1/position"
+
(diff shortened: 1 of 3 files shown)


contract/r/gnoswap/pool/pool_test.gno
```diff
@@ -25,98 +42,94 @@ func TestMint(t *testing.T) {
...
 	t.Run("invalid tick range should fail", func(t *testing.T) {
-		defer func() {
-			if r := recover(); r == nil {
-				t.Error("invalid tick range should fail")
-			}
-		}()
-		Mint(cross, token0Path, token1Path, fee, recipient, 100, 100, liquidityAmount, authorized)
-	})
-
-	t.Run("invalid fee amount should fail", func(t *testing.T) {
```
notJoon, line 56: I omitted this case becuase we cannot create pool with invalid tiers.

notJoon:
> @notJoon Can you add a test case of the known swap slippage issue? It can be easily reproduced, and it's best we tackle the known issues while making tests work.

@dongwon8247 While it would be efficient to handle the known issues together, from a change management perspective, I think it would be more appropriate to work on them in separate PRs.

dongwon8247:
It is a side effect caused by the interrealm spec, which we must address to make the contract functioning as previously so we can start from there. Do it in this PR.

notJoon:
> It is a side effect caused by the interrealm spec, which we must address to make the contract functioning as previously so we can start from there. Do it in this PR.

If it's just adding test cases, that wouldn't be a problem, but if it includes modifications as well, I believe it should be done in a separate PR as I mentioned earlier above. This is because the purpose of this work is to make existing tests run, and if we work on it in the same PR, it would be difficult to identify the fixes for that bug all at once when changes occur later.

For these reasons, I cannot agree with the suggestion to work on it here.

PR #668, opened 2025-06-19
fix: int64 conversion

dongwon8247:
The prior implementation simply cast any `uint64` into `int64`, which silently underflowed to negative numbers when the value exceeded `2^63−1`. This update panics instead of producing corrupted (negative) results.

### DIFF
--- a/contract/r/gnoswap/pool/utils_test.gno
+++ b/contract/r/gnoswap/pool/utils_test.gno
@@ -392,21 +392,6 @@ func TestToUint128(t *testing.T) {
 	}
 }
 
-func TestToUint128NilPanic(t *testing.T) {
-	defer func() {
-		if r := recover(); r != nil {
-			expected := "[GNOSWAP-POOL-006] invalid input data || value is nil"
-			if r != expected {
-				t.Errorf("expected panic message %v, got %v", expected, r)
-			}
-		} else {
-			t.Errorf("expected panic, but none occurred")
-		}
-	}()
-
-	toUint128(nil)
-}
-
 func TestToUint128InputPreservation(t *testing.T) {
 	tests := []struct {
 		name                      string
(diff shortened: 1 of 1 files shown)


contract/r/gnoswap/pool/utils.gno
```diff
@@ -41,14 +41,27 @@ func safeConvertToUint64(value *u256.Uint) uint64 {
...
+	res, overflow := value.Uint64WithOverflow()
+	if overflow || res > uint64(consts.INT64_MAX) {
+		panic(ufmt.Sprintf(
+			"%v: amount(%s) overflows int64 range (max %d)",
+			errOutOfRange,
+			value.ToString(),
+			consts.MAX_INT64,
+		))
+	}
+	return int64(res)
```
notJoon, line 64: This function should be tested. 

Previous version was unnecessary because I reused the `safeConvertToUint64` function which already tested. But since this is a new implementation, it would be good to add tests.

dongwon8247: Unlike uint64, which simply panics when a value exceeds its maximum, casting a uint64 directly into an int64 can silently produce a negative number whenever the input is greater than 2^63−1. To avoid corrupting downstream logic with negative values, we must explicitly check against MaxInt64 and panic if the value exceeds that signed‐range limit. 

Yes, we need to test this function. Can you add one please?

contract/r/gnoswap/pool/utils.gno
```diff
@@ -97,15 +130,17 @@ func safeConvertToInt128(value *u256.Uint) *i256.Int {
...
+//   - If masking is required, a new instance is returned without modifying the input.
 //   - MAX_UINT128 is a constant representing `2^128 - 1`.
 func toUint128(value *u256.Uint) *u256.Uint {
-	assertOnlyInitializedUint256(value)
+	assertOnlyNotNil(value)
 	if value.Gt(u256.MustFromDecimal(consts.MAX_UINT128)) {
 		mask := new(u256.Uint).Lsh(u256.One(), consts.Q128_RESOLUTION)
 		mask = mask.Sub(mask, u256.One())
-		value = value.And(value, mask)
+		return new(u256.Uint).And(value, mask)
```
dongwon8247, line 143: Need to check side effects of this change

contract/r/gnoswap/pool/utils.gno
dongwon8247: Btw, this is not a new function. It was already existed and being used. Only an overflow check has been added.

PR #684, opened 2025-06-25
test(pool): Scenario tests

notJoon:
# Description

- Remove `pool_` prefix from pool files 
- Add pool lifecycle scenario tests
- Remove gnoA tests


jinoosss (commented):
There are several points to check.

1. Please modify the output so that the scenario is explicitly visible.

    For example, the scenario for burnd is as follows:
    1. Create Pool
    3. Mint Position
    4. Burn

    The expected values are the burned liquidity and token change amount or error.

2. It appears that there are some missing scenarios.

PR #689, opened 2025-06-30
feat: enhance p/uint256

dongwon8247:
## Changes
- Fixed `SetBytes29` bounds check (was checking index 23, now correctly checks 28)
- Added `parseUintT` helper function for test parsing with proper error handling
- Moved `MulMod` after `Mod` function for better logical ordering
- Added Must* variants for critical arithmetic operations that panic on errors:
	- `MustDiv`, `MustMod`, `MustMulMod`, `MustDivMod` - panic on division by zero
	- `MustMul` - panics on overflow
- Minor code style improvements (using `Zero()` instead of `new(Uint)`)

## Testing
- Significantly expanded test coverage with comprehensive test cases
- Added edge case testing for all arithmetic operations
- Added overflow/underflow detection tests
- Added boundary value tests


contract/p/gnoswap/uint256/fullmath.gno
```diff
@@ -38,75 +36,107 @@ import (
...
+	// Calculate hi * twos
+	tempCalc.Mul(&hiProduct, &twos)
+
+	// Calculate lo | (hi * twos)
+	var combined Uint
+	combined.Or(&loProduct, &tempCalc)
+
+	// Final multiplication for the result
+	result.Mul(&combined, &inverse)
+	return &result
```
notJoon, line 139: These modifications are excessive and unnecessary because the previous version safely manipulates values using new. The currently modified code uses a for loop, but in this case, it performs more unnecessary operations at the machine level. The existing approach was intended to process values in a single-run to avoid such operations. Therefore, I think it would be better to maintain the existing version for the internal logic.

contract/p/gnoswap/uint256/fullmath_test.gno
```diff
@@ -2,13 +2,13 @@ package uint256
...
-var Q128 *Uint
+	"gno.land/p/demo/ufmt"
+)
 
-func init() {
-	Q128 = MustFromDecimal("340282366920938463463374607431768211456") // 2**128
-}
+// =============================================================================
+// Core Functionality Tests
+// =============================================================================
```
notJoon, line 11:
```suggestion
```

contract/p/gnoswap/uint256/fullmath_test.gno
```diff
@@ -483,13 +259,595 @@ func TestDivRoundingUp(t *testing.T) {
...
+}
+
+// =============================================================================
+// Benchmarks
+// =============================================================================
 
-// Benchmarks
 func BenchmarkMulDiv(b *testing.B) {
 	x := MustFromDecimal("123456789012345678901234567890")
 	y := MustFromDecimal("987654321098765432109876543210")
```
notJoon, line 853: gno does not support benchmark for now

notJoon (changes requested):
this changes are dangerous

notJoon (dismissed):
LGTM

notJoon (approved):

PR #690, opened 2025-07-01
feat: enhance int256

dongwon8247:
### Changes
- Removed unnecessary panic checks in `SetInt64` and `SetUint64` methods
- Added `Lte` and `Gte` comparison methods for API completeness
- Optimized `Rsh` (right shift) operation to reduce memory allocations

#### conversion.gno
- Removed dead code panic checks after `initiateAbs()` calls
- `initiateAbs()` already ensures z.abs is never nil

#### cmp.gno
- Added `Lte` (less than or equal) and `Gte` (greater than or equal) methods
- Maintains consistency with uint256 package API

#### bitwise.gno
- Optimized `Rsh` for negative numbers: reduced from 3 Int object allocations to 1 uint256 allocation
- Maintains exact same arithmetic behavior while improving gas efficiency

### Testing
- Added comprehensive test cases for new `Lte`/`Gte` methods
- Added edge case tests for `Rsh` optimization
- All existing tests pass without modification


notJoon:
The int256 package should be changed to an gno’s p/demo/int256 version.​​​​​​​​​​​​​​​​ Still, additional test and API seems good

notJoon (changes requested):
The int256 package should be changed to an gno’s p/demo/int256 version.​​​​​​​​​​​​​​​​ Still, additional test and API seems good

dongwon8247:
@notJoon Ok, I agree. Can you work on removing the int256 dependency in GnoSwap? I will close this PR (this PR only contains the int256 changes). You can review the uint256 changes in #689

PR #695, opened 2025-07-03
fix: remove cross parameter in dry swap route

jinoosss:
## Descriptions

This PR removes the unnecessary `cross` parameter from the `DrySwapRoute` function in the router package. The change simplifies the function signature while maintaining its core functionality.

### Changes

- Removed `cross` parameter from `DrySwapRoute` function signature
- Updated all test files and scenarios to remove the `cross` argument from `DrySwapRoute` calls
- Updated function documentation to reflect the parameter change

### Rationale

The `cross` parameter was redundant in `DrySwapRoute` because:
1. DrySwapRoute is a simulation function that doesn't modify state
2. The parameter wasn't being used in the actual dry swap logic
3. The function already has access to necessary realm information through other means

### Impact

- No functional changes to the swap simulation logic
- Cleaner API surface for dry swap operations
- Reduced parameter complexity in router tests and scenarios

### Testing
All existing tests have been updated and continue to pass:
- Unit tests in router_dry_test.gno
- Integration tests in scenario files
- Multi-hop swap simulations
- Complex routing scenarios

### Related Components

- Pool package's DrySwap function remains unchanged
- ExactInSwapRoute and ExactOutSwapRoute functions are not affected
- All swap simulation behavior remains consistent


notJoon (approved):
Need to check anonymous function usages later. but lgtm

PR #696, opened 2025-07-03
feat: verify scenario with issues

jinoosss:
WIP


notJoon (approved):
I verified tests locally. it should be fine to deploy after merging, regardless of the CI results

dongwon8247 (approved):

PR #701, opened 2025-07-07
feat(gov/governanace): Enhancing Data Structure

notJoon:
# Description

This PR improves the governance package structure by introducing clear separation of concers between time-based scheduling and user-initiated actions. The refactoring reduces duplication, and provides better tracking of proposal lifecycle events.

### The Time-Based Scheduling with `ProposalSchedule`

The introduction of the `ProposalSchedule` type funcdamentally changes how proposal timelines are managed. Instead of calculating time-based states on every access, all critical timestamps are now pre-calculated at proposal creation time using the `createProposalSchedule()` methos. This type contains five key timestamps: `CreateTime`, `ActiveTime`, `VotingEndTime`, `ExecutableTime` and `ExpiredTime`.

### Action State Separation with `ActionState`

The `ActionState` type represents a significant architectural improvement by separating user-initiating actions from time-based state transitions. This structure tracks who critical actions: _cancellation_ (with `Canceled` , `CanceledAt` and `CanceledBy` fields) and _execution_ (whith `Executed`, `ExecutedAt` and `ExecutedBy` fields).

The implementation prioritizes `ActionState` checks over time-based states in the `getCurrentStatus()` function, ensuring that once a user takes an action, it becomes the definitive state regardless of time progression. When a proposal is executed through `ExecuteProposalTransition`, the system now records not just that it was executed, but also who executed it and when.

Similarly, `CancelProposalTransition` captures the complete context of the cancellation. This separation allows the system to distinguish between a proposal that expired naturally versus one that was actively canceled by a user.

### Improved Data Structure Organization

The `ProposalInfo` struct has been reorganized into logical groups that better represent the different aspects of a proposal. The `Status` field now combines both `Schedule` and `ActionState`, providing a complete view of the proposal's current state. All voting-related information is grouped under a `VotingInfo` struct, including vote counts (`Yea`, `Nay`), the maximum voting weight, and the map of eligible voters with their weights. Descriptive information lives in `ProposalMetadata` (`Title`, `Description`), while type-specific data is contained in `ProposalData`. 

```go
// ProposalInfo represents all the information about a proposal
type ProposalInfo struct {
	Proposer      std.Address // The address of the proposer
	ConfigVersion uint64      // The version of the config
	Status        ProposalStatus
	Voting        VotingInfo
	Metadata      ProposalMetadata
	Data          ProposalData
}
```

This organization makes the code more intuitive and reduces the cognitive load when working with proposals.

### Enhanced State Management

The state management logic has been refined to work seamlessly with the new strucutre. The `getCurrentStatus()` method now implements a clear priority system:

1. First checks if the proposal has a schedule
2. Checks `ActionState` for any user-initiated terminal states (expired ot canceled)
3. Determines the current phase based on time and voting results

The proposal state upater continue to handle time-based transition but now uses the `Schedule` methods for cleaner phase detection. State transitions in functions like `ExecuteProposalTransition` and `CancelProposalTransition` propoerly update the `ActionState` while preserving the `Schedule` for historical reference. This dual approach ensures that the system can track both what was suppose to happen (`Schedule`) and what actually happeded (`ActionState`), providing a comple trail for governance actions.

### Remove Code Duplication and Pattern Improvements

The vote validation logic, previously duplicated across multiple functions, has been consolidated into a single ` validateVoteEligibility()` function. This function performs all necessary checks in a consistent order:

1. Verifying proposer existence
2. Checking if the voter has already voted
3. Validating the voting period using the `Schedule`
4. Confirming the proposal hasn't been canceled via `ActionState`
5. Ensuing the coter has sufficient voting weight

Both `CastVoteTransition` and `VoteTransition` now use this shared validation logic, ensuring consistency.

The state update middleware pattern, implemented through `withStateUpdate`, wraps all API getter functions to ensure proposal states are current before returning data. This pattern eliminates the repetitive code that was previously at the beginning of each getter function. For example, `GetProposal` now simply returns `withStateUpdate(func() string { ... })`, and the middleware handles the state update automatically. This approach has been applied to all public API functions indluding `GetProposalById()`, `GetVoteStatusFromProposalById()`, and others.

## Testing

All existing tests have been updated to work with the new structure while maintaining the same behaviour.


contract/r/gnoswap/gov/governance/type.gno
```diff
@@ -6,54 +6,90 @@ import (
...
+	Schedule    ProposalSchedule
+	ActionState ActionState
+}
 
-	// Over how many seconds the voting weight is averaged for proposal voting as creation/cancellation threshold
-	VotingWeightSmoothingDuration uint64
+// VotingInfo contains all voting-related information for a proposal
+type VotingInfo struct {
+	Yea             *u256.Uint // collected yes votes
+	Nay             *u256.Uint // collected no votes
```
notJoon, line 29: I think `uint64` seems sufficient for these two fields.

tests/scenario/gov/governance/governance_proposal_community_pool_spend_filetest.gno
```diff
@@ -116,7 +116,7 @@ func execute(proposalID uint64) {
 	proposalState := governance.GetExecutionStateByProposalId(proposalID)
 
 	ufmt.Printf("[EXPECTED] receiver's balance increased by %d\n", afterReceiverBalance-prevReceiverBalance)
-	ufmt.Printf("[EXPECTED] proposal's executed state is %v\n", proposalState.Executed)
```
notJoon, line 119: it has been modified to return a stringfied state instead of boolean

contract/r/gnoswap/gov/governance/proposal_state.gno
```diff
@@ -165,140 +176,98 @@ func (u *proposalStateUpdater) shouldUpdate() bool {
...
-}
-
 // Time period calculations
 func (u *proposalStateUpdater) getVotingPeriod() (start, end uint64) {
-	start = u.proposal.State.CreatedAt + u.config.VotingStartDelay
+	if u.proposal.Status.Schedule.CreateTime != 0 {
+		return u.proposal.Status.Schedule.ActiveTime, u.proposal.Status.Schedule.VotingEndTime
+	}
+	// Fallback: If `Status.Schedule` is not initialized,
+	// calculate from current time
```
notJoon, line 185: fallbacks are just for case. this branch should not happened

contract/r/gnoswap/gov/governance/config.gno
```diff
@@ -11,6 +11,26 @@ import (
...
+	VotingWeightSmoothingDuration uint64
+	// Quorum is the percentage of total GNS supply required for proposal approval
+	Quorum uint64
+	// ProposalCreationThreshold is the minimum average voting weight required to create a proposal
+	ProposalCreationThreshold uint64
+	// ExecutionDelay is the waiting period after voting ends before a proposal can be executed (in seconds)
+	ExecutionDelay uint64
+	// ExecutionWindow is the time window during which an approved proposal can be executed (in seconds)
+	ExecutionWindow uint64
+}
```
notJoon, line 32: Like `ProposalInfo` type, it seems better to divide it into smaller units.

contract/r/gnoswap/gov/governance/state_transition.gno
```diff
@@ -193,7 +175,7 @@ func UpdateProposalStateTransition() Transition {
```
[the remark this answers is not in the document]
notJoon: I believe we should keep that functon. It plays a crucial role in state management and maintaining data consistency. Because it automatically manages proposal lifecycle (upcoming -> active -> passed/rejected -> expired), and also ensures up-to-date proposal state before voting or execution. Lastly, it automatically archieves terminated proposals.

However, I agree it's quite expensive to use within `Vote` function. It was a consistent abstraction applied because all of them share the same logic, but it would be better to improve this by updating the state of only specific proposals.

PR #706, opened 2025-07-11
fix(r/emission): Add validation for `bptPct` parameter in `calculateAmount`

notJoon:
# Description

Added validation to ensure `bptPct` cannot exceed 10000


jinoosss (approved):
👍

PR #707, opened 2025-07-11
fix(r/emission): Correct `leftAmount` Check in `distributeToTarget`

notJoon:
# Description

The `distributeToTarget` function should checks local `leftAmount` variable instead of global `leftAmount`

## Changes

**Before**

```go
leftAmount := amount - totalSent
if leftGNSAmount > 0 {  // <- global variable
    setLeftGNSAmount(leftAmount)
}
```

**After**

```go
leftAmount := amount - totalSent
if leftAmount > 0 {  // <- local leftAmount
    setLeftGNSAmount(leftAmount)
}
```

This wrong variable usage caused the following issues:

- **Token Loss**: When `leftGNSAmount` was `0` (initial state), any undistributed tokens from rounding errors were permanently lost
- **Incorrect Tracking**: The leftover amount would only be saved if there was already a previous leftover amount (`leftGNSAmount > 0`)
- **Cumulative Loss**: Over many distributions, significant amounts of tokens could be lost due to rounding errors not being tracked


contract/r/gnoswap/emission/leftamount_test.gno
jinoosss, line 1: Please change it to a table-driven format or separate the tests.

notJoon: changed https://github.com/gnoswap-labs/gnoswap/pull/707/commits/19526b127fc4ca64c3784dc2c337ab557f36b166 👍

PR #725, opened 2025-07-18
fix(router): Ensure users receive exact amount in `ExactOut` swaps after protocol fee deduction

notJoon:
# Description

This PR fixes an issue where users were receiving less than the requested amount in `ExactOut` swaps due to protocol fee deduction. The fix ensures that when users specify an amount in an `ExactOut` swap, they receive at least that amount after all fees are deducted.

## Problem

Previously, in `ExactOut` swaps:
- Users would specify an amount they wanted to receive (e.g., 100 tokens)
- The pool would return approximately that amount
- After protocol fee deduction (0.15%), users would receive less than requested (e.g., 99.85 tokens)
- This violated the core principle of `ExactOut` swaps where users should receive exactly what they request

## Solution

1. **Added `calculateAmountWithFee` function in protocol_fee_swap.gno:**
   - Calculates the amount needed from the pool to ensure users receive their requested amount after fees
   - Formula: `poolAmount = userAmount * 10000 / (10000 - swapFee)`
   - Includes proper rounding to always favor the user

2. **Modified ExactOut swap logic in exact_out.gno:**
   - Now uses `calculateAmountWithFee` to request the correct amount from the pool
   - Passes the original user-requested amount for validation instead of the pool-adjusted amount

3. **Fixed validation order in router.gno:**
   - Protocol fee is now deducted before validating the output amount
   - Ensures validation checks the actual amount users will receive

## Test Changes

**Why the test case was modified:**

The test case "ExactOut: Real world slippage error" had incorrect values that didn't represent a valid `ExactOut` scenario:

**Before:**
- amountSpecified: 600901352 (what validation was checking against)
- resultAmountOut: 600901351 (what pool returned)
- User would receive: ~599999999 after fees (less than requested)

**After:**
- amountSpecified: 600000000 (what user actually wants to receive)
- resultAmountOut: 600901353 (what pool needs to return to cover fees)
- User receives: ~600000001 after fees (at least what was requested)

The test was incorrectly set up with the pool returning less than what the user requested, which would always fail in a proper `ExactOut` implementation. The corrected test now properly simulates a scenario where the pool returns enough tokens to ensure the user receives their requested amount after fee deduction.


jinoosss:
We need to be able to validate that the calculation is the correct value. cc @notJoon

PR #737, opened 2025-07-20
fix: change the pointer object query method to internal method

jinoosss:
## Descriptions

This PR improves encapsulation by converting the `GetCurrentConfig()` function from a public method to an internal method `getCurrentConfig()`, restricting direct external access to the configuration pointer and following Go's encapsulation principles.

#### 1. **Function Access Level Change**
- **Before**: `GetCurrentConfig()` - Public function (exported)
- **After**: `getCurrentConfig()` - Internal function (unexported)

```go
// Before: Public access
func GetCurrentConfig() *config {
    if currentConfig == nil {
        return nil
    }
    return currentConfig
}

// After: Internal access only
func getCurrentConfig() *config {
    if currentConfig == nil {
        return nil
    }

    return currentConfig
}
```

#### 2. **Updated Test References**
- Updated all test file references from `GetCurrentConfig()` to `getCurrentConfig()`
- Maintains test functionality while respecting the new access level

#### 1. **Improved Encapsulation**
- Prevents external packages from directly accessing internal configuration state
- Follows Go's convention of using capitalization to control access levels

#### 2. **Better Security**
- Restricts direct manipulation of configuration pointers from outside the package
- Reduces the attack surface by limiting external access points

#### 3. **Cleaner API Design**
- Internal configuration management should not be exposed as public API
- External packages should use dedicated getter functions like `GetAddress()` instead

### Impact
- **No breaking changes**: The function was only used internally within the access package and its tests
- **Improved maintainability**: Internal configuration management is now properly encapsulated

### DIFF
--- a/contract/r/gnoswap/access/access.gno
+++ b/contract/r/gnoswap/access/access.gno
@@ -109,11 +109,12 @@ func declareNewRole(roleName string, addr std.Address) error {
 	return nil
 }
 
-// GetCurrentConfig returns the current configuration
-func GetCurrentConfig() *config {
+// getCurrentConfig returns the current configuration
+func getCurrentConfig() *config {
 	if currentConfig == nil {
 		return nil
 	}
+
 	return currentConfig
 }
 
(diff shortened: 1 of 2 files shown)


notJoon (approved):
It would be more better to remove get prefix from the function name to make follow the Go's convention. 👍

PR #741, opened 2025-07-21
chore(staker): Remove Unnecessary Legacy Tests

notJoon:
# Description

These files should be removed


dongwon8247:
This should all be remained in the scenario tests after removing duplicated ones. The staker contract is dependent on this.

PR #742, opened 2025-07-21
fix: Refactor Access

notJoon:
WIP


tests/scenario/rbac/rbac_role_management_filetest.gno
notJoon, line 1: this test was tested permission related logics. But currently we were removed those functionalities, so it does not valid test anymore

jinoosss (approved):

PR #748, opened 2025-07-22
fix(access): convert package path to address

jinoosss:
## Descriptions

- Change convert package path method.
    - NewCoderealm to DerivePkgAddr


notJoon (approved):
I confirmed that the problem was fixed during local deployment

PR #759, opened 2025-07-28
fix(staker):  Improve test code quality and reduce duplication

notJoon:
# Description

The staker package had significant test code duplication and maintenance challenges:
 - Multiple test files with identical patterns and helper functions
 - Repetitive test cases that could be consolidated
 - Difficult to add new test cases or maintain existing ones

Additionally, when retrieving the contract address, we were calling the getter function every time. This has been changed to call the function only once to reduce the number of function calls.

Changed the type of `unstakingFee` from `uint64` to `int64` to reduce unnecessary type conversions and comply with the grc20 specification. Furthermore, this makes it easier to detect overflow.

## Changes

1. Test Consolidation with Data-Driven Approach 
 - Converted repetitive test cases into data-driven tests
 - Reduced code duplication by 25-30% modified test files
 
2. Generic Test Framework for Permission-Based Functions
 - Created `testPermissionBasedFunction` to handle common permission testing patterns
 - Eliminated duplicate permission checking logic
 - Standardized error message handling with constants

3. Specific File Improvements
`reward_calculation_warmup_test.gno`:
  - Consolidated 12 similar tick crossing tests into parameterized tests
  - Added helper functions to reduce assertion boilerplate
  - Improved test names for better clarity
  - Reduced from 400 to ~386 lines

`external_incentive_test.gno`:
  - Replaced 6 nearly identical test cases with data-driven approach
  - Improved test coverage by adding token and refundee verification
  - Reduced from 109 to 82 lines (25% reduction)

`external_deposit_fee_test.gno`:
  - Created generic permission testing framework
  - Consolidated 3 permission test functions into reusable
  pattern
  - Reduced from 228 to ~168 lines (30% reduction)

### DIFF
--- a/contract/r/gnoswap/referral/utils.gno
+++ b/contract/r/gnoswap/referral/utils.gno
@@ -11,31 +11,28 @@ import (
 	_ "gno.land/r/gnoswap/v1/rbac"
 )
 
-// validCallers is a lookup table of addresses that are authorized to modify referral data.
+// validCallerRoles is a list of roles that are authorized to modify referral data.
 // This includes governance contracts, router, position manager, and staker contracts.
-var validCallers = map[std.Address]bool{
-	getAddress(prabc.ROLE_GOVERNANCE.String()): true,
-	getAddress(prabc.ROLE_GOV_STAKER.String()): true,
-	getAddress(prabc.ROLE_ROUTER.String()):     true,
-	getAddress(prabc.ROLE_POSITION.String()):   true,
-	getAddress(prabc.ROLE_STAKER.String()):     true,
-	getAddress(prabc.ROLE_LAUNCHPAD.String()):  true,
+var validCallerRoles = []string{
+	prabc.ROLE_GOVERNANCE.String(),
+	prabc.ROLE_GOV_STAKER.String(),
+	prabc.ROLE_ROUTER.String(),
+	prabc.ROLE_POSITION.String(),
+	prabc.ROLE_STAKER.String(),
+	prabc.ROLE_LAUNCHPAD.String(),
 }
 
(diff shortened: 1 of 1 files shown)


contract/r/gnoswap/staker/consts.gno
```diff
@@ -2,11 +2,11 @@ package staker
...
+	GNOT string = "gnot"
 
 	// ref: https://github.com/gnolang/gno/blob/81a88a2976ba9f2f9127ebbe7fb7d1e1f7fa4bd4/examples/gno.land/r/demo/wugnot/wugnot.gno#L19
 	UGNOT_MIN_DEPOSIT_TO_WRAP uint64 = 1000
 
-	GNS_PATH string      = "gno.land/r/gnoswap/v1/gns"
-	WUGNOT_PATH string      = "gno.land/r/demo/wugnot"
+	GNS_PATH    string = "gno.land/r/gnoswap/v1/gns"
+	WUGNOT_PATH string = "gno.land/r/demo/wugnot"
 )
```
dongwon8247, line 12: It's okay to hardcode the paths that will not be changed/upgraded, but let's remove `v1` in the contract that will not be upgraded (gns, emission, halt, etc ...) and remain `v1` only in the contract that will be upgraded (pool, position, router, ...)

notJoon: I agree with the suggestion to remove `v1` from the package paths of all contracts except pool, position, router, staker, gov, and launchpad, and I think it's a necessary task.

However, when package paths are changed, contract addresses will also change simultaneously (since gno addresses are generated based on package paths), which could cause excessive conflicts with all currently open PRs. To avoid unnecessary work, how about we apply all the currently open PRs first and then make the changes?

PR #760, opened 2025-07-28
fix(router): Refactor `ExactInSwapRoute` and `ExactOutSwapRoute` functions

notJoon:
# Description

Extracted common logic from `ExactInSwapRoute` and `ExactOutSwapRoute` functions to eliminate code duplication in the router codebase.


contract/r/gnoswap/router/base.gno
```diff
@@ -209,3 +214,107 @@ func handleMultiSwap(
 		panic(errInvalidSwapType)
 	}
 }
+
+/* --------- common swap operation --------- */
```
dongwon8247, line 218:
```suggestion
```

contract/r/gnoswap/router/base.gno
```diff
@@ -209,3 +214,107 @@ func handleMultiSwap(
...
+func commonSwapRoute(cur realm, params SwapRouteParams) (string, string, string, error) {
+	// Common pre-swap checks
+	halt.AssertIsNotHaltedRouter()
+	halt.AssertIsNotHaltedWithdraw()
+
+	if err := checkDeadline(params.deadline); err != nil {
+		panic(addDetailToError(errExpired, err.Error()))
+	}
+
+	emission.MintAndDistributeGns(cross)
```
dongwon8247, line 244: can you follow this rule? https://www.notion.so/Gnoswap-23a17eb23ada8015a29cfb75067f3b8a?source=copy_link

notJoon: Since the `checkDeadline` function is an internal private function, that line explicitly handling panic at the upper level when an error occurs, as shown in the guide. I believe we are following the guide's recommendations.

notJoon: Moved the position of the `assertXXX` function to the beginning of the entry point function and modified it so that only errors are returned within the private section. https://github.com/gnoswap-labs/gnoswap/pull/760/commits/5c640652bc3f83230a7881081586687f6ed5a336

contract/r/gnoswap/router/exact_out.gno
```diff
@@ -42,56 +39,33 @@ func ExactOutSwapRoute(cur realm,
 	halt.AssertIsNotHaltedRouter()
 	halt.AssertIsNotHaltedWithdraw()
 
-	if err := checkDeadline(deadline); err != nil {
-		panic(addDetailToError(errExpired, err.Error()))
```
jinoosss, line 46: It seems like a good idea to define it as an assert function.
https://github.com/gnoswap-labs/gnoswap/blob/be99ee38b736304a5c506aa3483d968037ad2913/contract/r/gnoswap/position/position.gno#L193

contract/r/gnoswap/router/base.gno
```diff
@@ -204,8 +208,106 @@ func handleMultiSwap(
...
+
+	switch typedOp := op.(type) {
+	case *ExactInSwapOperation:
+		userBeforeWugnotBalance = typedOp.userBeforeWugnotBalance
+		userWrappedWugnot = typedOp.userWrappedWugnot
+		limitValue = u256.MustFromDecimal(params.limitAmount)
+	case *ExactOutSwapOperation:
+		userBeforeWugnotBalance = typedOp.userBeforeWugnotBalance
+		userWrappedWugnot = typedOp.userWrappedWugnot
+		limitValue = u256.MustFromDecimal(params.limitAmount)
```
jinoosss, line 289: In my opinion, branching by type in a common function can be a black box.

contract/r/gnoswap/router/exact_out.gno
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/760/commits/ee84dd12dfec4105dc795a45135505b3b63548dd

contract/r/gnoswap/router/base.gno
notJoon: Modified the `MintAndDistrubitionGns` function to be checked directly at the upper level. I think it would be better to check the deadline validation part right before creating a new object, so I created a new assert function and modified it to perform the validation (it was also @jinoosss's suggestion).
https://github.com/gnoswap-labs/gnoswap/pull/760/commits/ee84dd12dfec4105dc795a45135505b3b63548dd

dongwon8247 (approved):
ltgm

contract/r/gnoswap/router/base.gno
dongwon8247: Sorry I didn't see this comment. @notJoon Can you check?

dongwon8247: How about this?

```
switch typedOp := op.(type) {
case *ExactInSwapOperation:
	userBeforeWugnotBalance = typedOp.userBeforeWugnotBalance
	userWrappedWugnot = typedOp.userWrappedWugnot
	limitValue, err = u256.FromDecimal(params.limitAmount)
	if err != nil {
		return "", "", "", errors.New("invalid limitAmount")
	}
case *ExactOutSwapOperation:
	userBeforeWugnotBalance = typedOp.userBeforeWugnotBalance
	userWrappedWugnot = typedOp.userWrappedWugnot
	limitValue, err = u256.FromDecimal(params.limitAmount)
	if err != nil {
		return "", "", "", errors.New("invalid limitAmount")
	}
default:
	return "", "", "", errors.New("unexpected RouterOperation type")
}
```

If panic is need here, we can keep the MustFromDecimal ones.

contract/r/gnoswap/router/base.gno
```diff
@@ -204,8 +207,105 @@ func handleMultiSwap(
...
+	deadline    int64
+	referrer    string
+	typ         SwapType
+	exactAmount string // amountIn for ExactIn, amountOut for ExactOut
+	limitAmount string // amountOutMin for ExactIn, amountInMax for ExactOut
+}
+
+// commonSwapRoute handles the common logic for both ExactIn and ExactOut swaps
+// Returns inputAmount, outputAmount, and actualReferrer
+func commonSwapRoute(cur realm, params SwapRouteParams) (string, string, string, error) {
```
jinoosss, line 229: In my opinion, it seems like a good idea to separate the swap operation into a function that creates the swap operation and a function that executes the swap operation.
Even if only the functions are separated, they can be managed more explicitly.

notJoon:
The suggested changes were modified and approved simultaneously yesterday. I will merge this after confirmed CI passes.

PR #779, opened 2025-08-01
test(position): Add Low Liquidity Behaviour Checks for `Reposition` Function

notJoon:
# Description

This PR adds a test scenario to verify the `Reposition` function's behavior in low liquidity environments. The test addresses concerns about potential failures when insufficient liquidity might prevent swaps during position adjustments.

## Problem Statement

There was a concern that the `Reposition` function might fail when:
- Pool liquidity is extremely low
- Swaps cannot be executed due to insufficient liquidity
- Price has moved significantly from the original position range

## Solution

After thorough analysis of the `Reposition` implementation, it shouldn't be poosible because that:

1. The function does not perform any internal swaps
2. It directly adds liquidity using provided tokens
3. Token ratios are automatically calculated based on current price

A new test file has been added to demonstrate this behavior.

## Test Scenario Details

The test creates an extreme environment with:
- Minimal initial liquidity (63,740,878 - about 1/5 of normal tests)
- Large swap that moves price significantly (tick 10000 → 8991)
- Successful reposition to new range (6000-10000)
- Verification of normal operation after reposition

### DIFF
--- a/tests/scenario/position/position_reposition_low_liquidity_filetest.gno
+++ b/tests/scenario/position/position_reposition_low_liquidity_filetest.gno
@@ -1,0 +1,316 @@
+// position reposition with low liquidity scenario
+package main
+
+import (
+	"std"
+	"testing"
+
+	"gno.land/p/demo/ufmt"
+
+	prbac "gno.land/p/gnoswap/rbac"
+
+	"gno.land/r/gnoswap/v1/access"
+	"gno.land/r/gnoswap/v1/pool"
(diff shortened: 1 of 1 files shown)


jinoosss:
Please check the reposition function and add a reposition flow test with a multimessage in txtar:

Reposition is done in the following order: 
- decrease liquidity -> swap route -> reposition.

https://github.com/gnoswap-labs/gnoswap/blob/main/tests/integration/testdata/upgrade_contracts.txtar#L116-L140

PR #790, opened 2025-08-05
fix(governance): Remove unnecessary `success` fields

notJoon:
# Description

- Removed success field from `ProposalActionStatus`
- Removed `IsSuccess()` method that was never called
- Removed `updateResult()` and `updateExecuteResult()` methods
- Since proposal execution either succeeds or panics, tracking success state was redundant
- Removed `halt.AssertIsNotHaltedWithdraw()` from `ProposeText` and `ProposeParameterChange` to follows the halt-checking documentation


dongwon8247 (approved):
Looks good, but let's give @jinoosss a chance to review

PR #793, opened 2025-08-05
Feature/timestamp based reward

mconcat:


contract/r/gnoswap/gov/staker/staker_delegate.gno
```diff
@@ -355,7 +355,7 @@ func unDelegate(
 	delegatee std.Address,
 	amount,
 	currentTime,
-	currentHeight int64,
+	currentTimestamp int64,
```
jinoosss, line 358: currentTime, currentTimestamp contain the same information.
The system is time-based, but it seems like we need to record the block height at which the transaction occurred.

notJoon: It doesn't seem like it would particularly affect the functionality, so I didn't pass the time separately when deleting, but it would definitely be good to record both height and time together as you mentioned. I've checked the modified code and it looks fine.

PR #800, opened 2025-08-07
feat: improve tick cross hook for batch process

jinoosss:
## Descriptions

This PR introduces a significant performance optimization for tick cross processing during swaps by implementing a batch processing system. The previous approach called `pool.modifyDeposit()` for every individual tick cross, which led to substantial computational overhead for swaps crossing multiple ticks.

## Problem Statement

In the previous implementation:
- Each tick cross triggered an immediate `tickCrossHook()` call
- Every hook call resulted in an expensive `pool.modifyDeposit()` operation
- For swaps crossing N ticks, this meant N separate expensive calculations
- Performance degraded significantly for large swaps crossing many ticks

### New Hook System Architecture

The solution introduces a three-hook system to enable batch processing:

1. **`swapStartHook(poolPath, timestamp)`** - Initializes batch processing when a swap begins
2. **`tickCrossHook(poolPath, tickId, zeroForOne, timestamp)`** - Accumulates tick cross data during swap execution
3. **`swapEndHook(poolPath)`** - Processes all accumulated tick crosses in a single batch operation

### Batch Processing Implementation

**SwapBatchProcessor Structure:**
```go
type SwapBatchProcessor struct {
    poolPath  string
    pool      *Pool
    crosses   []SwapTickCross
    timestamp int64
    isActive  bool
}
```

**Processing Flow:**
1. **Initialization**: `swapStartHook()` creates a new batch processor
2. **Accumulation**: Each `tickCrossHook()` call adds tick cross data to the batch
3. **Batch Processing**: `swapEndHook()` processes all accumulated changes in one operation

### Computational Complexity Reduction

| Tick Crosses | Previous (modifyDeposit calls) | Current (modifyDeposit calls) | Improvement |
|--------------|-------------------------------|-------------------------------|-------------|
| 3 ticks      | 3 calls                       | 1 call                        | 67% reduction |
| 10 ticks     | 10 calls                      | 1 call                        | 90% reduction |
| 20 ticks     | 20 calls                      | 1 call                        | 95% reduction |
| 50 ticks     | 50 calls                      | 1 call                        | 98% reduction |

### Memory vs Performance Trade-off

- **Memory Cost**: +32N bytes per swap (where N = number of tick crosses)
- **CPU Savings**: 90-98% reduction in expensive calculations
- **Overall Impact**: Significant net performance gain, especially for larger swaps

### Pool Contract Updates (`pool/swap.gno`)

- Moved hook declarations from `type.gno` to `swap.gno` for better organization
- Added `swapStartHook` and `swapEndHook` alongside existing `tickCrossHook`
- Updated hook signature to include timestamp parameter
- Integrated hook calls into swap lifecycle with proper timing

### Staker Contract Updates (`staker/reward_calculation_tick.gno`)

- Implemented `SwapBatchProcessor` for accumulating tick cross data
- Added batch processing logic with cumulative delta calculations
- Maintained backward compatibility with immediate processing fallback
- Optimized memory allocation patterns

## Backward Compatibility

The implementation includes a fallback mechanism (`processTickCrossImmediate`) that ensures:
- Non-swap tick crosses are processed immediately using the original logic
- System behavior remains consistent for edge cases
- No breaking changes to existing functionality


contract/r/gnoswap/v1/staker/reward_calculation_tick.gno
```diff
@@ -142,46 +144,159 @@ func (self *Tick) updateCurrentOutsideAccumulation(timestamp int64, acc *u256.Ui
...
+	}
+
+	batch := currentSwapBatch
+	pool := batch.pool
+	timestamp := batch.timestamp
+
+	// Calculate cumulative liquidity delta
+	cumulativeDelta := i256.Zero()
+	for _, tickCross := range batch.crosses {
+		cumulativeDelta = i256.Zero().Add(cumulativeDelta, tickCross.delta)
```
notJoon, line 270: Overflow may occur when many tick crosses accumulate.

contract/r/gnoswap/v1/staker/reward_calculation_tick.gno
```diff
@@ -142,46 +144,159 @@ func (self *Tick) updateCurrentOutsideAccumulation(timestamp int64, acc *u256.Ui
...
+}
+
+// processBatchedTickCrosses processes all accumulated tick crosses at once
+func processBatchedTickCrosses() {
+	if currentSwapBatch == nil || len(currentSwapBatch.crosses) == 0 {
+		return
+	}
+
+	batch := currentSwapBatch
+	pool := batch.pool
```
notJoon, line 264: nil reference seems possible

contract/r/gnoswap/v1/staker/reward_calculation_tick.gno
```diff
@@ -142,46 +144,159 @@ func (self *Tick) updateCurrentOutsideAccumulation(timestamp int64, acc *u256.Ui
...
+	liquidityDelta := tick.stakedLiquidityDelta
+	if zeroForOne {
+		liquidityDelta = i256.Zero().Neg(liquidityDelta)
+	}
+
+	newAcc := pool.modifyDeposit(liquidityDelta, timestamp, nextTick)
+	tick.updateCurrentOutsideAccumulation(timestamp, newAcc)
+}
+
+// processBatchedTickCrosses processes all accumulated tick crosses at once
```
notJoon, line 257: I think that if a failure occurs in the middle of batch processing, only part of the data might be updated. worth consider to add rollback system.

notJoon: additionally, this is a minor point, but this function appears to be the main logic, it would be good to add detailed context to the comments rather than just simple explanations

notJoon (changes requested):
It seems you've achieved maximum performance improvement with minimal changes. I would appreciate it if you could check a few additional comments

contract/r/gnoswap/v1/staker/reward_calculation_tick.gno
jinoosss: The maximum liquidity value is 2^128.
The amount of liquidity that can change in a single swap cannot exceed int256.

PR #801, opened 2025-08-07
fix: improve governance proposal cancel condition

jinoosss:
## Descriptions

This PR improves the governance proposal cancellation mechanism by adding proper authorization checks to ensure that only the original proposer can cancel their own proposals. The previous implementation lacked this essential security validation, which could potentially allow unauthorized users to cancel proposals.

## Problem Statement

In the previous implementation:
- The `Cancel()` function did not verify that the caller was the original proposer
- Any user could potentially cancel any proposal, regardless of who created it
- This represented a significant security vulnerability in the governance system
- Test cases incorrectly expected unauthorized cancellations to succeed

### Added Proposer Authorization Check

**New assertion function:**
```go
func assertCallerIsProposer(proposalID int64, caller std.Address) {
    proposal, exists := getProposal(proposalID)
    if !exists {
        panic(errProposalNotFound)
    }

    if !proposal.IsProposer(caller) {
        panic(errNotProposer)
    }
}
```

**Enhanced Proposal struct:**
```go
func (p *Proposal) IsProposer(addr std.Address) bool {
    return p.proposer == addr
}
```

### Updated Cancellation Flow

The `Cancel()` function now:
1. **First** validates that the caller is the original proposer
2. **Then** proceeds with the existing cancellation logic
3. Provides clear error messages for unauthorized attempts

### Authorization Validation

- **Before**: No proposer verification - any user could cancel any proposal
- **After**: Strict validation ensures only the original proposer can cancel their proposal

### Error Handling Enhancement

- Added specific error code `GNOSWAP-GOVERNANCE-022` for unauthorized cancellation attempts
- Improved error specificity from generic "data not found" to precise "not proposer"
- Better error messaging helps with debugging and user experience


contract/r/gnoswap/v1/gov/governance/proposal.gno
```diff
@@ -220,6 +220,10 @@ func (p *Proposal) ConfigVersion() int64 {
 	return p.configVersion
 }
 
+func (p *Proposal) IsProposer(addr std.Address) bool {
+	return p.proposer == addr
+}
```
notJoon, line 225: This function is only called from the `assertCallerIsProposer` function. it's a simple conditional statement, I think it would be good to inline it.

notJoon (approved):
optimization chance is available, changes LGTM

PR #826, opened 2025-08-20
feat: txtar runner

notJoon:
WIP


notJoon:
running txtar tests requires additional complex setup, so it would be better to handle it manually using setup scripts(of course, there are also some that don't work)

PR #829, opened 2025-08-26
perf(staker): Optimize `PoolTier` counting with cached counters

notJoon:
## Description

Optimizes the way `PoolTier` tracks the number of pools per tier.

Previously, methods like `CurrentAllTierCounts()` and `CurrentCount()` scanned the entire `membership` tree on every call. This made counting $O(n)$ with respect to the number of pools, even though counts only change when a pool moves tiers.

### Changes

- Added a `counts` field to `PoolTier` as a cached counter
- Updated `changeTier()` to inc/dec the counts whenever a pool is added, removed or moved between tiers
- Updated `CurrentCount()` and `CurrentAllTierCounts()` to return results directly from the cached counter

By doing so, all tier count queries are now $o(1)$, avoiding repeated full-tree iteration.

dependent with #828

### DIFF
--- a/contract/r/gnoswap/v1/staker/reward_calculation_pool_tier.gno
+++ b/contract/r/gnoswap/v1/staker/reward_calculation_pool_tier.gno
@@ -96,7 +96,7 @@ type PoolTier struct {
 	counts [AllTierCount]uint64
 
 	lastRewardCacheTimestamp int64
-	lastRewardCacheHeight int64
+	lastRewardCacheHeight    int64
 
 	currentEmission int64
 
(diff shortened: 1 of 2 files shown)


contract/r/gnoswap/v1/staker/reward_calculation_pool_tier.gno
```diff
@@ -5,6 +5,7 @@ import (
 	"gno.land/p/demo/ufmt"
 )
 
+// TODO: define Tier type as uint8 size
```
dongwon8247, line 8: can you define it in this pr to improve type safety?

jinoosss (approved):

contract/r/gnoswap/v1/staker/reward_calculation_pool_tier.gno
```diff
@@ -192,19 +190,32 @@ func (self *PoolTier) changeTier(currentHeight int64, currentTime int64, pools *
...
 		// removed from the tier
 		self.membership.Remove(poolPath)
+
+		// count decrement
+		if currentTier > 0 {
+			if self.counts[currentTier] == 0 {
+				panic("counts underflow: removing from empty tier")
+			}
+			self.counts[currentTier]--
+		}
```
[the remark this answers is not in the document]
notJoon: https://github.com/gnoswap-labs/gnoswap/pull/829/commits/468d89e3d34affb2beb65a5d42cec11eb8506d83

contract/r/gnoswap/v1/staker/reward_calculation_pool_tier.gno
dongwon8247: It's only for the issues they find to keep it within the scope. Anyways, just don't forget to memo somewhere and follow up on it then. I've seen you say that but not follow up to address it many times. Also, DONOT mention anything about the audit process publicly. Delete these comments.

notJoon: removed https://github.com/gnoswap-labs/gnoswap/pull/829/commits/372d94bcd8e1ba56c8f3c8c202367a41d6ac73b3

PR #831, opened 2025-08-27
fix(pool): Clarify Tick Bitmap Semantics

notJoon:
## Description

This PR removes ambiguity in the `tickBitmap` implementation by explicitly handling edge cases. The updated test validate rules more clearly, ensuring correctness at boundaries such as negative ticks and word transitions.

**Before*

```go
shifted := u256.Zero().Lsh(u256.One(), bitPos+1) // (1 << 256) ?!
mask := shifted.Sub(shifted, u256.One())
```

**After**
if bitPos == 255 {
    return u256.Zero().Not(u256.Zero()) // 2^256 - 1
}
shifted := u256.Zero().Lsh(u256.One(), bitPos+1)
return u256.Zero().Sub(shifted, u256.One())
```

### Why This Was Problem?

Previously, some aspects of the tick bitmap logic were implicit or left to library behaviour, which could cause ambiguity:

1. Edge case for masks
    - At `bitPos == 255`, the expression `(1 << (bitPos+1))` implied `1 << 256`, which in some implementations evaluates `0`. This happened to produce the correct result ($2^256-1$) but only by accident.
    - At`bitPos == 0`, the GT mask was `~((1<<0) -1) = ~0`, which produces **all ones**. This was correct but not documented.
2. Native tick handling
    - Go and Gno's integer division truncartes toward zero, bit bitmap compression requires floor division toward $-\infty$. Without explicit correction, negative ticks could map to the wrong word.

3. Updated tests to focus on boundary conditions (e.g., ticks at -1, 0, 255, 256) and dense spacing rather than line coverage.

    - Positioning: verifies tickBitmapPosition across positive, negative, and boundary ticks.
    - Mask generation: checks correctness at bitPos=0, bitPos=7, bitPos=254, bitPos=255.
    - Flip behavior: verifies set/unset toggling and panic on misaligned tick spacing.
    - Search semantics: validates LTE (MSB search) and GT (LSB search) within a word, and “no initialized ticks” paths.
    - Dense spacing: ensures consecutive initialized ticks are found correctly when tickSpacing=1.
    - Word boundaries: verifies transitions across −257, −256, −255, −1, 0, 1, 254, 255, 256.
    - Floor division check: ensures tick >> 8 is equivalent to floor division by 256.

### DIFF
--- a/contract/r/gnoswap/v1/pool/tick_bitmap.gno
+++ b/contract/r/gnoswap/v1/pool/tick_bitmap.gno
@@ -113,19 +113,15 @@ func (p *Pool) initTickBitmap(wordPos int16) {
 
 // tickBitmapPosition calculates the word and bit position for a given tick
 func tickBitmapPosition(tick int32) (int16, uint8) {
-	wordPos := int16(tick >> 8)  // arithmetic right shift, floor(tick/256)
-	bitPos := uint8(tick & 0xff) // tick % 256, optimized with bit mask
-	return wordPos, bitPos
+	return int16(tick >> 8), uint8(tick) & bitMask8
 }
 
 // getWordAndBitPos gets tick's wordPos and bitPos depending on the swap direction
 func getWordAndBitPos(tick int32, lte bool) (int16, uint8) {
-	// Branch-free version: tick += (!lte ? 1 : 0)
-	increment := int32(0)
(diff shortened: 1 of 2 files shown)


dongwon8247 (approved):
LTGM 👍

PR #832, opened 2025-08-28
fix(staker): Introduce `Tier` type and simplify casting in reward logic

notJoon:
## Description

This PR introduces a dedicated `Tier` type and refactors related code to improve type safety and reduce unnecessary casting.

### Background

Previously, tier values were represented as raw `uint64`. This causes several issues:

- Frequent casting between `uint64` and `int64` when performing reward calculations (emssions and ratios are `int64`).
- Higher risk of accidental misuse, as any `uint64` value could be passed in as a tier without validation.
- Reduced clarity when reading code that mixed _tier as a nunber_ with _tier as a sementic concept_.

### Changes

- Added a new `Tier` type (`uint8`)
- Changed `CurrentCount` to return `int64` instead of int to align with reward calculation paths.
- Refactored method signatures (`CurrentReward`, `CurrentCount`, etc.) and internal logic to accept/use `Tier` instead of raw integers where appropriate


dongwon8247 (approved):
Approving. Please wait for @jinoosss's review before merging

PR #852, opened 2025-09-16
fix: `IsExternallyIncentivizedPool` to Check Active Incentives Only

notJoon:
## Description

Modified `IsExternallyIncentivizedPool` to check for active incentives at the current timestamp rather than just checking collection size.

When `EndExternalIncentive` is called, the incentive is marked as ended (`unclaimableRefunded = true`) but never removed from the pool's incentives collection. Therefore, it allowing users to stake positions in pools with no active reward programs.


jinoosss:
It seems fine not to change the test file location.

PR #853, opened 2025-09-16
fix: improve twap oracle prices

jinoosss:
## Descriptions

This PR addresses critical flaws in the TWAP oracle implementation that compromised price feed accuracy and could lead to manipulable or inaccurate price feeds for protocols relying on the oracle.

### Issues Fixed

**1. Incorrect Tick Cumulative Calculation**
- **Problem**: Oracle's `updatePriceCumulatives` was called after swap state was applied, causing tick cumulative to incorrectly use post-swap tick values
- **Fix**: Modified swap logic to conditionally update oracle with pre-swap tick values only when tick actually changes

**2. Missing Oracle Updates During Position Operations**
- **Problem**: When positions were modified through Mint/Burn operations where current tick falls within position range, active liquidity changed without oracle updates
- **Fix**: Removed redundant oracle calls from `tickTransition` to prevent duplicate updates and ensure proper state tracking

**3. Incomplete Historical Data Tracking**
- **Problem**: Implementation lacked tracking of historical values and `secondsPerLiquidity` metrics essential for robust TWAP calculations
- **Fix**: Implemented comprehensive `ObservationState` with circular buffer storing:
  - Historical tick cumulatives
  - Liquidity cumulatives  
  - Seconds per liquidity cumulative (X128 precision)
  - Block timestamps

**4. Limited Oracle Functionality**
- **Problem**: Oracle only updated after swap completion, missing tick crosses and liquidity changes
- **Fix**: Added new oracle methods:
  - `GetTWAP()`: Calculate time-weighted average price over specified period
  - `Observe()`: Retrieve historical cumulative data with interpolation
  - `IncreaseObservationCardinalityLimit()`: Expand historical data storage

### Technical Implementation

**New Oracle Architecture**
- Replaced simple `Observation` struct with robust `ObservationState` managing circular buffer
- Added support for up to 65,535 historical observations
- Implemented binary search and interpolation for historical queries
- Added proper error handling and edge case management

**Enhanced Data Structures**
```go
type ObservationState struct {
    observations     [65535]Observation
    index            uint16
    cardinality      uint16  
    cardinalityLimit uint16
}

type Observation struct {
    blockTimestamp                    int64
    tickCumulative                    int64
    liquidityCumulative               *u256.Uint
    secondsPerLiquidityCumulativeX128 *u256.Uint
    initialized                       bool
}
```

**Optimized Update Logic**
- Oracle updates now occur before state changes to capture pre-modification values
- Conditional updates prevent unnecessary writes when tick doesn't change
- Proper handling of zero liquidity scenarios and edge cases


contract/r/gnoswap/v1/pool/oracle.gno
```diff
@@ -0,0 +1,350 @@
...
+	index uint16,
+	liquidity *u256.Uint,
+	cardinality uint16,
+) ([]int64, []*u256.Uint, error) {
+	if len(secondsAgos) == 0 {
+		return nil, nil, errors.New("secondsAgos must not be empty")
+	}
+
+	tickCumulatives := make([]int64, len(secondsAgos))
+	secondsPerLiquidityCumulativeX128s := make([]*u256.Uint, len(secondsAgos))
```
notJoon, line 330: It seems better to store it in a variable and use it rather than calling `len(secondsAgos)` multiple times.

contract/r/gnoswap/v1/pool/pool_type.gno
```diff
@@ -209,64 +210,117 @@ func (p *Pool) Clone() *Pool {
...
+	// Check if we have enough observations
+	if p.observationState.cardinality == 0 {
+		return 0, errors.New("no observations available")
 	}
+
+	// Get observations for current time and secondsAgo
+	secondsAgos := []uint32{secondsAgo, 0}
+	currentTime := time.Now().Unix()
+
+	tickCumulatives, _, err := p.observationState.observe(currentTime, secondsAgos, 0, p.observationState.cardinality, p.liquidity, p.observationState.cardinality)
```
notJoon, line 290: When calling observe, tick is passed as 0, which appears to prevent reflecting the actual tick value.

When using the actual tick, `tickCumulative` seems to be calculated as follows:

```plain
tickCumulative = previousValue + actualTick * timeDelta
```

However, when tick is passed as 0, it will always use only the previous value, resulting in no change.

How about passing `p.slot0.tick` things instead of `0`?

notJoon (changes requested):

contract/r/gnoswap/v1/pool/pool_type.gno
jinoosss: a428470ba96696902b8a2c8a1b5de854cc2c9c6b

contract/r/gnoswap/v1/pool/oracle.gno
jinoosss: d2f5c99870c707835138620e11d0bbe8da0f0b05

notJoon (approved):

PR #863, opened 2025-09-20
fix(staker): Handle Precision Loss in External Incentive Rewards with rewardDust

notJoon:
## Description

The external incentive mechanism in the staker contract suffered from precision loss that permanently locked residual reward tokens. When `CreateExternalIncentive` was called, the reward distribution rate was calculated by dividing the total `rewardAmount` by the incentive duration through integer division to determine `rewardPerSecond`. This integer division truncated the decimal part, leaving a small amount of residual tokens that were neither distributed to incentivized positions nor tracked to be returned to the external incentive creator.

### Example
- Reward amount: 10,000,000,007 tokens
- Duration: 7,776,000 seconds (90 days)
- Calculation: 10,000,000,007 / 7,776,000 = 1,286 tokens/second
- **Lost tokens**: 10,000,000,007 - (1,286 × 7,776,000) = **64,007 tokens** (permanently locked)

## Solution

Implemented a dust refund mechanism by adding a `rewardDust` field to the `ExternalIncentive` struct that tracks the precision loss from integer division.

### Key Changes

1. **Added `rewardDust` field** to `ExternalIncentive` struct:
   ```go
   type ExternalIncentive struct {
       // ... existing fields
       rewardDust int64  // dust amount from precision loss (rewardAmount % duration)
   }
   ```

2. **Calculate dust during incentive creation** in `NewExternalIncentive`:
   ```go
   rewardPerSecond := rewardAmount / incentiveDuration
   rewardDust := rewardAmount - (rewardPerSecond * incentiveDuration)
   ```

3. **Include dust in refund** when ending incentive:
   ```go
   refund := incentive.rewardLeft
   refund += pool.incentives.calculateUnclaimableReward(incentive.incentiveId)
   refund += incentive.rewardDust  // Add dust amount to refund
   ```

## Test Results

Created a test (`external_incentive_dust_refund_filetest.gno`) that verifies the dust refund mechanism works correctly.

### Test Scenario
- Created an external incentive with **10,000,000,007 tokens** for **90 days**
- Expected dust: **64,007 tokens** (from 10,000,000,007 % 7,776,000)
- Staked a position to participate in the incentive
- Ended the incentive after the duration

### Verification Results
```
[INFO] create external incentive with reward amount: 10000000007
[INFO] duration: 7776000 seconds
[INFO] expected rewardPerSecond: 1286
[INFO] expected rewardDust: 64007

[INFO] bar balance before ending: 100000000
[INFO] bar balance after ending: 100070437
[RESULT] bar tokens refunded: 70437
[RESULT] gns tokens refunded: 1000000000
[SUCCESS] Dust refund mechanism verified - dust amount (64007) properly included in total refund (70437)!
```

### Analysis
- **Total refunded**: 70,437 tokens
  - Dust amount: 64,007 tokens
  - Unallocated rewards: 6,430 tokens
- **GNS deposit**: 1,000,000,000 tokens (fully refunded)

The test confirms that:
1. Precision loss is correctly calculated and stored in `rewardDust`
2. The dust amount is properly included in the refund when the incentive ends
3. No tokens are permanently locked due to integer division

### DIFF
--- a/contract/r/gnoswap/v1/staker/external_incentive.gno
+++ b/contract/r/gnoswap/v1/staker/external_incentive.gno
@@ -238,6 +238,8 @@ func endExternalIncentive(pool *Pool, incentiveId string, caller std.Address, cu
 	// Always calculate unclaimed rewards since this function is called only once
 	refund := incentive.rewardLeft
 	refund += pool.incentives.calculateUnclaimableReward(incentive.incentiveId)
+	// Add dust amount from precision loss to refund
+	refund += incentive.rewardDust
 
 	return incentive, refund, nil
 }
--- a/contract/r/gnoswap/v1/staker/type.gno
+++ b/contract/r/gnoswap/v1/staker/type.gno
@@ -19,6 +19,7 @@ type ExternalIncentive struct {
 	rewardAmount     int64       // total reward amount
 	rewardLeft       int64       // remaining reward amount
 	rewardPerSecond  int64       // reward per second
+	rewardDust       int64       // dust amount from precision loss (rewardAmount % duration)
 	refundee         std.Address // refundee address
 
 	refunded bool // whether incentive has been refunded (includes GNS deposit and unclaimed rewards)
@@ -96,6 +97,10 @@ func (e ExternalIncentive) RewardAmount() int64 {
 	return e.rewardAmount
 }
 
+func (e ExternalIncentive) RewardDust() int64 {
+	return e.rewardDust
+}
+
 func (self *ExternalIncentive) RewardSpent(currentTimestamp int64) int64 {
 	// Still check timestamps for state validation
 	if currentTimestamp < self.startTimestamp {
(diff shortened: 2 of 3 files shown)


notJoon:
> I propose a method to calculate the remaining rewards.
> 
> Utilize the staker pool's `calculateGlobalRewardRatioAccumulation` method to compute the actual reward quantity to be distributed. This is done by calculating the difference between the total reward ratio at the incentive end time and the reward ratio at the distribution start time.
> 
> Then, determine the remaining quantity by comparing this result with the quantity predicted to be distributed using `rewardPerSecond`.

@jinoosss 
I checked your comment. First, Global reward ratio accumulation is understood to be a mechanism for calculating rewards per position, serving to track the ratio of the entire pool's staked liquidity over time. 

However, the dust raised in the issue occurs from integer division at the time of incentive creation and is unrelated to the pool's state (liquidity, tick ...), so I believe the current fix is sufficient.

PR #885, opened 2025-09-24
chore: Apply missing overflow checks

notJoon:
## Overview

This PR addresses multiple identified uncaught overflow and underflows. The changes ensure safe arithmetic operations by using checked arithmetic functions (`AddOverflow`, `SubOverflow`) instead of unchecked operations.

## Excluded Changes and Rationale

- `Mul` functions
  - `MulOverflow` functions already exist for cases where overflow checking is needed
  - Changing the basic `Mul` behavior would be a breaking change affecting the entire codebase
  - Callers are responsible for choosing between `Mul` (unchecked) and `MulOverflow` (checked) based on their requirements

- `Lsh` functions in uint256/int256
  - The functions already handle shifts >= 256 bits correctly by returning 0
  - This is the expected behavior for bitwise operations in fixed-size integers

- Fee growth calculations
  - These calculations use **modular arithmetic** intentionally
  - This is the same pattern used in Uniswap V3 where underflow is the expected behavior
  - The calculations work correctly due to the properties of unsigned integer arithmetic in a fixed bit-width
  - Adding overflow checks would break the intended mathematical properties

- Fee growth difference calculations
   - **Rationale**: Same as tick.gno - uses modular arithmetic intentionally

-`abs` function (line 241-247)
  - The function is protected by `assertValidTickRange` which limits ticks to [-887272, 887272]
  - The problematic input value is impossible to reach in practice
  - Adding a check would add unnecessary overhead for an impossible case

### DIFF
--- a/contract/r/gnoswap/gns/getter.gno
+++ b/contract/r/gnoswap/gns/getter.gno
@@ -146,12 +146,18 @@ func GetEmissionEndTimestamp() int64 {
 func GetHalvingYearInfo(timestamp int64) (int64, int64, int64) {
 	state := getEmissionState()
 
-	endTimestamp := state.getEndTimestamp()
-	startTimestamp := state.getStartTimestamp()
+	year := state.getCurrentYear(timestamp)
 
-	year := getEmissionState().getCurrentYear(timestamp)
+	// If outside emission period, return 0 values
+	if year == 0 {
+		return 0, 0, 0
+	}
 
-	return year, startTimestamp + (SECONDS_IN_YEAR * year), endTimestamp
+	// Calculate year-specific start and end timestamps
+	yearStartTimestamp := state.getStartTimestamp() + (SECONDS_IN_YEAR * (year - 1))
+	yearEndTimestamp := yearStartTimestamp + SECONDS_IN_YEAR - 1
+
+	return year, yearStartTimestamp, yearEndTimestamp
 }
 
 // GetHalvingInfo returns comprehensive halving information as JSON string.
--- a/contract/r/gnoswap/gns/getter_test.gno
+++ b/contract/r/gnoswap/gns/getter_test.gno
@@ -1,0 +1,110 @@
+package gns
+
+import (
+	"testing"
+)
+
+// TestGetHalvingYearInfo verifies that GetHalvingYearInfo correctly calculates
+// year-specific start and end timestamps and maintains consistency with other functions.
+func TestGetHalvingYearInfo(t *testing.T) {
+	const testStartTimestamp = int64(1000)
+	emissionState = NewEmissionState(1, testStartTimestamp)
+	halvingData := emissionState.getHalvingData()
+
+	tests := []struct {
+		name              string
+		timestamp         int64
+		expectedYear      int64
+		expectedStartTime int64
+		expectedEndTime   int64
+	}{
+		{
+			name:              "Year 1 - Beginning",
+			timestamp:         testStartTimestamp,
+			expectedYear:      1,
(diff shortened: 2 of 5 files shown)


contract/p/gnoswap/uint256/bitwise.gno
jinoosss: This modification appears to be a validation check on the input value.
We must verify whether shifting this number results in an overflow or underflow.

notJoon: To check for actual overflow, we would need to examine all the most significant bits of the original value before the operation. However, this would be an unnecessary check in most use cases and could impact performance (this will make iterates all 256 bits in every call, even if the small value), which is why we've implemented the intermediate value checks as they are now.

I also considered using reverse operations to compare with the original value, but for left shift operations, the lower bits are filled with zeros, making direct comparison impossible. For example, left shifting 1 by 255 bits should be treated as a normal operation, but it could be lead false-positive results.

In 256-bit integers, shift operations are designed to discard upper/lower bits by design. The overflow checks in Lsh and Rsh operations serve more as safety guards, which is why I've applied the current approach as it seems sufficient for this purpose. How do you think?

contract/p/gnoswap/uint256/bitwise.gno
```diff
@@ -72,14 +72,23 @@ func (z *Uint) Lsh(x *Uint, n uint) *Uint {
...
+		if n >= 64 {
+			panic("Lsh: shift amount overflow")
+		}
 		goto sh128
 	case n > 64:
 		z.lsh64(x)
 		n -= 64
+		if n >= 64 {
+			panic("Lsh: shift amount overflow")
+		}
```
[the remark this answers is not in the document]
jinoosss: Everything is handled in this clause.
https://github.com/gnoswap-labs/gnoswap/pull/885#diff-71956ba46bbad3b1696f39c0db2a899022028b9ab8dc6627a86baed23eae22d1R70-R72

contract/p/gnoswap/uint256/bitwise.gno
```diff
@@ -46,8 +50,46 @@ func (z *Uint) Xor(x, y *Uint) *Uint {
...
+	if n == 0 || x.IsZero() {
+		if n >= 256 {
+			return z.Clear()
+		}
+		if n == 0 {
+			return z.Set(x)
+		}
+		// x == 0 or n > 0
+		return z.Clear()
+	}
```
jinoosss: Please simplify it.

```
if x.IsZero() {
	return z.Clear()
}

if n == 0 {
	return z.Set(x)
}
```

contract/p/gnoswap/uint256/bitwise_test.gno
```diff
@@ -252,28 +254,36 @@ func TestXor(t *testing.T) {
...
+		{"31337", 192, "196705537081812415096322133155058642481399512563169449530621952", false},
+		{"31337", 193, "393411074163624830192644266310117284962799025126338899061243904", false},
+		// 31337 requires 15 bits (binary: 111101001101001). When shifted left by 255 bits,
+		// it would require 15 + 255 = 270 bits total, which exceeds the 256-bit capacity.
+		// The maximum safe left shift for 31337 is 241 bits (256 - 15 = 241).
+		{"31337", 255, "0", true},
+		{"31337", 256, "0", true},
+
+		// no overflow, carry propagation check
+		{"18446744073709551615", 1, "36893488147419103230", false},
```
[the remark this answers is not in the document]
jinoosss: 1e7c16ea1a1322d37776e0bf7577538f038c89d3

PR #906, opened 2025-10-10
test(staker): Track unclaimable period when pool receives initial tier

notJoon:
## Description

Added test that checks `chageTier` to properly start the unclaimable period when a pool receives a tier for the first time with zero liquidity.


notJoon:
Screenshot 2025-10-15 at 1 54 22 PM

The router scenario test is failing because it appears that the gno RPC is not recognizing the package path again. I confirmed that there are no issues when running it locally.

PR #916, opened 2025-10-16
feat(router): implements router upgradable pattern

jinoosss:


contract/r/gnoswap/router/store.gno
```diff
@@ -0,0 +1,59 @@
...
+	return string(s)
+}
+
+const (
+	StoreKeySwapFee StoreKey = "swapFee" // Swap fee in basis points
+)
+
+const (
+	defaultSwapFeeBPS = uint64(15) // 0.15%
+)
```
notJoon: this constant is not being used. `defaultSwapFeeBPS` is defined in the `v1/init.gno` file and set in the `initStoreData` function, so I think this can be removed.

jinoosss: https://github.com/gnoswap-labs/gnoswap/pull/916/commits/d912587d542eb07fba7d8e3b659b5eb6df60c555

PR #930, opened 2025-10-30
feat(common): remove version control system from RBAC and Halt

jinoosss:
### Descriptions

This PR removes the version control mechanism from the RBAC and Halt systems, simplifying the contract architecture by eliminating version-specific function calls and handlers. This change reduces complexity while maintaining core functionality after the upgradable pattern implementation.

#### 1. Remove Version Control in RBAC and Halt (Commit: 02440b0a)

**RBAC Package Changes:**
- Removed `UpdateRoleAddressNextVersion` method and simplified `UpdateRoleAddress` to no longer require version parameter
- Eliminated version-related errors: `ErrVersionDoesNotExist` and `ErrCannotUpdateVersion`
- Removed methods: `GetRoleAddresses`, `GetRoleLatestVersion`, and `GetRoleVersionAddress`
- Simplified role address management by removing multi-version address tracking

**Halt Package Changes:**
- Removed version-based halt state management functions
- Eliminated `AssertIsNotHaltedXXXWithVersion` functions
- Removed getters for version-specific halt states
- Simplified halt status checking to single-version model

**Access Package Changes:**
- Removed swap whitelist functionality entirely
- Deleted files: `swap_whitelist.gno` and `swap_whitelist_test.gno`
- Removed version-specific assertion methods
- Significantly reduced test complexity (552 lines removed from `assert_test.gno`)

**Files changed:** 20 files
**Lines:** -1,741 / +195

#### 2. Update Contract Calls to Remove Version Parameters (Commit: 1188a86e)

Updated all contract calls across the codebase to remove `WithVersion` suffix:

**Updated Contracts:**
- `community_pool`: Changed `AssertIsNotHaltedCommunityPoolWithVersion(1)` to `AssertIsNotHaltedCommunityPool()`
- `gnft`: Updated all halt and access assertions to non-versioned calls
- `xgns`: Removed version parameters from halt checks
- `protocol_fee`: Simplified halt assertions
- `staker`: Updated both halt and access function calls

**Files changed:** 8 files
**Lines:** -39 / +29

#### 3. Remove Deprecated Methods from Governance Handlers (Commit: d9953f1e)

Cleaned up the governance parameter registry by removing obsolete handler methods:

**Removed Access Control Handlers:**
- `UpdateSwapWhiteList`
- `RemoveFromSwapWhiteList`

**Simplified Halt Handlers:**
- Removed `SetOperationStatusNextVersion`, kept simplified `SetOperationStatus` (now takes 2 params instead of 3)

[...]

PR #935, opened 2025-10-31
fix(access): prevent duplicate entries in GetWhitelistedSwaps

junghoon-vans:
## Summary
Fixed a bug in `GetWhitelistedSwaps` function that returned duplicate router addresses when the official router was also added to the swap whitelist.

## Problem
The `GetWhitelistedSwaps` function previously:
1. Added the official router first
2. Then iterated through all whitelisted routers and added them

This caused duplication when the official router address was also manually added to the whitelist map.

**Example scenario:**
```go
// Official router: 0xAAA
// Whitelist: [0xAAA, 0xBBB]
// Result: [0xAAA, 0xAAA, 0xBBB] ❌ (duplicate)
```

## Solution

Modified the logic to skip whitelisted routers that match the official router address:

```go
// Add whitelisted routers (skip if already added as official router)
for router := range swapWhitelist {
    if ok && router == officialRouter {
        continue // Skip duplicate
    }
    routers = append(routers, router)
}
```

After fix:

```
// Official router: 0xAAA
// Whitelist: [0xAAA, 0xBBB]
// Result: [0xAAA, 0xBBB] ✅ (no duplicate)
```

## Performance Improvement

- Time complexity: O(n) (single pass through whitelist)
- Space complexity: O(1) additional space

## Changes

Modified Files

- swap_whitelist.gno: Fixed GetWhitelistedSwaps to prevent duplicates
- swap_whitelist_test.gno: Added test case for duplication scenario

### DIFF
--- a/contract/r/gnoswap/access/swap_whitelist.gno
+++ b/contract/r/gnoswap/access/swap_whitelist.gno
@@ -68,8 +68,11 @@ func GetWhitelistedSwaps() []address {
 		routers = append(routers, officialRouter)
 	}
 
-	// Add whitelisted routers
+	// Add whitelisted routers (skip if already added as official router)
 	for router := range swapWhitelist {
+		if ok && router == officialRouter {
+			continue // Skip duplicate
+		}
 		routers = append(routers, router)
 	}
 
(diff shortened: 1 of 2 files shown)


jinoosss:
> contract/r/gnoswap/access/swap_whitelist.gno should be deleted as depreciated. Double check with @jinoosss 

Please check the contents of this PR. #930

junghoon-vans:
I've checked it. I'll close the PR now.

PR #948, opened 2025-11-03
txtar test

notJoon:
Added txtar tests for key functions in each contract. Due to limitations in the test environment where timestamps and block heights cannot be adjusted, tests for time-dependent functions were configured to verify that the calls fail as expected.

Changes other than test additions are as follows:
1. Emit contract update related events when calling `UpgradeImpl`
2. Update contract deployment scripts
3. Change test configuration to manage mnemonic settings via environment variables

## Test for Upgradable pattern

This PR also added tests related to the upgradable pattern. The cases covered by each file are as follows:

| File | What it verifies |
| --- | --- |
| `store_permission_enforcement.txtar` | Proves that the version manager revokes write access from old implementations. A probe package writes to the pool store while active, upgrades away, and then fails with `write permission denied` when it tries to mutate storage post-upgrade. |
| `rich_state_preservation.txtar` | Builds a WUGNOT/GNS pool, mints liquidity, accrues/collects fees, upgrades from `v1` to `v3_valid`, and repeats the workflow. Liquidity, owed-fee counters, and pool metadata are asserted before and after to ensure the proxy preserves live state through the upgrade. |
| `initializer_security.txtar` | Covers the version-manager guards. An out-of-domain package attempts to call `RegisterInitializer` (expected “caller is not in the domain path”), then an in-domain package registers once successfully and immediately hits the “initializer already registered” panic on a second attempt. |
| `broken_version_recovery.txtar` | Demonstrates how proposal can roll back from a faulty release. After confirming swaps succeed on `v1`, the test upgrades to `v2_invalid`, observes router-swap panics, and then upgrades to `v3_valid` to confirm functionality is restored without rebuilding pool state. |

### DIFF
--- a/tests/integration/testdata/upgradable/broken_version_recovery.txtar
+++ b/tests/integration/testdata/upgradable/broken_version_recovery.txtar
@@ -1,0 +1,92 @@
+loadpkg gno.land/p/nt/avl
+loadpkg gno.land/p/nt/ufmt
+loadpkg gno.land/p/onbloc/json
+loadpkg gno.land/p/nt/ownable
+loadpkg gno.land/p/demo/tokens/grc20
+loadpkg gno.land/p/demo/tokens/grc721
+loadpkg gno.land/r/demo/defi/foo20
+
+loadpkg gno.land/r/sys/users
+loadpkg gno.land/r/gnoland/wugnot
+loadpkg gno.land/r/demo/defi/grc20reg
+
+loadpkg gno.land/r/gnoswap/access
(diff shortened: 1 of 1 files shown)


dongwon8247:
@notJoon let's get this done first

notJoon:
@dongwon8247 I'm changing the approach from checking the entire data to checking the structure instead, because some data in the JSON generated from txtar is not deterministic (even though I masked all suspected parts, it still fails, I think the order changes occasionally). For example, when checking events for the Mint function, I'm modifying it to verify whether `Mint` exists in the event type.

PR #956, opened 2025-11-04
fix(common): send ugnot amount check methods

jinoosss:
## Descriptions

This PR fixes the UGNOT amount validation logic in the `common` package to properly handle zero-amount cases.

### Changes

**Modified Files:**
- `contract/r/gnoswap/common/assert.gno`
- `contract/r/gnoswap/common/coins.gno`

### What Changed

**1. AssertIsUserSendGNOTAmount function (assert.gno)**
- Removed the `&& amount > 0` condition from the validation check
- Previous behavior: Only validated non-zero amounts
- New behavior: Validates all amounts including zero

**2. isUserSendGNOTAmount function (coins.gno)**
- Refactored the logic to handle cases where no GNOT is sent
- Previous behavior: Returned `false` immediately when GNOT was not found in sent coins
- New behavior: Defaults `gnotAmount` to `0` when GNOT is not present, then compares with expected amount
- Updated documentation to remove the outdated comment about "No GNOT was sent" case

### Rationale

The previous implementation had an issue where:
- When `amount = 0` (expecting no GNOT to be sent), the validation was skipped due to the `amount > 0` check
- This could lead to inconsistent validation behavior

The fix ensures:
- Zero-amount validation works correctly
- When no GNOT is sent and `amount = 0` is expected, validation passes
- When no GNOT is sent but `amount > 0` is expected, validation fails
- More predictable and consistent validation behavior across all amount values

### DIFF
--- a/contract/r/gnoswap/common/coins_test.gno
+++ b/contract/r/gnoswap/common/coins_test.gno
@@ -21,7 +21,7 @@ func TestCoins_IsUserSendGNOTAmount(t *testing.T) {
 			name:           "no_coins_sent",
 			sendCoins:      map[string]int64{},
 			expectedAmount: 0,
-			expectedResult: false,
+			expectedResult: true,
 			description:    "No coins sent, expected amount 0 - should return false",
 		},
 		{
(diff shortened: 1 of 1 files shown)


junghoon-vans (approved):
👍

PR #967, opened 2025-11-07
test(p/fuzzutils): improve catch panics in fuzz runner

jinoosss:
## Summary

This PR enhances the panic handling mechanism in the fuzz test runner to properly catch both regular panics and cross-panics during fuzz test execution.

### Enhanced Panic Handling in `RunFuzzTestWithCrossPanic`

The `RunFuzzTestWithCrossPanic` function now handles both regular panics and cross-panics by nesting `handlePanic` and `handleCrossPanic` handlers:

- **Before**: Only `handleCrossPanic` was used, which could miss regular panics that occur during test execution
- **After**: Both `handlePanic` and `handleCrossPanic` are nested to ensure all panic types are properly caught and reported


junghoon-vans (approved):
👍

PR #984, opened 2025-11-09
fix(router): Remove redundant error checks

notJoon:
## Description

The max value is already checked earlier in `SetSwapFee`, there is no need to check it again.


contract/r/gnoswap/router/v1/protocol_fee_swap.gno
```diff
@@ -56,19 +57,6 @@ func (r *routerV1) SetSwapFee(fee uint64) {
 	)
 }
 
-// setSwapFee validates and updates the swap fee rate.
```
jinoosss: This logic should not be removed.

jinoosss (changes requested):

PR #990, opened 2025-11-12
fix(launchpad): add the missing collect reward function

jinoosss:
## Description

This PR adds the missing `CollectRewardByDepositId` function to the launchpad module, which was inadvertently omitted from the implementation.

### Interface Definition
- Added `CollectRewardByDepositId(depositID string) int64` method to the `ILaunchpadDeposit` interface in `types.gno`

### Proxy Layer
- Implemented proxy function `CollectRewardByDepositId` in `proxy.gno` that delegates calls to the underlying implementation
- Added proper documentation comment for the new function


dongwon8247:
👍 

can you check the CI?

PR #998, opened 2025-11-20
refactor(launchpad): manages data types in the domain contract

jinoosss:
## Descriptions

During contract upgrades, type conversion failures occurred due to incompatible data types between different contract versions. When upgrading from one version to another, the system attempted to cast stored data structures to the new version's types, which resulted in type mismatch errors such as "has no field rewardPerDepositX128".

### Changes

This PR resolves the upgrade type conversion issue by migrating core data structures to the domain contract level and having versioned contracts reference these shared types.

### Changes

**Commit 1: Move data structures to domain contract** 
- Moved core data structures from v1 to domain contract level:
  - `Deposit`: Manages user deposit information
  - `Project`: Handles project metadata and state
  - `ProjectTier`: Manages tier-specific configurations
  - `RewardManager`: Controls reward distribution logic
  - `RewardState`: Tracks reward state per deposit
- Created new domain contract files with proper encapsulation and getter/setter methods
- Refactored v1 contract to use domain-level data structures
- Updated 25 files (2,360 insertions, 1,843 deletions)

**Commit 2: Apply domain data type usage** 
- Updated v1 contract to explicitly use domain contract types:
  - Changed return types from local types to `launchpad.Deposit`, `launchpad.RewardState`, etc.
  - Added import for `gno.land/r/gnoswap/launchpad` package
  - Replaced direct method calls with helper functions for better abstraction
  - Updated all API, JSON builder, and test files to use domain types

### DIFF
--- a/contract/r/gnoswap/launchpad/project.gno
+++ b/contract/r/gnoswap/launchpad/project.gno
@@ -90,7 +90,7 @@ func (p *Project) Conditions() map[string]*ProjectCondition {
 		conditions[tokenPath] = condition.Clone()
 	}
 
-	return p.conditions
+	return conditions
 }
 
 // SetConditions sets the conditions map of the project.
@@ -106,7 +106,7 @@ func (p *Project) Tiers() map[int64]*ProjectTier {
 		tiers[duration] = tier.Clone()
 	}
 
-	return p.tiers
+	return tiers
 }
 
 // SetTiers sets the tiers map of the project.


contract/r/gnoswap/launchpad/counter.gno
```diff
@@ -13,10 +15,10 @@ func NewCounter() *Counter {
 }
 
 // Next increments the counter and returns the next ID.
-func (c *Counter) Next() int64 {
+func (c *Counter) Next() string {
```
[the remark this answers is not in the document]
jinoosss: 0b3f5ec3cfeb10fd799761358584e78e5b487dad

junghoon-vans (commented):
## Bug:  field not initialized in 

**Location**: `contract/r/gnoswap/launchpad/project.gno:140-158`

**Issue**: The `tiersRatios` map field is not initialized in the `NewProject` constructor function, while other map fields (`conditions`, `tiers`) are properly initialized with `make()`.

```gno
func NewProject(...) *Project {
    return &Project{
        ...
        conditions:    make(map[string]*ProjectCondition),
        tiers:         make(map[int64]*ProjectTier),
        // tiersRatios is missing initialization!
        createdHeight: createdHeight,
        createdAt:     createdAt,
    }
}
```

**Impact**: When `p.TiersRatios()` is called, it returns `nil`. Any subsequent operations attempting to read from or write to this map will cause a runtime panic.

**Suggested Fix**:
```gno
func NewProject(...) *Project {
    return &Project{
        ...
        conditions:    make(map[string]*ProjectCondition),
        tiers:         make(map[int64]*ProjectTier),
        tiersRatios:   make(map[int64]int64),  // Add this line
        createdHeight: createdHeight,
        createdAt:     createdAt,
    }
}
```

junghoon-vans (changes requested):
Found an issue that needs to be fixed

junghoon-vans (commented):
Reviewing code for potential issues

junghoon-vans (approved):

contract/r/gnoswap/launchpad/project.gno
```diff
@@ -0,0 +1,199 @@
...
+
+// GetConditions returns the conditions map of the project.
+func (p *Project) Conditions() map[string]*ProjectCondition {
+	conditions := make(map[string]*ProjectCondition)
+
+	for tokenPath, condition := range p.conditions {
+		conditions[tokenPath] = condition.Clone()
+	}
+
+	return p.conditions
```
junghoon-vans: **Critical Bug**: Wrong variable returned

Line 93 returns `p.conditions` (original) but should return `conditions` (cloned copy).

Same bug at line 109 in `Tiers()`.

contract/r/gnoswap/launchpad/project.gno
```diff
@@ -0,0 +1,199 @@
...
+
+// GetTiers returns the tiers map of the project.
+func (p *Project) Tiers() map[int64]*ProjectTier {
+	tiers := make(map[int64]*ProjectTier)
+
+	for duration, tier := range p.tiers {
+		tiers[duration] = tier.Clone()
+	}
+
+	return p.tiers
```
junghoon-vans: **Same bug**: Returns `p.tiers` instead of `tiers`

contract/r/gnoswap/launchpad/project.gno
jinoosss: fc01c83b976a67f572271671465611ab3155b051

junghoon-vans (approved):

PR #1006, opened 2025-11-26
fix(pool): update reentrancy lock in swap

jinoosss:
## Descriptions

The reentrancy lock mechanism in the `Swap` function was not properly applied to the actual pool state. The lock operations were only modifying a local copy (`slot0Start`) without updating the pool's state, making the reentrancy protection ineffective.

### Changes
**1. Apply lock to actual pool state at swap start:**
- Added `pool.SetSlot0(slot0Start)` after setting the unlock flag to false
- This ensures the reentrancy lock is properly reflected in the pool's state before executing the swap

**2. Release lock from actual pool state at swap end:**
- Changed from unlocking the local variable `slot0Start` to fetching the current pool state
- Now retrieves the latest slot0 state via `pool.Slot0()` in the defer function
- Sets unlock to true and applies it back to the pool state via `pool.SetSlot0(slot0End)`
- This ensures the lock is released on the actual current state, not a stale copy

### DIFF
--- a/contract/r/gnoswap/router/v1/_helper_test.gno
+++ b/contract/r/gnoswap/router/v1/_helper_test.gno
@@ -98,6 +98,8 @@ var (
 
 	registeredTestStore = false
 	registeredTestPool  = false
+
+	mockInstance *routerV1
 )
 
 func TokenFaucet(t *testing.T, tokenPath string, to address) {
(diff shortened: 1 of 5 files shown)


dongwon8247:
can you add a simple test to verify? call `Swap` inside the same pool as `swapCallback` -> unlocked == false (fail).

jinoosss:
Related test: https://github.com/gnoswap-labs/gnoswap/blob/d1931776b175fdaa636060e2cb096d25021c7768/contract/r/gnoswap/pool/v1/swap_test.gno#L433-L452

PR #1160, opened 2026-01-14
feat(gnodoc): generate docs from contract codebase

notJoon:
## Description

This PR introduces the full gnodoc pipeline to generate Markdown documentation directly from the contract codebase. It provides a local, source-driven alternative to `gno doc`, allowing documentation to be produced before deployment while preserving pkg.go.dev-style structure. The implementation covers parsing, modeling, rendering, and CLI behaviors so contracts can be documented consistently in development workflows.

### Implementation details

- Added .gno-aware parsing: when a package contains .gno files, the parser falls back to per-file parsing instead of ParseDir, which only handles .go files. This enables documentation generation for contract packages that are Gno-only.
- Implemented gnomod.toml module resolution: the parser detects module root via gnomod.toml and reads its module field to build ImportPath values used in Markdown. Source position filenames are normalized relative to the module root when available.
- Extracted examples from test files: examples are collected from *_test.go files and rendered in Markdown. Example output is trimmed for stable presentation.
- Added return-aware documentation: function/method docs include a Returns block based on named returns and return statement expressions. The renderer deduplicates repeated return expressions and renders nil once at the end.

### Markdown Output Example (Simplified)

~~~plain
# pkgname

`import "module/path/pkgname"`

Package overview text.

## Index
- [TypeA](#typea)
- [FuncX](#funcx)

## Constants
```go
const (
	MAX = 1
)
```

## Functions
### FuncX
```go
func FuncX(arg int) (result int, err error)
```
Description...

#### Returns
- named: result, err
- return: result, nil

## Types
### TypeA
```go
type TypeA struct
```
Description...

#### Methods
##### MethodY
```go
func (t *TypeA) MethodY() error
```

#### Returns
- return: nil
~~~

## Summary by CodeRabbit

* **New Features**
  * GnoDoc CLI to generate/export package documentation (export, list, default commands) with flags for output, tests inclusion, filtering, and source links.
  * Markdown renderer producing structured README output with stable anchors.

* **Parser & Model**
  * Rich package parser and documentation model capturing funcs, types, values, examples, notes, and indexes.

* **Documentation**
  * Comprehensive README describing usage, flags, output examples, and limitations.

* **Tests**
  * Extensive unit and integration tests covering CLI, parser, renderer, and model.

* **Chore**
  * .gitignore updated, module file and CLI entrypoint added.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


junghoon-vans:
I think it would be best to move this to a separate repository.

notJoon:
migrated to https://github.com/gnoswap-labs/gnoswap-docs-generator

PR #1196, opened 2026-03-18
fix: use mul-first ordering for TWAP tick interpolation

notJoon:
## Description

`observeSingle` was computing tick interpolation as `(delta / observationTimeDelta) * targetDelta`, causing integer division truncation before multiplication. This could produce up to hundreds of tick-seconds of error in interpolated `tickCumulative` values.

To fix this, switch to `(delta * targetDelta) / observationTimeDelta` via `u256.MulDiv` with sign seperation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved oracle interpolation precision using enhanced arithmetic calculations to prevent truncation and overflow errors in tick cumulative values.

* **Tests**
  * Added comprehensive test cases verifying interpolation accuracy and consistency for oracle observations.


jinoosss:
Maintain the current state.

PR #1198, opened 2026-03-19
fix: Apply CEI patterns

notJoon:
## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified abort reasons during transfers to report "insufficient balance" instead of overflow.
  * Fixed overflow checks in liquidity/burn flows to abort with clearer overflow diagnostics.

* **Improvements**
  * Reward collection now persists reward-manager state and commits collection state before transfers for greater consistency.
  * Withdrawal flow now reports explicit fee and post-fee amounts and centralizes withdrawal-fee handling.
  * Distribution updates simplified to accept four explicit percentage inputs.


jinoosss (commented):
It appears that the CEI pattern is being applied incorrectly.
Asset transfers must occur after the status update.

PR #1217, opened 2026-03-30
feat(staker): pool-scoped updates and skip redundant writes

jinoosss:
## Summary
This PR optimizes staker internal reward caching by switching from tier-wide cache updates to pool-scoped caching, and by skipping cache writes when the cached value doesn’t change at a given timestamp. It also removes unused “reward cache height” state across the staker storage/types/tests.

## Key Changes
- Pool-scoped reward caching: only compute/cache internal reward accumulators for the pool currently being evaluated.
- Change detection: avoid writing GlobalRewardRatioAccumulation / HistoricalTick when the input tick/value hasn’t changed.
- State cleanup: remove unused `PoolTierLastRewardCacheHeight` from store/types/mocks and related tests.

## Motivation
The previous approach updated reward caches across all tiered pools more broadly than needed, and could perform redundant writes at the same timestamp. Pool-scoped caching reduces unnecessary iterations, and change detection reduces storage writes.

## Summary by CodeRabbit

* **Refactor**
  * Pool-tier logic now uses timestamps only (chain-height removed).
  * Reward caching changed to per-pool scope and removed legacy height-based cursor.

* **Performance / Stability**
  * Reduced unnecessary state writes by only recording historical ticks when they change.

* **Tests**
  * Removed legacy height-based cache tests; added tests for per-pool caching and historical-tick behavior.


contract/r/gnoswap/staker/deposit.gno
```diff
@@ -43,6 +43,10 @@ func (d *Deposit) Liquidity() *u256.Uint {
 }
 
 func (d *Deposit) SetLiquidity(liquidity *u256.Uint) {
+	if liquidity == nil {
+		return
+	}
```
notJoon: passing `nil` seems invalid here. I think it would be better to panic rather than ignoring it

notJoon (changes requested):

contract/r/gnoswap/staker/v1/reward_calculation_pool_tier.gno
```diff
@@ -277,9 +273,75 @@ func (self *PoolTier) cacheReward(currentHeight int64, currentTimestamp int64, p
...
+	poolResolver := NewPoolResolver(pool)
+
+	if len(halvingTimestamps) == 0 {
+		// No emission change within the range => use current emission.
+		tierRewards := self.computeTierRewards(self.currentEmission)
+		poolReward, ok := tierRewards[tierNum]
+		if !ok {
+			return
+		}
+		poolResolver.cacheInternalReward(currentTimestamp, poolReward)
```
junghoon-vans: It seems like we could combine the logic that is repeated three times into a single function, similar to `applyCacheToAllPools`.

contract/r/gnoswap/staker/deposit.gno
jinoosss: [fa8e893](https://github.com/gnoswap-labs/gnoswap/pull/1217/commits/fa8e89300b002f418b319fa222276f4cc5832897)

notJoon (approved):

jinoosss (commented):
It must add a test that runs whenever a tick change occurs.
Additionally, verify that it behaves the same as the existing logic.

dongwon8247 (approved):

PR #1232, opened 2026-04-02
fix(staker): optimize `EndExternalIncentive` refund calculation

notJoon:
## Description

Replace the O(n) deposit iteration in `endExternalIncentive` with an O(1) refund calculation based on unclaimable period tracking.

### Changes

**Before:** 

`EndExternalIncentive` iterated over every deposit in the pool to compute uncollected rewards. An attacker could create thousands of minimal staked positions, making the transaction prohibitively expensive and blocking the incentive creator from recovering their remaining rewards and GNS deposit.

**After:** 

```
refund = unclaimableReward + remainder + collectedPenalties
```

- `unclaimableReward`: rewards for zero-liquidity periods (via `calculateUnclaimableReward`, already tracked by `modifyDeposit`)
- `remainder`: integer division truncation from `rewardPerSecond` (L-14 dust)
- `collectedPenalties`: warmup penalties retained in staker during `CollectReward`, derived from existing accounting fields without iteration: `(totalRewardAmount - rewardAmount) - distributedRewardAmount`
- A `maxRefund` cap (`totalRewardAmount - distributedRewardAmount`) prevents over-refund

### Trade-off

Per-deposit `MulDiv` floor rounding dust (2–5 units per incentive) now remains in the staker contract instead of being returned to the creator. This is because computing per-deposit rounding requires the iteration we're removing. The amounts are negligible and the direction is conservative (users' collectable rewards are never affected).

### Related Works

- #852: Precision loss in `rewardPerSecond` leading to dust rewards
- #894 fixed this by introducing the deposit iteration
- #1221: Changed external incentive penalty flow — penalties are no longer sent to community pool during `CollectReward`, but retained in staker and returned to creator via `EndExternalIncentive`. `distributedRewardAmount` now tracks rewards only (not penalties), which required adding `collectedPenalties` to the O(1) formula

## Summary by CodeRabbit

* **Bug Fixes**
  * More efficient O(1) external-incentive refund calculation with tighter bounds, proper handling of unclaimable amounts, remainders and collected penalties, and clamped refunds.

* **Tests**
  * Converted tests to deterministic forward-looking timelines.
  * Added coverage for zero-liquidity transitions.
  * Relaxed refund assertions to allow realistic tolerances.

### DIFF
--- a/contract/r/gnoswap/staker/v1/external_incentive.gno
+++ b/contract/r/gnoswap/staker/v1/external_incentive.gno
@@ -237,7 +237,17 @@ func (s *stakerV1) endExternalIncentive(resolver *PoolResolver, incentiveResolve
 		)
 	}
 
-	// O(1) refund: unclaimable reward during zero-liquidity periods plus division remainder.
+	// O(1) refund calculation:
+	//   refund = unclaimableReward + remainder + collectedPenalties
+	//
+	// - unclaimableReward: rewards for zero-liquidity periods (no depositors to earn them)
+	// - remainder: integer division truncation from rewardPerSecond (L-14 dust)
+	// - collectedPenalties: warmup penalties retained in staker during CollectReward,
+	//   routed back to creator on incentive end (PR #1221).
+	//   Derived from accounting fields:
+	//     rewardAmount decreases by (reward + penalty) on each collect,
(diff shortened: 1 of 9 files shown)


jinoosss:
Closed by #1290

PR #1236, opened 2026-04-03
fix(launchpad): remove current height param

notJoon:
## Summary by CodeRabbit

* **Breaking Changes**
  * Removed accumulated-height tracking from the launchpad reward system.
  * Removed the public getter for project-tier accumulated height.

* **Refactoring**
  * Reward accumulation and update logic now use timestamps only (height removed).
  * Internal APIs and call sites simplified accordingly.
  * Tests updated to reflect time-only semantics and updated constructors/calls.

PR #1238, opened 2026-04-07
refactor(launchpad): remove unused currentHeight params and fix naming

notJoon:
## Summary

Continue the cleanup started in #1236 by removing remaining unused `currentHeight` parameters and fixing misleading naming in launchpad contract tests.

### Removed parameters (dead code)

| Function | File | Reason |
|----------|------|--------|
| `collectDepositReward(deposit, currentHeight, currentTime)` | `launchpad_reward.gno` | `currentHeight` was accepted but never referenced in the function body |
| Test struct fields (`currentHeight`) in `UpdateRewardPerDepositX128`, `CollectReward`, `newRewardManager` | `reward_manager_test.gno` | Defined in test structs but never accessed via `tt.currentHeight` |

### Kept parameters (structurally required)

| Function / Struct | File | Reason |
|-------------------|------|--------|
| `createProjectParams.currentHeight` | `launchpad_project.gno` | Passed to `NewProject()` which uses it for `MakeProjectID(tokenPath, createdHeight)` — project ID generation depends on it |
| `NewDeposit(..., createdHeight, ...)` | `launchpad_deposit.gno` | Stored as `Deposit.createdHeight` for audit tracking |
| `withdrawDeposit(d, currentHeight, currentTime)` | `deposit.gno` | Passed to `Deposit.SetWithdrawn()` which sets `withdrawnHeight`, used by `IsWithdrawn()` |
| `withdrawDeposit(deposit, currentHeight, currentTime)` method | `launchpad_withdraw.gno` | Passes height through to the helper; caller provides `runtime.ChainHeight()` |

### Propagation fix

`runtime.ChainHeight()` is now called at the top-level entry point (`CollectDepositGns`) and passed down as a parameter, instead of being called redundantly in nested functions. This makes the call chain consistent: entry points capture height/time, inner functions receive them as arguments.

### Renamed test fields (`currentHeight` → `currentTime`)

Several test struct fields were named `currentHeight` but their values were passed to **time-based** functions (`isRewardStateClaimable`, `IsActivated`, `IsEnded`), which accept `currentTime int64`. Renamed for accuracy:

- `reward_state_test.gno` — field + test descriptions
- `project_tier_test.gno` — field name

## Summary by CodeRabbit

* **Bug Fixes**
  * Reward collection and withdrawals now use timestamps only (current time) instead of blockchain height for consistent accrual and outcomes.

* **Breaking Changes**
  * Height-based deposit getters and deposit height fields removed; deposit state and public APIs now expose/track time-only values.

* **Tests**
  * Test suite updated to validate timestamp-driven reward collection, withdrawal flows, and adjusted deposit constructors.


contract/r/gnoswap/launchpad/v1/launchpad_withdraw.gno
```diff
@@ -128,7 +128,7 @@ func (lp *launchpadV1) withdrawDeposit(deposit *launchpad.Deposit, currentHeight
 	project.SetTier(deposit.Tier(), projectTier)
 
 	// Finalize withdrawal
-	withdrawalAmount := withdrawDeposit(deposit, currentHeight, currentTime)
+	withdrawalAmount := withdrawDeposit(deposit, runtime.ChainHeight(), currentTime)
```
junghoon-vans: Why do chain height data need to be queried separately here?

notJoon: https://github.com/gnoswap-labs/gnoswap/pull/1238/commits/1c68dc9de8a3040d0a9587a38d4308063ec102b1

junghoon-vans (approved):

jinoosss (commented):
We should consider whether removing the height information from the deposit is the right approach.
From the perspective of contract users, this data may be necessary.
Additionally, since launchpad deposits are reversible, we should consider implementing this change.

PR #1248, opened 2026-04-10
fix(uint256): panic on zero modulo

junghoon-vans:
## Summary
- align `uint256` division and modulo helpers with Solidity/Uniswap-style fail-fast semantics for zero divisors/moduli
- make `Mod` and `MulMod` panic on zero modulus, keep `DivMod` fail-fast, and update unit/fuzz coverage accordingly
- simplify `sqrt_price_math` by switching the amount1 add helper from `MustDiv` to `Div` now that `Div` is fail-fast by default

## Verification
- make test PKG=gno.land/p/gnoswap/uint256
- make test PKG=gno.land/p/gnoswap/gnsmath

## Summary by CodeRabbit

* **Bug Fixes**
  * Division, modulo, and related modular operations now panic on zero divisors instead of returning zero, preventing silent incorrect results.

* **Refactor**
  * Removed redundant "Must" helper variants and consolidated division/modulo error behavior across math utilities.
  * Simplified zero-division checks to rely on unified division behavior.

* **Tests**
  * Updated tests and fuzz parameters to expect panics for zero divisors and adjusted assertions/messages accordingly.


notJoon (approved):
Since the existing logic was already written to panic in the case of a division by zero, it seems that the existing logic will become dead code.

However, the change still looks reasonable in that it prevents any unforeseen edge cases.

notJoon (approved):
LGTM 👍

jinoosss (approved):

PR #1249, opened 2026-04-10
fix(staker): apply warmup when summing collectable external reward for EndExternalIncentive refund

jinoosss:
## Summary

`EndExternalIncentive` computes the refund from remaining incentive budget after subtracting what stakers have effectively “earned” toward that incentive (`totalReward`) and any already-recorded `DistributedRewardAmount`. That `totalReward` is built by summing per-deposit external accrual via `RewardState.calculateCollectableExternalReward`.

Previously, `calculateCollectableExternalReward` updated per-warmup rewards but **did not** call `applyWarmup()` before aggregating. Gross accrual was therefore treated as fully attributable to stakers, so the protocol **under-refunded** creators: tokens corresponding to the warmup **penalty** (the portion users never collect) were left in the staker instead of being returned on end.

This change calls `applyWarmup()` in `calculateCollectableExternalReward` (aligned with `calculateExternalReward`), so the summed amount matches **post-warmup collectable** reward. The refund then correctly includes the penalty slice that was never user-collectable.

## Testing

- Updated scenario filetest golden outputs: `external_incentive_drain_filetest`, `external_incentive_dust_refund_filetest`, `single_gns_external_ends_filetest`.
- Added `external_incentive_dust_refund_with_unstake_before_filetest` and `external_incentive_dust_refund_with_unstake_after_filetest` to cover dust / division-remainder and unstake ordering around end-refund.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where reward calculations were not properly applying warmup ratio adjustments and penalties before returning collectible rewards.


jinoosss:
Closed by #1252

PR #1250, opened 2026-04-14
fix(router): handle exact out router fee

jinoosss:
## Summary

Fix ExactOut swap router fee double-inflation issue. Previously, the router inflated the user's `amountOut` via `calculateExactOutWithRouterFee` before requesting from the pool, then deducted fee from the pool result — causing users to receive more tokens than intended (amountOut + ~0.15% extra).

### Root Cause

The user's `amountOut` already includes the router fee (calculated by the API). The contract was inflating it a second time via `ExpectedExactAmountByFee`, leading to:
- Pool receiving an over-inflated request
- Fee deduction on the inflated pool result not matching the original user amount
- User receiving `amountOut` instead of `amountOut - fee`

### Changes

- **`base.gno`**: Remove `ExpectedExactAmountByFee` call in `createSwapOperation` for ExactOut. Pass `ExactAmount()` directly to the pool — the user's input already includes the fee.
- **`router.gno`**: Update `exactOutAmount` validation in `finalizeSwap` to compare against fee-deducted expected amount (`amountSpecified - routerFee`), since `amountSpecified` now includes the fee.
- **`router_dry.gno`**: Remove `calculateExactOutWithRouterFee` inflation in DrySwap. Pass user amount directly with sign flip only.
- **`router_test.gno`**: Add `TestFinalizeSwapExactOutWithFee` (4 cases covering 0.15%/1% fee, slippage error) and `TestCreateSwapOperationExactOutNoFeeInflation`.
- **`router_dry_test.gno`**: Update expected values for ExactOut DrySwap cases.
- **Filetests**: Update golden test expectations across 19 scenario files.

### ExactOut Flow (After Fix)

| Step | Action | Value |
|------|--------|-------|
| 1 | User input | amountOut = 100,150,225 (includes 0.15% fee, from API) |
| 2 | Contract requests from pool | 100,150,225 |
| 3 | Pool returns | ~100,150,225 |
| 4 | handleSwapFee | fee = 100,150,225 * 15/10000 = 150,225 -> protocol fee |
| 5 | User receives | 100,150,225 - 150,225 = 100,000,000 |

## Release Notes

* **Bug Fixes**
  * Corrected ExactOut swap fee handling so the exact output amount is calculated without additional fee inflation, with router fees applied separately from the specified amount.
  * Updated slippage validation to properly account for fee deduction in ExactOut swaps.

* **Tests**
  * Added comprehensive test coverage for ExactOut swap operations with router fee scenarios and slippage protection validation.

### DIFF
--- a/contract/r/gnoswap/router/v1/router_test.gno
+++ b/contract/r/gnoswap/router/v1/router_test.gno
@@ -169,6 +169,153 @@ func TestFinalizeSwap(t *testing.T) {
 	}
 }
 
+func TestFinalizeSwapExactOutWithFee(t *testing.T) {
+	tests := []struct {
+		name              string
+		resultAmountIn    int64
+		resultAmountOut   int64
+		tokenAmountLimit  int64
+		amountSpecified   int64
+		swapFee           uint64
+		swapCount         int64
+		expectedAmountOut int64
+		expectPanic       bool
+		errorMessage      string
+	}{
+		{
+			// User sends amountOut=100,150,225 (fee included).
+			// Pool returns 100,150,225. Fee = 100,150,225 * 15 / 10000 = 150,225.
+			// User receives 100,150,225 - 150,225 = 100,000,000.
+			name:              "ExactOut with 0.15% fee: user receives exact amount after fee deduction",
+			resultAmountIn:    72434254,
+			resultAmountOut:   100150225,
+			tokenAmountLimit:  80000000,
+			amountSpecified:   100150225,
+			swapFee:           15,
+			swapCount:         2,
+			expectedAmountOut: 100000000,
+			expectPanic:       false,
+		},
+		{
+			// 10,000 with 0.15% fee: fee = 10,000 * 15 / 10000 = 15.
+			// User receives 10,000 - 15 = 9,985.
+			name:              "ExactOut with 0.15% fee: small amount",
+			resultAmountIn:    12000,
+			resultAmountOut:   10000,
+			tokenAmountLimit:  15000,
+			amountSpecified:   10000,
+			swapFee:           15,
+			swapCount:         1,
+			expectedAmountOut: 9985,
+			expectPanic:       false,
+		},
+		{
+			// 1,000,000 with 1% fee: fee = 1,000,000 * 100 / 10000 = 10,000.
+			// User receives 1,000,000 - 10,000 = 990,000.
+			name:              "ExactOut with 1% fee: larger fee rate",
+			resultAmountIn:    1200000,
+			resultAmountOut:   1000000,
+			tokenAmountLimit:  1500000,
+			amountSpecified:   1000000,
+			swapFee:           100,
+			swapCount:         1,
+			expectedAmountOut: 990000,
+			expectPanic:       false,
+		},
+		{
+			// Slippage: amountIn exceeds tokenAmountLimit
+			name:             "ExactOut with fee: slippage error - spent too much",
+			resultAmountIn:   90000000,
+			resultAmountOut:  100150225,
+			tokenAmountLimit: 80000000,
+			amountSpecified:  100150225,
+			swapFee:          15,
+			swapCount:        2,
+			expectPanic:      true,
+			errorMessage:     "ExactOut: too much spent (max:80000000, used:90000000)",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			initRouterTest(t)
+
+			router := mockRouter()
+
+			testing.SetRealm(adminRealm)
+			func(cur realm) {
+				router.SetSwapFee(tt.swapFee)
+			}(cross)
+
+			// Transfer output token to router so handleSwapFee can send fee to protocol_fee
+			testing.SetRealm(testing.NewUserRealm(adminAddr))
+			bar.Transfer(cross, routerAddr, tt.resultAmountOut)
+
+			func(cur realm) {
+				testing.SetRealm(routerRealm)
+
+				if tt.expectPanic {
+					uassert.PanicsContains(t, tt.errorMessage, func() {
+						router.finalizeSwap(
+							bazPath,
+							barPath,
+							tt.resultAmountIn,
+							tt.resultAmountOut,
+							ExactOut,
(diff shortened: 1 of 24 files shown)


contract/r/scenario/router/exact_out_swap_route_with_wrapped_native_output_token_filetest.gno
```diff
@@ -293,11 +293,11 @@ func exactOutSwapRouteBy(
...
 // [INFO] queryRatios: 100
 // [INFO] maxAmountIn: 100000000000
-// [EXPECTED] inputTokenAmount: 1001565
-// [EXPECTED] outputTokenAmount: -1000000
+// [EXPECTED] inputTokenAmount: 1000073
+// [EXPECTED] outputTokenAmount: -998500
 // [EXPECTED] WUGNOT balance changed: 0
-// [EXPECTED] GNS balance changed: 1000000
-// [EXPECTED] BAR balance changed: -1001565
+// [EXPECTED] GNS balance changed: 998500
```
dongwon8247, line 299: This is a bad UX

dongwon8247: `amountOut` via `ExactOutSwapRoute` should match the user's input

contract/r/scenario/router/router_swap_route_1route_1hop_all_liquidity_exact_out_filetest.gno
```diff
@@ -217,4 +217,6 @@ func bazApprove(t *testing.T, owner, spender address, amount int64) {
 // [INFO] Attempting ExactOut swap (bar → 119820 baz) with fee is almost 120000
 // [INFO] Expected to fail due to insufficient liquidity
 // [EXPECTED] Swap should fail with error: [GNOSWAP-ROUTER-001] slippage check failed || ExactOut: too much spent (max:0, used:168406)
-// [INFO] Scenario completed
+
+// Error:
+// nil pointer dereference
```
dongwon8247, line 222: ?

dongwon8247: this should never happen. Investigate separately

dongwon8247 (changes requested):
This breaks the ExactOut semantics:
 
```
User requests:  amountOut = 1,000,000
Pool returns:   1,000,000
Fee deducted:   1,000,000 × 0.15% = 1,500
User receives:  998,500  ← NOT what they asked for
```
 
### Proposed Fix
 
**Move the gross-up into the contract. Remove it from the API.**
 
The contract inflates the pool request once, so that after fee deduction, the user receives exactly the specified amount:
 
```
User requests:  amountOut = 1,000,000
Contract:       grossAmount = 1,000,000 / (1 - 0.0015) = 1,001,502
Pool returns:   ~1,001,502
Fee deducted:   1,001,502 × 0.15% = 1,502 → protocol_fee
User receives:  1,000,000  ← exact
```
 
## Code Changes
 
### `base.gno` — restore single gross-up on raw amount
 
```go
case ExactOut:
    routerFee := r.store.GetSwapFee()
    grossAmount := calculateExactOutWithRouterFee(params.ExactAmount(), routerFee)
    pp := NewExactOutParams(baseParams, grossAmount, params.limitAmount)
    return NewExactOutSwapOperation(r, pp), nil
```
 
### `router.gno` — validate against user's original amount
 
```go
if swapType == ExactOut && !isSetSqrtPriceLimitX96 {
    if err := validator.exactOutAmount(
        resultAmountOutWithoutFee, amountSpecified, swapCount,
    ); err != nil {
        panic(addDetailToError(errSlippage, err.Error()))
    }
}
```
 
`amountSpecified` = user's raw request (pre-gross-up). No fee subtraction needed here.
 
### `router_dry.gno` — same gross-up for DrySwap consistency
 
```go
if swapType == ExactOut {
    grossAmount := calculateExactOutWithRouterFee(safeAbsInt64(amountSpecified), swapFee)
    amountSpecified = -grossAmount
}
```
 
Validation uses `originalAmountSpecified` (unchanged).
 
### API — remove gross-up.
 
API must pass the user's raw `amountOut` to the contract. If API still inflates, double-inflation recurs.

Investigate [this nill issue](https://github.com/gnoswap-labs/gnoswap/pull/1250/changes#diff-9dd88f7f105e1b65b927b58684be95f4eda3f3c2bdce9c81ce10a8771107327bR221) separately

junghoon-vans:
Could you also confirm the `TestCreateSwapOperationExactOutNoFeeInflation` coverage? As CodeRabbit pointed out, the current test does not actually exercise `createSwapOperation` and only re-checks `params.ExactAmount()`, so it would still pass even if the ExactOut path regressed back to fee inflation.

contract/r/gnoswap/router/v1/router_test.gno
```diff
@@ -169,6 +169,153 @@ func TestFinalizeSwap(t *testing.T) {
...
+		routeArr:    "gno.land/r/onbloc/bar:gno.land/r/onbloc/baz:3000",
+		quoteArr:    "100",
+		typ:         ExactOut,
+		exactAmount: 100150225,
+		limitAmount: 80000000,
+	}
+
+	// ExactAmount() should return the user's original value (no fee inflation)
+	uassert.Equal(t, int64(100150225), params.ExactAmount())
+}
```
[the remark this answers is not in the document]
junghoon-vans: It only checks params.ExactAmount(), so it would still pass even if the ExactOut path in base.gno regressed back to ExpectedExactAmountByFee(...).

jinoosss:
I'll keep the current contract functionality as is.
I'll close this PR and proceed with the API changes.

contract/r/scenario/router/router_swap_route_1route_1hop_all_liquidity_exact_out_filetest.gno
jinoosss: In this case, the test code was incorrectly modified during a previous code change.
As the actual quantity used changed, the error message failed to match.
a144e0149b3378172c7fec6db53e20bd8d8c662e

PR #1251, opened 2026-04-15
fix(staker): skip CollectReward state update when user reward is zero

jinoosss:
## Summary

Fix `CollectReward` to skip state updates (including `lastCollectTime`) when the user's warmup-applied reward is zero. Previously, when a position's reward was zero but a penalty existed (due to warmup truncation), the state was still updated — advancing `lastCollectTime` and effectively discarding the accrued period. This caused tiny positions to permanently lose small rewards that should have accumulated over time.

### Changes

**`contract/r/gnoswap/staker/v1/staker.gno`**

- **External rewards**: Changed the skip condition from `rewardAmount == 0 && penalty == 0` to `rewardAmount == 0`. When user reward is zero, the entire external incentive update is skipped regardless of penalty, preserving `lastCollectTime` so rewards continue to accrue.
- **Internal rewards (GNS)**: Added `skipInternalUpdate` flag. When `reward.Internal == 0`, the fee handling, `lastCollectTime` update, penalty transfer, and emission tracking are all skipped. This prevents zero-reward collections from advancing the internal reward clock.

### Scenario Tests (3 new filetests)

| File | What it tests |
|------|---------------|
| `collect_zero_reward_skip_external_filetest.gno` | Single external incentive: collects every 5 sec, observes `0→0→0→1→0→0→0→1→...` pattern (20-sec cycle) |
| `collect_zero_reward_skip_internal_filetest.gno` | Internal GNS emission: same pattern with `0→1→0→1→...` (10-sec cycle due to higher emission rate) |
| `collect_zero_reward_skip_external_multi_filetest.gno` | Two external incentives (bar 128/sec, qux 1286/sec): verifies skip/collect is independent per incentive — bar=0 while qux>0 at 5 sec |

### Key Behavior

- Zero-reward skip does **not** advance `lastCollectTime`, so skipped periods are included in the next successful collection.
- Each external incentive is evaluated independently — one can skip while another collects.
- After sufficient accumulation, the full accrued amount is delivered (no reward loss).

## Summary by CodeRabbit

* **Bug Fixes**
  * External incentive processing now continues correctly when a user's external reward is zero, preventing incorrect skips due to penalty checks or stale timestamps.

* **Refactor**
  * Reward collection avoids unnecessary internal payout steps, transfers, and protocol event emissions when there are no internal rewards, reducing redundant processing and noise in logs.

### DIFF
--- a/contract/r/gnoswap/staker/v1/staker.gno
+++ b/contract/r/gnoswap/staker/v1/staker.gno
@@ -712,7 +712,7 @@ func (s *stakerV1) CollectReward(positionId uint64) (string, string, map[string]
 			"poolPath", depositResolver.TargetPoolPath(),
 			"recipient", depositResolver.Owner().String(),
 			"rewardToken", GNS_PATH,
-			"rewardAmount", formatAnyInt(internalRewardToUser),
+			"rewardAmount", formatAnyInt(reward.Internal),
 			"rewardToUser", rewardToUser,
 			"rewardToFee", formatAnyInt(internalRewardToFee),
 			"rewardPenalty", rewardPenalty,
--- a/contract/r/gnoswap/staker/v1/staker.gno
+++ b/contract/r/gnoswap/staker/v1/staker.gno
@@ -606,6 +606,7 @@ func (s *stakerV1) CollectReward(positionId uint64) (string, string, map[string]
 		)
 	}
 
+	internalReward := int64(0)
 	internalRewardToUser := int64(0)
 	internalRewardToFee := int64(0)
 	internalRewardPenalty := int64(0)
@@ -622,6 +623,7 @@ func (s *stakerV1) CollectReward(positionId uint64) (string, string, map[string]
 			panic(err.Error())
 		}
 
+		internalReward = reward.Internal
 		internalRewardToUser = toUser
 		internalRewardToFee = feeAmount
 		internalRewardPenalty = reward.InternalPenalty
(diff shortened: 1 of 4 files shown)


contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -603,10 +606,25 @@ func (s *stakerV1) CollectReward(positionId uint64) (string, string, map[string]
 		)
 	}
 
+	internalRewardToUser := int64(0)
+	internalRewardToFee := int64(0)
```
dongwon8247, line 611: What's this fee for? unstaking fee?

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -642,67 +660,71 @@ func (s *stakerV1) CollectReward(positionId uint64) (string, string, map[string]
...
-	)
+		chain.Emit(
+			"CollectReward",
+			"prevAddr", previousRealm.Address().String(),
+			"prevRealm", previousRealm.PkgPath(),
+			"positionId", formatUint(positionId),
+			"poolPath", depositResolver.TargetPoolPath(),
+			"recipient", depositResolver.Owner().String(),
+			"rewardToken", GNS_PATH,
+			"rewardAmount", formatAnyInt(internalRewardToUser),
```
junghoon-vans: This looks like an event payload regression. rewardAmount in the internal CollectReward event used to mean the gross reward (reward.Internal), but after this change it emits the net-to-user amount (internalRewardToUser). That changes the field semantics and could break downstream consumers relying on the previous meaning.

contract/r/gnoswap/staker/v1/staker.gno
jinoosss: Yes, this is the unstaking fee for the user's internal rewards.

contract/r/gnoswap/staker/v1/staker.gno
jinoosss: I fixed events. 4ff70664ed52bba5bcb62e355b52276d923a36b4

contract/r/gnoswap/staker/v1/staker.gno
jinoosss: This variable is not stored in the store; it is a temporary variable used only within the function.

junghoon-vans (approved):

dongwon8247 (approved):

PR #1252, opened 2026-04-16
feat(staker): add CollectExternalIncentivePenalty function for penalty collection

jinoosss:
## Summary

This PR introduces explicit collection of external incentive warmup penalties and updates the external incentive refund flow to avoid double-accounting.

When `CollectReward` is called during warmup, the penalty portion is now accumulated per incentive and can be claimed separately by the incentive creator after the incentive is ended.

## Changes

- Added `accumulatedPenaltyAmount` to `ExternalIncentive` state.
- Added accumulation logic during `CollectReward`:
  - warmup penalty is tracked per incentive via `addAccumulatedPenaltyAmount`.
- Added new manager/proxy method:
  - `CollectExternalIncentivePenalty(targetPoolPath, incentiveId, refundAddress) int64`
- Added new getter:
  - `GetIncentiveAccumulatedPenaltyAmount(poolPath, incentiveId) int64`
- Implemented `CollectExternalIncentivePenalty` in `v1/external_incentive.gno` with:
  - pool/incentive existence checks
  - ended-incentive (`Refunded`) requirement
  - creator-only permission guard
  - zero-penalty no-op event emission
  - transfer amount capped by actual staker token balance
  - state update to reduce/reset accumulated penalty after transfer
- Refactored `endExternalIncentive` refund calculation:
  - switched to `unclaimableReward + remainder` model (bounded by max refundable amount)
  - added safety check for negative refund
- Updated interfaces/types/getters/proxy wiring to expose the new functionality.
- Added/updated scenario and unit tests around penalty accumulation and post-end collection flows.

## Motivation

Previously, warmup penalties were not exposed as a first-class collectible amount for external incentive creators, which made post-end settlement less explicit and harder to reason about.

This change:
- makes penalty accounting transparent,
- provides a dedicated collection path,
- and separates penalty collection from end/refund semantics.

## Summary by CodeRabbit

* **New Features**
  * Users can collect accumulated warmup penalties from ended external incentive programs.
  * Added a read-only query to view accumulated penalty amounts for external incentives.

* **Improvements**
  * Refund calculation for ending external incentives updated with safer, bounded arithmetic and explicit refund constraints.
  * Penalty amounts tracked and persisted when rewards are collected.

* **Tests**
  * Added tests for penalty collection, accounting invariants, refund calculations, and zero-liquidity edge cases.

### DIFF
--- a/contract/r/gnoswap/staker/v1/external_incentive.gno
+++ b/contract/r/gnoswap/staker/v1/external_incentive.gno
@@ -268,7 +268,7 @@ func (s *stakerV1) endExternalIncentive(resolver *PoolResolver, incentiveResolve
 // This function transfers the accumulated penalty to the specified refund address.
 // Returns the penalty amount collected.
 //
-// Only callable by the incentive creator.
+// Only callable by the incentive creator or admin.
 func (s *stakerV1) CollectExternalIncentivePenalty(
 	targetPoolPath string,
 	incentiveId string,
--- a/contract/r/gnoswap/staker/v1/external_incentive.gno
+++ b/contract/r/gnoswap/staker/v1/external_incentive.gno
@@ -301,7 +301,7 @@ func (s *stakerV1) CollectExternalIncentivePenalty(
 
 	if !incentiveResolver.Refunded() {
 		panic(makeErrorWithDetails(
-			errCannotEndIncentive,
+			errIsNotEndedIncentive,
 			ufmt.Sprintf("incentive(%s) must be ended first (call EndExternalIncentive)", incentiveId),
 		))
 	}
@@ -317,17 +317,6 @@ func (s *stakerV1) CollectExternalIncentivePenalty(
 
 	penaltyAmount := incentiveResolver.AccumulatedPenaltyAmount()
 	if penaltyAmount == 0 {
-		previousRealm := runtime.PreviousRealm()
-		chain.Emit(
-			"CollectExternalIncentivePenalty",
-			"prevAddr", previousRealm.Address().String(),
-			"prevRealm", previousRealm.PkgPath(),
-			"targetPoolPath", targetPoolPath,
-			"incentiveId", incentiveId,
-			"refundAddress", refundAddress.String(),
-			"penaltyAmount", formatAnyInt(int64(0)),
-		)
-
 		return 0
 	}
 
--- a/contract/r/gnoswap/staker/v1/external_incentive.gno
+++ b/contract/r/gnoswap/staker/v1/external_incentive.gno
@@ -307,11 +307,11 @@ func (s *stakerV1) CollectExternalIncentivePenalty(
 	}
 
 	caller := runtime.PreviousRealm().Address()
-	if caller != incentiveResolver.Creator() {
+	if !access.IsAuthorized(prbac.ROLE_ADMIN.String(), caller) && caller != incentiveResolver.Creator() {
+		adminAddr := access.MustGetAddress(prbac.ROLE_ADMIN.String())
 		panic(makeErrorWithDetails(
 			errNoPermission,
-			ufmt.Sprintf("only creator(%s) can collect penalty, called from %s",
-				incentiveResolver.Creator(), caller),
+			ufmt.Sprintf("only creator(%s) or admin(%s) can collect penalty, but called from %s", incentiveResolver.Creator(), adminAddr.String(), caller),
 		))
 	}
 
--- a/contract/r/gnoswap/staker/v1/external_incentive.gno
+++ b/contract/r/gnoswap/staker/v1/external_incentive.gno
@@ -3,7 +3,6 @@ package v1
 import (
 	"chain"
 	"chain/runtime"
-	"math"
 	"time"
 
 	prbac "gno.land/p/gnoswap/rbac"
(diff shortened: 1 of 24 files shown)


contract/r/gnoswap/staker/v1/external_incentive.gno
```diff
@@ -238,31 +237,117 @@ func (s *stakerV1) endExternalIncentive(resolver *PoolResolver, incentiveResolve
...
+	if !access.IsAuthorized(prbac.ROLE_ADMIN.String(), caller) && caller != incentiveResolver.Creator() {
+		adminAddr := access.MustGetAddress(prbac.ROLE_ADMIN.String())
+		panic(makeErrorWithDetails(
+			errNoPermission,
+			ufmt.Sprintf("only creator(%s) or admin(%s) can collect penalty, but called from %s", incentiveResolver.Creator(), adminAddr.String(), caller),
+		))
+	}
+
+	penaltyAmount := incentiveResolver.AccumulatedPenaltyAmount()
+	if penaltyAmount == 0 {
```
junghoon-vans, line 319: If `penaltyAmount == 0`, this returns before `chain.Emit`, so a valid no-op collection attempt leaves no audit trail in PR/event consumers. I think it would be better to still emit `CollectExternalIncentivePenalty` with `penaltyAmount=0` before returning.

Also, the function doc comment above still says creator-only, but the implementation now allows admin as well. Please update that comment to say `creator or admin` so the contract API comment matches the permission check.

jinoosss: Why is an event logged when the penalty count is 0?

contract/r/scenario/staker/collect_external_incentive_penalty_filetest.gno
```diff
@@ -0,0 +1,347 @@
+// PKGPATH: gno.land/r/gnoswap/v1/main
+
+// SCENARIO:
+// Tests CollectExternalIncentivePenalty function:
+// 1. Create pool and external incentive (QUX reward, 90 days)
+// 2. Stake position and collect rewards multiple times (accumulates warmup penalty)
+// 3. End external incentive
+// 4. Admin collects accumulated penalty → community pool
```
junghoon-vans: The scenario description says `Admin collects accumulated penalty -> community pool`, but this file actually executes `collectPenalty()` as `externalCreatorRealm`, refunds to `externalCreatorUser`, and the golden output below reports creator QUX increasing. It would be better to update the scenario text so the documented actor/destination matches the behavior this filetest really covers.

contract/r/scenario/staker/collect_external_incentive_penalty_filetest.gno
```diff
@@ -0,0 +1,347 @@
...
+		cross,
+		"gno.land/r/onbloc/bar:gno.land/r/onbloc/qux:100",
+		actualIncentiveID,
+		externalCreatorUser,
+	)
+
+	ufmt.Printf("[INFO] second collection returned: %d\n", collected)
+
+	if collected == 0 {
+		println("[EXPECTED] second collection returns 0 (no more penalty)")
```
junghoon-vans: This expectation reads like the second penalty collection should be a no-op, but the checked-in golden output for this scenario shows a positive second collection amount instead. I think this scenario should make the intended contract explicit: either verify `0` here, or rename/reword the step to say it is validating that new penalty can accrue again after later `CollectReward` calls.

contract/r/scenario/staker/collect_external_incentive_penalty_with_multi_staking_filetest.gno
```diff
@@ -0,0 +1,368 @@
+// PKGPATH: gno.land/r/gnoswap/v1/main
+
+// SCENARIO:
+// Tests CollectExternalIncentivePenalty function:
+// 1. Create pool and external incentive (QUX reward, 90 days)
+// 2. Stake position and collect rewards multiple times (accumulates warmup penalty)
+// 3. End external incentive
+// 4. Admin collects accumulated penalty → community pool
```
junghoon-vans: Same issue here as the single-staking scenario: the header says `Admin collects accumulated penalty -> community pool`, but the actual scenario runs penalty collection as the external creator and the output shows creator QUX increasing. Updating the scenario prose would make the filetest much easier to trust as documentation.

contract/r/scenario/staker/external_incentive_complete_distribution_filetest.gno
```diff
@@ -492,19 +492,18 @@ func positionIdFrom(positionId int) grc721.TokenID {
...
 // [EXPECTED] creator receives all remaining reward token: 1512000003
 // [EXPECTED] creator receives fixed GNS deposit: 1000000000
 // [EXPECTED] Total reward distribution:
 //   - Initial reward amount: 7776000000 QUX
 //   - Rewards collected by staker: 6263999997 QUX
-//   - Undistributed reward refunded: 1512000003 QUX
-//   - Total accounted for: 7776000000 QUX
-// [EXPECTED] all rewards accounted for with no loss
+//   - Undistributed reward refunded: 0 QUX
+//   - Total accounted for: 6263999997 QUX
```
junghoon-vans, line 516: The updated golden output here no longer matches the scenario narrative at the top of the file. The scenario still says `EndExternalIncentive` refunds all undistributed reward and that `distributed + refunded = total reward`, but this checked-in block now records `Undistributed reward refunded: 0` and `Total accounted for: 6263999997 QUX` instead. I think the scenario description and the expected/output section should be brought back into sync so the filetest documents one consistent accounting model.

jinoosss: 54bcf20124ead9e59e9cbacab3b31a194202b3ae

contract/r/scenario/staker/collect_external_incentive_penalty_with_multi_staking_filetest.gno
jinoosss: 54bcf20124ead9e59e9cbacab3b31a194202b3ae

contract/r/scenario/staker/collect_external_incentive_penalty_filetest.gno
jinoosss: 54bcf20124ead9e59e9cbacab3b31a194202b3ae

contract/r/scenario/staker/collect_external_incentive_penalty_filetest.gno
jinoosss: 54bcf20124ead9e59e9cbacab3b31a194202b3ae

contract/r/scenario/staker/external_incentive_complete_distribution_filetest.gno
```diff
@@ -408,8 +413,14 @@ func endExternalIncentiveAndVerifyNoDust() {
 	totalDistributed := totalRewardsCollected + undistributedRewardRefunded
 	ufmt.Printf("  - Total accounted for: %d QUX\n", totalDistributed)
 
-	if totalDistributed == REWARD_AMOUNT {
-		println("[EXPECTED] all rewards accounted for with no loss")
+	// Leftover QUX retained in the staker contract (accumulated penalty share).
+	// This is NOT refunded by EndExternalIncentive; it must be collected via
+	// CollectExternalIncentivePenalty.
+	stakerRetainedPenalty := stakerQuxAfter
+	ufmt.Printf("  - Staker-retained penalty QUX: %d\n", stakerRetainedPenalty)
```
junghoon-vans, line 420: The new retained-penalty explanation here is good, but there is still one stale expected line a few lines above: `[EXPECTED] creator receives all remaining reward token: ...`. With the updated accounting model in this block, the remaining QUX stays in the staker contract and is **not** refunded by `EndExternalIncentive`, so that earlier line should be reworded or removed to match this output section.

junghoon-vans (approved):

PR #1255, opened 2026-04-20
refactor(gov/governance): remove unused param in hasActiveProposal

junghoon-vans:
## Summary
- remove the unused `current` parameter from `hasActiveProposal` now that proposal activity is already determined during stale-proposal cleanup
- update all governance proposal entrypoints to keep the existing prune-then-check flow while calling the helper with its actual dependency surface
- extend the instance cleanup test to assert `hasActiveProposal` matches the remaining proposal set after inactive proposals are removed

## Testing
- make test PKG=gno.land/r/gnoswap/gov/governance/v1 RUN=TestInstance_RemoveInactiveUserProposals
- make test PKG=gno.land/r/gnoswap/gov/governance/v1

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined proposal eligibility checks: activity is now determined by the presence of existing proposals, reducing false negatives and preventing conflicting submissions.
* **Tests**
  * Updated governance tests to cover pre-voting and executable proposal scenarios and to verify the revised active-proposal behavior.

### DIFF
--- a/contract/r/gnoswap/gov/governance/v1/instance_test.gno
+++ b/contract/r/gnoswap/gov/governance/v1/instance_test.gno
@@ -53,6 +53,7 @@ func TestInstance_RemoveInactiveUserProposals(t *testing.T) {
 		currentTime            int64
 		expectedRemainingCount int
 		expectedRemainingIDs   []int64
+		expectedHasActive      bool
 	}{
 		{
 			name:                   "No proposals for user",
@@ -60,6 +61,7 @@ func TestInstance_RemoveInactiveUserProposals(t *testing.T) {
 			currentTime:            baseCreatedAt,
 			expectedRemainingCount: 0,
 			expectedRemainingIDs:   nil,
+			expectedHasActive:      false,
 		},
 		{
 			name: "All proposals are active - none should be removed",
@@ -77,6 +79,7 @@ func TestInstance_RemoveInactiveUserProposals(t *testing.T) {
 			currentTime:            baseCreatedAt + config.VotingStartDelay + 50, // during voting period
 			expectedRemainingCount: 1,
 			expectedRemainingIDs:   []int64{1},
+			expectedHasActive:      true,
 		},
 		{
 			name: "All proposals are expired - all should be removed",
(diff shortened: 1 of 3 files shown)


jinoosss (approved):

contract/r/gnoswap/gov/governance/v1/instance_test.gno
```diff
@@ -200,6 +208,7 @@ func TestInstance_RemoveInactiveUserProposals(t *testing.T) {
 
 			proposals := gv.getUserProposals(proposerAddr.String())
 			uassert.Equal(t, len(proposals), tc.expectedRemainingCount)
+			uassert.Equal(t, gv.hasActiveProposal(proposerAddr), tc.expectedHasActive)
```
junghoon-vans, line 256: [review-agent] **[note] Test gap: Upcoming and Executable states not covered by `hasActiveProposal` post-prune assertion**

The new `expectedHasActive` assertion covers: no proposals, active (voting), expired, mixed, canceled, multiple expired/active. However `ProposalResolver.IsActive` also returns `true` for `StatusUpcoming` (pre-vote-start delay) and `StatusExecutable` (post-vote awaiting execution window). Adding two cases would close this gap:

```
// Upcoming: currentTime < activeAt — proposer should be blocked
currentTime: baseCreatedAt + config.VotingStartDelay - 10,
expectedHasActive: true,

// Executable: currentTime in execution window — proposer should be blocked
currentTime: expiredTime - 50,
expectedHasActive: true,
```

Not a defect — the logic is correct by inspection — but the extra cases would provide direct regression protection for the full "any live state blocks the proposer" invariant.

junghoon-vans (commented):
[review-agent] PR #1255 — automated review summary

**Verdict: safe to merge.** This is a clean dead-parameter removal with no correctness regression.

---

**What was verified**

The old `hasActiveProposal(proposerAddress address, current int64)` body was `return len(getUserProposals(addr)) > 0` — `current` was provably never read. The active-state filtering was entirely performed by `removeInactiveUserProposals(addr, current)`, which always runs immediately before `hasActiveProposal` at all three call sites. The PR makes the contract explicit and removes the misleading parameter.

`hasActiveProposal` is a private method on `governanceV1` (not in the governance interface), so the signature change has zero impact outside the package.

---

**[note] Text proposals in StatusPassed immediately release the one-per-address lock**

`ProposalResolver.IsActive` (`proposal.gno:43-46`) returns `false` for text proposals in `StatusPassed` (`return !r.IsTextType()`). Once a text proposal vote passes, `removeInactiveUserProposals` removes it from the user list and the proposer can submit again immediately — no wait for an execution window.

This is intentional and correct (text proposals have no execution phase), but is a non-obvious asymmetry vs. executable proposal types (CommunityPoolSpend, ParameterChange), which remain active through `StatusExecutable` until executed or expired. No action required.

PR #1256, opened 2026-04-20
fix(referral): rate limit cancellation

junghoon-vans:
## Summary
- apply the referral cooldown check to cancellation requests as well, and keep updating `lastOps` on successful removal
- add regression coverage for blocked immediate cancellation, allowed cancellation after cooldown, and blocked re-registration after removal
- update referral docs to describe the current cooldown and fallback semantics

## Verification
- make test PKG=gno.land/r/gnoswap/referral

## Release Notes

* **New Features**
  * Added `GetLastOpTimestamp` to retrieve the timestamp of the last successful referral operation

* **API Changes**
  * `TryRegister` now returns the effective referrer address (string) instead of a boolean success flag

* **Improvements**
  * Rate limiting now covers registration, updates, and removal operations
  * Successful removal operations now refresh the cooldown timestamp, preventing immediate re-registration
  * Removal mechanism now triggered via contract address instead of zero address

* **Documentation**
  * Updated referral contract documentation to reflect new API and rate-limiting behavior

### DIFF
--- a/contract/r/gnoswap/referral/keeper.gno
+++ b/contract/r/gnoswap/referral/keeper.gno
@@ -38,19 +38,21 @@ func (k *keeper) register(addr, refAddr address) (address, error) {
 	addrStr := addr.String()
 	refAddrStr := refAddr.String()
 
+	if err := k.checkRateLimit(addrStr); err != nil {
+		return zeroAddress, err
+	}
+
 	if isRemovalRequest(refAddr) {
 		if k.has(addr) {
 			_, ok := k.store.Remove(addrStr)
 			if !ok {
 				return zeroAddress, ErrNotFound
 			}
+
+			k.lastOps.Set(addrStr, time.Now().Unix())
 		}
 
 		return zeroAddress, nil
-	}
-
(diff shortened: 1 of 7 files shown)


contract/r/gnoswap/referral/keeper.gno
```diff
@@ -38,21 +38,23 @@ func (k *keeper) register(addr, refAddr address) (address, error) {
 	addrStr := addr.String()
 	refAddrStr := refAddr.String()
 
+	if err := k.checkRateLimit(addrStr); err != nil {
+		return zeroAddress, err
+	}
+
```
jinoosss, line 44: Previously, registration and cancellation were not treated as the same operation.
However, under these conditions, registration and cancellation are now treated as the same operation, and the policy has been updated to allow cancellation 24 hours after registration.
Please verify that this policy is correct.

junghoon-vans: I've confirmed this. I'll make the necessary changes so that the cooldown does not apply when canceling.

PR #1259, opened 2026-04-21
refactor(gov/governance): enhance invalid parameter proposals

junghoon-vans:
## Summary
- reject malformed `CollectProtocol` proposal params at proposal creation time instead of failing during execution
- reject semantically invalid RBAC proposal actions such as updating the admin role or removing system roles
- add governance proposal and parameter registry coverage for the new early-validation paths

## Testing
- make test PKG=gno.land/r/gnoswap/gov/governance/v1

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened proposal validation: enforce non-negative numeric values for protocol-collection parameters and allow one-sided zero requests.
  * Enhanced RBAC constraints to reject empty role names and to block updating or removing protected admin/system roles.

* **Tests**
  * Expanded test coverage for parameter parsing, one-sided requests, and RBAC edge cases (invalid amounts, empty names, admin immutability).

### DIFF
--- a/contract/r/gnoswap/gov/governance/v1/parameter_registry.gno
+++ b/contract/r/gnoswap/gov/governance/v1/parameter_registry.gno
@@ -3,6 +3,7 @@ package v1
 import (
 	"strings"
 
+	prbac "gno.land/p/gnoswap/rbac"
 	avl "gno.land/p/nt/avl/v0"
 	ufmt "gno.land/p/nt/ufmt/v0"
 
@@ -346,8 +347,8 @@ func CreateParameterHandlers() *ParameterRegistry {
 				stringValidator,  // token1Path
 				uint64Validator,  // fee
 				addressValidator, // recipient
-				nil,              // amount0Requested (validated in pool logic)
-				nil,              // amount1Requested (validated in pool logic)
+				nonNegativeInt64Validator("amount0Requested"),
+				nonNegativeInt64Validator("amount1Requested"),
 			},
 			handlerFunc: func(params []string) error {
 				pl.CollectProtocol(
@@ -616,7 +617,7 @@ func CreateParameterHandlers() *ParameterRegistry {
 			function:   "RegisterRole",
 			paramCount: 2,
 			paramValidators: []paramValidator{
-				stringValidator,  // roleName
+				roleNameValidator,
 				addressValidator, // roleAddress
 			},
 			handlerFunc: func(params []string) error {
@@ -634,7 +635,7 @@ func CreateParameterHandlers() *ParameterRegistry {
 			function:   "UpdateRoleAddress",
 			paramCount: 2,
 			paramValidators: []paramValidator{
-				stringValidator,  // roleName
+				updatableRoleNameValidator,
 				addressValidator, // roleAddress
 			},
 			handlerFunc: func(params []string) error {
@@ -652,7 +653,7 @@ func CreateParameterHandlers() *ParameterRegistry {
 			function:   "RemoveRole",
 			paramCount: 1,
 			paramValidators: []paramValidator{
-				stringValidator, // roleName
+				removableRoleNameValidator,
 			},
 			handlerFunc: func(params []string) error {
 				roleName := params[0]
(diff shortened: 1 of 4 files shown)


jinoosss (approved):

junghoon-vans (commented):
[review-agent] PR #1259 — automated review summary

**Verdict: safe to merge.** The validation improvements are correct; malformed proposals are now rejected at creation time rather than failing at execution.

---

**What was verified**

- `nonNegativeInt64Validator` — `parseInt64` panics on bad input; `runValidator` correctly catches and returns the error. Chain is sound. Test expectations match exact panic strings from `utils.gno`.
- `roleNameValidator` / `updatableRoleNameValidator` / `removableRoleNameValidator` — whitespace trimming is applied before all comparisons; padded inputs (e.g. `"  admin  "`) are correctly caught.
- `runValidator` deferred recover converts all panics to errors, including nested `panic(err)` inside `updatableRoleNameValidator`.
- Success test update (`RemoveRole*EXE*admin` → `RegisterRole*EXE*new_role,<addr>`) is correct — the old value would now fail the new system-role guard.

---

**[note] Governance can update any system role address except `admin` — including the `governance` role itself**

`updatableRoleNameValidator` only blocks `admin` from `UpdateRoleAddress`, while `removableRoleNameValidator` blocks all 13 system roles from `RemoveRole`. This means governance proposals can redirect the address of `devops`, `governance`, `gov_staker`, `pool`, `router`, and the other system roles.

This is **confirmed intentional** by the GnoSwap trust-authority model (chunk `kb_ec176b040abf96cfffbec2d4`):
> "Treat admin and governance as high-impact trusted actors… role registration and role-address updates" are a privileged governance capability.

Governance is sovereign over system role addresses (needed for contract upgrades). Only admin is immutable because admin is the root authority above governance.

**Suggested follow-up (doc only):** Add a comment in `updatableRoleNameValidator` noting that governance intentionally retains authority to update all non-admin system role addresses. This prevents future reviewers from reading the missing `IsSystemRole()` check as an oversight.

contract/r/gnoswap/gov/governance/v1/parameter_registry.gno
```diff
@@ -927,6 +928,52 @@ func uint8RangeValidator(name string) paramValidator {
...
+			panic(ufmt.Sprintf("system role cannot be removed: %s", roleName))
+		}
+	})
+}
+
+func nonNegativeInt64Validator(name string) paramValidator {
+	return func(s string) error {
+		return runValidator(func() {
+			value := parseInt64(s)
+			if value < 0 {
```
junghoon-vans, line 970: [review-agent] **[note] `nonNegativeInt64Validator` accepts `amount = 0` — confirm `CollectProtocol` semantics**

The validator passes `value == 0`, so a governance proposal with `amount0Requested=0` or `amount1Requested=0` will succeed creation and, if passed, invoke `pool.CollectProtocol` with a zero-amount request.

In Uniswap V3 / GnoSwap pool semantics, `amountRequested=0` could mean "collect nothing" (no-op) or "collect all" depending on implementation. If it is a true no-op, a zero-amount proposal wastes governance resources without effect. Consider documenting the expected behavior, or adding a positive-only check if zero-amount collection is not useful.

contract/r/gnoswap/gov/governance/v1/parameter_registry.gno
```diff
@@ -927,6 +928,52 @@ func uint8RangeValidator(name string) paramValidator {
...
+		if roleName == "" {
+			panic(ufmt.Sprintf("role name is empty: %q", s))
+		}
+	})
+}
+
+func updatableRoleNameValidator(s string) error {
+	return runValidator(func() {
+		roleName := strings.TrimSpace(s)
+		if err := roleNameValidator(s); err != nil {
```
junghoon-vans: [review-agent] **[note] Style inconsistency — pass `roleName` (already trimmed) instead of `s` to `roleNameValidator`**

```go
roleName := strings.TrimSpace(s)
if err := roleNameValidator(s); err != nil {  // passes original s, not roleName
```

`roleNameValidator` performs its own `TrimSpace` internally, so this is functionally correct. However, passing the already-trimmed `roleName` would be more consistent and make the control flow easier to follow. Same pattern applies in `removableRoleNameValidator` (line 956).

junghoon-vans: [91cfd2c](https://github.com/gnoswap-labs/gnoswap/pull/1259/commits/91cfd2ca95207155ac2ed8631baaedc9c234e4d7)

contract/r/gnoswap/gov/governance/v1/parameter_registry.gno
junghoon-vans: [91cfd2c](https://github.com/gnoswap-labs/gnoswap/pull/1259/commits/91cfd2ca95207155ac2ed8631baaedc9c234e4d7)

jinoosss (approved):

notJoon (approved):

PR #1263, opened 2026-04-21
refactor(staker): unify reward `lastCollectTime` invariants

notJoon:
## Description

This PR makes the staker reward cursor invariant explicit and symmetric across internal GNS rewards and external incentive rewards.

Both reward streams maintains a `lastCollectTime` cursor. That cursor is expected to move monotonically forward from the effective last collection time, where an unset to zero cursor falls back to the deposit's `StakeTime`. Before this change, the internal and external reward paths enforced that rule through different read paths, which made the invariant weaker for external rewards in defensive cases.

### Problem

`CollectReward` settles two independent reward streams:

- internal GNS emissions
- external incentive token rewards

Both streams update a `lastCollectTime` cursor to prevent collecting rewards over an invalid earlier interval. However the update paths were asymmetric:

- Internal reward updates used `DepositResolver.InternalRewardLastCollectTime()`, which falls back to `StakeTime()` when the stored cursor is zero.
- External reward updates read `Deposit.GetExternalRewardLastCollectTime(...)` directly, which bypassed the wrapped fallback behavior.

As a result, the external path did not enforce the same lower bound in two edge cases:

- when the external reward cursor did not exist yet
- when an external reward cursor entry existed but stored `0`

In those cases, the external path could skip or weaken the monotonic validation that the internal path already performed. The normal production flow is not expected to create those states intentionally, but the implementation still encoded two different strengths of the same invariant. That made the reward cursor logic harder to reason about and easier to regress later.

## Changes

- Add a shared monotonic collect-time assertion on `DepositResolver`.
- Route both internal and external reward cursor updates through the shared assertion.
- Validate external reward cursor updates through `ExternalRewardLastCollectTime(incentiveID)`, so missing or zero cursor state falls back to `StakeTime()` like the internal path.
- Remove the redundant external reward cursor map guard because `Deposit.SetExternalRewardLastCollectTime` already lazy-initializes the map.
- Align the public getter read path so the internal last collect timestamp uses the same wrapped resolver behavior as the external getter path.
- Consolidate the internal `CollectReward` update flow so reward accounting, cursor updates, transfers, and event emission live under one positive `hasInternalReward` branch.
- Narrow the scope of internal reward accounting variables and return explicit zero values from the no-internal-reward path.

## Release Notes

* **Bug Fixes**
  * Improved internal reward collection handling with fallback timestamp logic when the last-collect timestamp is unavailable, now defaults to the deposit's stake time.

* **Tests**
  * Added test coverage for internal reward collection fallback behavior.

### DIFF
--- a/contract/r/gnoswap/staker/v1/getter.gno
+++ b/contract/r/gnoswap/staker/v1/getter.gno
@@ -284,9 +284,7 @@ func (s *stakerV1) GetDepositLiquidityAsString(lpTokenId uint64) string {
 
 // GetDepositInternalRewardLastCollectTimestamp returns the last collect timestamp of a deposit.
 func (s *stakerV1) GetDepositInternalRewardLastCollectTimestamp(lpTokenId uint64) int64 {
-	deposit := s.getDeposit(lpTokenId)
-
-	return deposit.InternalRewardLastCollectTime()
+	return s.getDepositResolver(lpTokenId).InternalRewardLastCollectTime()
 }
 
 // GetDepositCollectedInternalReward returns the collected internal reward amount.
--- a/contract/r/gnoswap/staker/v1/getter_test.gno
+++ b/contract/r/gnoswap/staker/v1/getter_test.gno
@@ -320,6 +320,14 @@ func TestGetterDepositBasicProperties(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


jinoosss:
At this time, we need to separate the functions to optimize gas costs.
Due to a difference in direction, I will close this PR.

PR #1264, opened 2026-04-21
test(scenario): fix time-unit and reward-accounting assertions

notJoon:
## Description

This PR fixes scenario tests that were using block height skips as if they were wall-clock seconds, and clarifies the staking warmup penalty accounting that was previously mixing separate reward flows into one number.

The changes make the scenario intent explicit by deriving skip heights from `5 seconds / block` and updating the golden outputs to match.

## Why This Change Is Needed

`testing.SkipHeights(N)` skips `N` blocks. In the GnoVM test environment, one block advances time by 5 seconds. Some scenario tests treated `N` as seconds, which made the scenarios run much later than their labels claimed.

Two concrete cases were fixed:

- The staking warmup penalty scenario intended to collect during the warmup penalty period, but `SkipHeights(3600)` advanced time by 18,000 seconds instead of 3,600 seconds.
- The launchpad scenarios used `SkipHeights(7776000)` for a 180-day project, but `7,776,000` blocks at 5 seconds per block is 450 days. The correct block count for 180 days is:

```text
180 days * 24 hours/day * 60 minutes/hour * 60 seconds/minute / 5 seconds/block
= 3,110,400 blocks
```

Leaving these values as magic numbers made the scenarios easy to misread and could hide regressions around time-sensitive reward logic.

## What Changed

- Updated the warmup penalty scenario to compute the skip height from seconds:
  - `blockTimeSeconds = 5`
  - `3600 seconds / 5 seconds per block = 720 blocks`
- Added a guard that fails the warmup test if collection happens after the cumulative warmup penalty period.
- Split staking reward accounting into separate values:
  - user reward net of warmup reduction and unstaking fee
  - actual warmup penalty returned by `CollectReward`
  - community pool emission distribution
  - total community pool balance increase
- Updated launchpad project-end scenarios to compute 180-day skips as `3,110,400` blocks.
- Preserved the existing grace block behavior in the no-deposit launchpad refund scenario.
- Updated scenario golden outputs to match the corrected behavior.

## Does the New Values Are Correct

The corrected launchpad value, `3,110,400`, is the exact number of 5-second blocks in 180 days. This matches the launchpad tier model, where the longest tier ends at 180 days.

The corrected staking warmup value, `720`, is the exact number of 5-second blocks in 3,600 seconds. The resulting collection time is `3,605` seconds after stake because the stake itself and the later collection happen on block boundaries. This is still inside the cumulative warmup penalty period of `12,600` seconds, so the test now validates the intended warmup path.

The launchpad reward and refund amounts did not change after replacing the 450-day overskip with the correct 180-day skip. That confirms the tests only needed to reach the fully ended project state; the previous extra 270 days were unnecessary and misleading.

The warmup penalty numbers changed because the test now measures the actual `CollectReward` warmup penalty separately from unrelated community pool emission distribution. The community pool balance delta is intentionally larger than the warmup penalty because it includes both:

- the warmup penalty transferred during reward collection
- the community pool's normal emission distribution from `MintAndDistributeGns`


junghoon-vans (approved):
Looks good to me overall. The changes make the scenario intent much clearer

jinoosss (approved):

PR #1265, opened 2026-04-21
refactor(router): cache precomputed sqrtPriceLimit values in init

aronpark1007:
## Description                  

  Precomputes TickMathGetSqrtRatioAtTick results for all fee tier and swap direction combinations in init() and caches them in a map, replacing per-swap tick math computation with a  
  simple map lookup.
                                                                                                                                                                                       
## Why This Change Is Needed

  TickMathGetSqrtRatioAtTick involves heavy u256 bit arithmetic and was called on every swap to compute the sqrt price limit. Since the inputs are determined by a fixed set of 4 fee  
  tiers and 2 swap directions, all 8 combinations can be precomputed once at initialization. This was confirmed to be the largest gas reduction among the optimizations explored in
  this series.  

## Test Plan

**cache precomputed sqrtPriceLimit values in init**
- `TestMustGetSqrtPriceLimit` (swap_inner_test.gno) — validates cached sqrtPriceLimit values across all 4 fee tiers × 2 directions, and panics on unknown fee

## Summary by CodeRabbit

* **Chores**
  * Improved swap performance by using shared, per-fee/direction price limits to reduce per-swap work while still honoring any non-zero limit supplied by the caller.
  * Added explicit failure for unsupported fee tiers so invalid configurations surface clearly.
  * Updated integration test expectations related to deployment gas measurements.
* **Tests**
  * Added unit tests validating cached limit lookups for supported fee tiers and verifying clear failure for unknown fees.


aronpark1007:
@jinoosss
Extracted repeated map lookup into `mustGetSqrtPriceLimit` and added test coverage for the extracted function.

junghoon-vans (approved):

jinoosss (approved):

PR #1267, opened 2026-04-22
refactor(pool): optimize cache block timestamp

aronpark1007:
## Summary                                                                                                                                                                           
                  
- Cache block timestamp once at swap entry instead of calling `time.Now()` on every tick cross during the swap loop

## Test Plan

No new tests required. Existing tests cover all changes:

**cache block timestamp during swap**
- `TestOracle_SwapScenarios` — covers oracle write timestamp path inside swap
- `TestOracle_TWAPConsistencyCases` — covers TWAP timestamp consistency including tick crossing
- `TestSwap_PriceLimitEdgeCase_ZeroAmount` — reflects newSwapCache signature change (already updated to pass 0)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timestamp consistency across swap flows to ensure deterministic behavior during execution and simulations
* **Refactor**
  * Reduced redundant system time calls in swap processing for cleaner, more consistent runtime behavior
* **Tests**
  * Adjusted swap tests to align with the timestamp handling changes
* **Chores**
  * Updated integration deployment expectations for increased gas/fee values

### DIFF
--- a/contract/r/gnoswap/pool/store.gno
+++ b/contract/r/gnoswap/pool/store.gno
@@ -175,11 +175,10 @@ func (s *poolStore) HasSwapStartHook() bool {
 }
 
 // GetSwapStartHook retrieves the swap start hook function.
-// Returns nil if no hook is set.
 func (s *poolStore) GetSwapStartHook() func(cur realm, poolPath string, timestamp int64) {
 	result, err := s.kvStore.Get(StoreKeySwapStartHook.String())
 	if err != nil {
-		return nil
+		panic(err)
 	}
 
 	swapStartHook, ok := result.(func(cur realm, poolPath string, timestamp int64))
@@ -201,11 +200,10 @@ func (s *poolStore) HasSwapEndHook() bool {
 }
 
 // GetSwapEndHook retrieves the swap end hook function.
-// Returns nil if no hook is set.
 func (s *poolStore) GetSwapEndHook() func(cur realm, poolPath string) error {
 	result, err := s.kvStore.Get(StoreKeySwapEndHook.String())
 	if err != nil {
-		return nil
+		panic(err)
 	}
 
 	swapEndHook, ok := result.(func(cur realm, poolPath string) error)
--- a/contract/r/gnoswap/pool/store.gno
+++ b/contract/r/gnoswap/pool/store.gno
@@ -175,10 +175,11 @@ func (s *poolStore) HasSwapStartHook() bool {
 }
 
 // GetSwapStartHook retrieves the swap start hook function.
+// Returns nil if no hook is set.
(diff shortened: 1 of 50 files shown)


contract/r/gnoswap/pool/store.gno
```diff
@@ -200,10 +201,11 @@ func (s *poolStore) HasSwapEndHook() bool {
 }
 
 // GetSwapEndHook retrieves the swap end hook function.
+// Returns nil if no hook is set.
 func (s *poolStore) GetSwapEndHook() func(cur realm, poolPath string) error {
 	result, err := s.kvStore.Get(StoreKeySwapEndHook.String())
 	if err != nil {
-		panic(err)
+		return nil
 	}
```
jinoosss, line 207: Errors occurring in the store are ignored. 
This differs from how other query functions in the store are handled.

jinoosss (commented):
The changes to swapCache look fine.
However, the naming convention for functions that query hooks is changing, which may differ from the convention used in existing functions.

contract/r/gnoswap/pool/store.gno
aronpark1007: Instead of changing the store's error handling, we can keep `GetSwapStartHook` / `GetSwapEndHook` panicking on error and simplify the call site by trusting `Has*` as the guard:

```go
if i.store.HasSwapStartHook() {
    i.store.GetSwapStartHook()(cross, pool.PoolPath(), blockTimestamp)
}
```

Since `Has*` already confirms the key exists, the nil check after `Get*` is unnecessary. This removes the redundant double-check without touching the store's error handling convention.

How about reverting the `store.gno` change and updating the call site this way instead?

aronpark1007: 14d95ad2

contract/r/gnoswap/pool/v1/swap.gno
```diff
@@ -169,15 +169,11 @@ func (i *poolV1) Swap(
...
+	blockTimestamp := time.Now().Unix()
 
 	// Call swap start hook if set
 	if i.store.HasSwapStartHook() {
-		swapStartHook := i.store.GetSwapStartHook()
-
-		if swapStartHook != nil {
-			currentTime := time.Now().Unix()
-			swapStartHook(cross, pool.PoolPath(), currentTime)
-		}
```
jinoosss, line 180: The function type of `hook` can be `nil`.
When actually using the function, it would be best to add a `nil` check for defensive programming.

aronpark1007: As discussed, rolled back all store-related changes, keeping only the block timestamp caching. 
See revert commit: 294d158f

jinoosss (approved):

junghoon-vans (approved):

PR #1269, opened 2026-04-22
fix(staker): carry forward pool tier dust

junghoon-vans:
## Summary
- carry forward tier-level internal reward dust instead of dropping it during pool-tier reward division
- rename the persisted staker store surface to `PoolTierLeftAmount` to match local contract naming conventions
- add focused staker reward tests proving left-amount accumulation and later distribution
- add a staker scenario filetest for repeated short collections versus delayed collection in the same tier, plus fuzz mock support for the renamed store API
- refresh affected staker scenario golden outputs to reflect the carry-forward behavior across the full scenario package

## Validation
- make test PKG=gno.land/r/gnoswap/staker
- make test PKG=gno.land/r/gnoswap/staker/v1
- make test PKG=gno.land/r/gnoswap/test/fuzz
- make test PKG=gno.land/r/gnoswap/scenario/staker

## Summary by CodeRabbit

* **New Features**
  * Added per-tier left amount tracking for enhanced pool tier management.

* **Bug Fixes**
  * Improved reward caching with interval-aware calculation and proper carryover handling across halving boundaries.
  * Fixed remainder distribution during pool tier transitions.

### DIFF
--- a/contract/r/gnoswap/staker/v1/_mock_test.gno
+++ b/contract/r/gnoswap/staker/v1/_mock_test.gno
@@ -26,6 +26,7 @@ type MockStakerStore struct {
 	poolTierMemberships              *avl.Tree
 	poolTierRatio                    sr.TierRatio
 	poolTierCounts                   [sr.AllTierCount]uint64
+	poolTierLeftAmount               [sr.AllTierCount]int64
 	poolTierLastRewardCacheTimestamp int64
 	poolTierCurrentEmission          int64
 	poolTierGetEmission              func() int64
@@ -236,6 +237,20 @@ func (s *MockStakerStore) SetPoolTierCounts(counts [sr.AllTierCount]uint64) erro
 	return nil
 }
 
+// PoolTierLeftAmount
+func (s *MockStakerStore) HasPoolTierLeftAmountStoreKey() bool {
+	return s.poolTierLeftAmount != [sr.AllTierCount]int64{}
+}
+
+func (s *MockStakerStore) GetPoolTierLeftAmount() [sr.AllTierCount]int64 {
+	return s.poolTierLeftAmount
+}
+
+func (s *MockStakerStore) SetPoolTierLeftAmount(leftAmount [sr.AllTierCount]int64) error {
+	s.poolTierLeftAmount = leftAmount
+	return nil
+}
+
 // PoolTierLastRewardCacheTimestamp
 func (s *MockStakerStore) HasPoolTierLastRewardCacheTimestampStoreKey() bool {
 	return s.poolTierLastRewardCacheTimestamp != 0
@@ -354,6 +369,7 @@ func newMockStakerStore() *MockStakerStore {
 			Tier3: 0,
 		},
 		poolTierCounts:                   [sr.AllTierCount]uint64{},
+		poolTierLeftAmount:               [sr.AllTierCount]int64{},
 		poolTierLastRewardCacheTimestamp: 0,
 		poolTierCurrentEmission:          0,
 		poolTierGetEmission:              func() int64 { return 0 },
--- a/contract/r/gnoswap/staker/v1/_mock_test.gno
+++ b/contract/r/gnoswap/staker/v1/_mock_test.gno
@@ -27,6 +27,7 @@ type MockStakerStore struct {
 	poolTierRatio                    sr.TierRatio
 	poolTierCounts                   [sr.AllTierCount]uint64
 	poolTierLeftAmount               [sr.AllTierCount]int64
+	hasPoolTierLeftAmount            bool
 	poolTierLastRewardCacheTimestamp int64
 	poolTierCurrentEmission          int64
 	poolTierGetEmission              func() int64
@@ -239,7 +240,7 @@ func (s *MockStakerStore) SetPoolTierCounts(counts [sr.AllTierCount]uint64) erro
 
 // PoolTierLeftAmount
 func (s *MockStakerStore) HasPoolTierLeftAmountStoreKey() bool {
-	return s.poolTierLeftAmount != [sr.AllTierCount]int64{}
+	return s.hasPoolTierLeftAmount
 }
 
 func (s *MockStakerStore) GetPoolTierLeftAmount() [sr.AllTierCount]int64 {
@@ -248,6 +249,7 @@ func (s *MockStakerStore) GetPoolTierLeftAmount() [sr.AllTierCount]int64 {
 
 func (s *MockStakerStore) SetPoolTierLeftAmount(leftAmount [sr.AllTierCount]int64) error {
 	s.poolTierLeftAmount = leftAmount
+	s.hasPoolTierLeftAmount = true
 	return nil
 }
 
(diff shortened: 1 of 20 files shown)


contract/r/gnoswap/staker/v1/reward_calculation_pool_tier.gno
```diff
@@ -304,27 +308,41 @@ func (self *PoolTier) cacheRewardForPool(currentTimestamp int64, pools *Pools, p
 	// Determine halving boundaries since the pool's last cached reward timestamp.
 	halvingTimestamps, halvingEmissions := self.getHalvingBlocksInRange(lastTimestamp, currentTimestamp)
 	poolResolver := NewPoolResolver(pool)
+	localTierLeftAmount := self.tierLeftAmount
```
junghoon-vans, line 319: [review-agent] **[note] `cacheRewardForPool` discards its updated dust — per-pool path diverges from global state**

`localTierLeftAmount` is a snapshot of `self.tierLeftAmount` at the start of this call. `applyCacheToPool` updates `*tierLeftAmount` locally (via pointer), but after `cacheRewardForPool` returns, the updated carryover is never written back to `self.tierLeftAmount`.

This means every subsequent `cacheRewardForPool` call (from `CollectReward`) starts from the same stale dust level as the last full `cacheReward`. In contrast, `applyCacheToAllPools` does persist via `self.tierLeftAmount = nextTierLeftAmount`.

In practice each pool still receives the correct 1/N share (the formula divides by `tierCount × duration`), so there is no per-pool over-distribution. The scenario test bounds the aggregate divergence at ≤2 units. However, the behaviour is undocumented.

**Recommendation**: Add a comment here stating that `localTierLeftAmount` is intentionally not written back (optimization — full dust persistence only happens on `cacheReward`). If exact dust fidelity in the per-pool path becomes a requirement, the updated array must be persisted after the call.

contract/r/gnoswap/staker/v1/reward_calculation_pool_tier.gno
```diff
@@ -420,6 +448,52 @@ func calculatePoolReward(emission int64, tierRatio int64, tierCount int64) int64
 	return tierReward / tierCount
 }
 
+func calculatePoolRewardForInterval(emission int64, tierRatio int64, tierCount int64, intervalDuration int64, carryover int64) (int64, int64) {
+	if emission < 0 || tierRatio < 0 || tierCount < 0 || intervalDuration < 0 || carryover < 0 {
```
junghoon-vans, line 460: [review-agent] **[note] `carryover < 0` guard is unreachable under normal operation**

`carryover` enters this function as the output of `safeSubInt64(tierRewardForInterval, distributedReward)`. Since `distributedReward = poolReward × tierCount × intervalDuration` and `poolReward` is integer-truncated from `tierRewardForInterval / (tierCount × intervalDuration)`, we always have `distributedReward ≤ tierRewardForInterval`, guaranteeing `carryover ≥ 0`.

The guard is defensive and does no harm, but may mislead a future reader into treating negative carryover as a realistic scenario.

**Recommendation**: Add a comment clarifying that `carryover` is non-negative by construction (integer division remainder), or remove the guard if it is considered noise.

contract/r/gnoswap/staker/v1/reward_calculation_pool_tier.gno
```diff
@@ -420,6 +448,52 @@ func calculatePoolReward(emission int64, tierRatio int64, tierCount int64) int64
...
+		panic(errCalculationError)
+	}
+
+	if emission == 0 || tierRatio == 0 || tierCount == 0 || intervalDuration == 0 {
+		return 0, carryover
+	}
+
+	tierReward := safeMulDivInt64(emission, tierRatio, 100)
+	tierRewardForInterval := safeAddInt64(safeMulInt64(tierReward, intervalDuration), carryover)
```
junghoon-vans: [review-agent] **[note] `safeMulInt64(tierReward, intervalDuration)` can panic if emission × interval grows too large**

`tierReward` can be up to the full emission (100% ratio, single-pool tier). `intervalDuration` is the number of seconds elapsed since the last full `cacheReward` call. If the contract goes a long time without a tier operation, stake/unstake, or emission hook (all of which trigger `cacheReward`), `intervalDuration` can become very large.

`safeMulInt64` panics on overflow, so the failure mode is a contract-level panic when the next `CollectReward` or tier operation is attempted — effectively halting reward operations for all users until the overflow condition is resolved.

With typical GNS emission values (≤10^12/sec) and daily intervals, the product is ~8.6×10^16 — well within `int64` range. Risk only materialises at high emission values combined with week+ intervals.

**Recommendation**: Document the expected maximum safe interval duration given the protocol's peak emission rate. Consider periodic caching from the emission hook to prevent the gap from growing unbounded.

contract/r/gnoswap/staker/v1/_mock_test.gno
```diff
@@ -236,6 +237,20 @@ func (s *MockStakerStore) SetPoolTierCounts(counts [sr.AllTierCount]uint64) erro
 	return nil
 }
 
+// PoolTierLeftAmount
+func (s *MockStakerStore) HasPoolTierLeftAmountStoreKey() bool {
```
junghoon-vans, line 242: [review-agent] **[note] `HasPoolTierLeftAmountStoreKey` uses value-equality instead of key-presence semantics**

```go
func (s *MockStakerStore) HasPoolTierLeftAmountStoreKey() bool {
    return s.poolTierLeftAmount != [sr.AllTierCount]int64{}
}
```

The real `stakerStore` returns `true` once the key has been written — even if the stored value is all-zeros. This mock returns `false` for the zero-valued initial state, causing `initStoreData`'s guard to re-execute `SetPoolTierLeftAmount` with the same zero value. The write is idempotent so no data corruption occurs, but the mock diverges from real-store semantics.

This is the same pattern used by the other `Has*` mock methods (`HasPoolTierCountsStoreKey`, etc.), so it's a systemic mock design choice rather than an isolated mistake.

**Recommendation**: Store an explicit `initialized bool` flag per key (or a `map[string]bool` of initialized keys) in the mock to match real-store key-presence semantics. At minimum, add a comment on each `Has` method explaining the zero-equals-absent assumption so future test authors are aware.

junghoon-vans (commented):
[review-agent] PR #1269 review — fix(staker): carry forward pool tier dust

## Overall

The core fix is correct and well-motivated. Two pre-existing bugs (`lastTimestamp` not advanced in the halving loop; `currentEmission` initialized to `0` instead of `self.currentEmission` in `cacheRewardForPool`) are fixed as part of this work. The new `calculatePoolRewardForInterval` math is sound and the scenario filetest confirms the fix holds within a ≤2-unit tolerance between frequent and delayed collectors.

Findings are posted as inline comments where possible. One cross-cutting note is included here:

---

**[note] Accumulated dust is preserved when a tier becomes empty and survives pool re-entry**

`computeTierRewardsForInterval` skips a tier when `tierCount == 0` (continues without touching `tierLeftAmount[tier]`). This means if all pools leave a tier, the orphaned dust persists in `tierLeftAmount[tier]`. When a new pool joins that tier later, it inherits the old dust as a one-time bonus on the first reward interval.

The magnitude is bounded (dust is always less than `tierCount × intervalDuration` emission units at the time of last computation), so this cannot cause a large accounting error. But it is semantically surprising: a freshly-joined pool receives a carry-over from a prior, unrelated pool set.

**Recommendation**: Reset `tierLeftAmount[tier]` to zero inside `changeTier` when `self.counts[currentTier]` reaches 0 after decrement.

PR #1275, opened 2026-04-29
refactor(position): support explicit mint recipients

junghoon-vans:
## Summary
- Simplify `position.Mint` so the caller is derived from `runtime.PreviousRealm()` while `mintTo` remains an explicit recipient parameter.
- Remove the `MintAndStake` surfaces and migrate callers to the explicit `Mint -> Approve -> StakeToken` flow.
- Add coverage for third-party recipients, contract-owned positions, and a full vault-owned lifecycle including fee collection, liquidity changes, repositioning, staking, reward collection, and unstaking.

## Test Plan
- `gno test ./gno.land/r/gnoswap/position/v1`
- `gno test ./gno.land/r/gnoswap/scenario/position`


jinoosss (approved):

jinoosss (commented):
As the position's Mint function has been updated to allow calls from contracts,
the following considerations must be addressed:

If the position contract's Mint function is called inside the pool's Swap callback,
the pool's balance will increase.

The pool's Swap checks the contract's token balance after the callback is invoked,
at which point the increased balance from the position Mint may substitute the expected amount.

To prevent this case, a global swap lock should be introduced in the pool contract
to block any state changes to the pool during an ongoing Swap.

jinoosss (changes requested):

jinoosss (approved):

dongwon8247 (commented):
check uniswap v4 and ekubo for the singleton pool design @junghoon-vans

jinoosss (approved):

PR #1282, opened 2026-05-12
feat(pool, staker): remove ticks logging and enhance tick cross event

jinoosss:
## Summary

Splits the bulky `ticks` payload that was attached to every `Swap` and `SwapTickCross` event into per-cross `PoolTickCross` / `StakerTickCross` events so the backend can index `pool_tick` rows without hitting per-event size limits.

## Background

The previous `Swap` event packed a `ticks` field built from `ticksToString(pool, startTick, slot0Tick)`, which serialized every initialized tick within the swap's price range as a JSON array. Each entry carried 8 uint128/uint256 string fields (`lg`, `ln`, `fg0`, `fg1`, `tco`, `spl`, `so`, `i`) and reached ~500–600 bytes per tick.

For wide-range swaps that crossed many initialized ticks, the resulting `Swap` event payload grew large enough to risk exceeding the backend's per-event ingestion limit. The staker side suffered the same problem via `SwapTickCross.crossedTicks`.

Because the backend computes position rewards itself, it requires accurate `fee_growth_outside0/1_x128` values per cross. These cannot be reliably reconstructed from the swap-final `feeGrowthGlobal` when multiple ticks are crossed in one swap (each cross sees a different intermediate global), so the data must be emitted at the exact cross moment.

### Pool (`contract/r/gnoswap/pool/v1/`)

- **`tick.gno`**: emit `PoolTickCross` inside `tickCross()` right after the tick's outside accumulators are updated. The payload reflects the post-cross state, which is the snapshot the backend needs for `pool_tick`.
  - `poolPath`
  - `tick` — JSON object via `NewTickEventInfo(tick, thisTick).ToString()` carrying `t`, `lg`, `ln`, `fg0`, `fg1`, `tco`, `spl`, `so`, `i`
- **`swap.gno`**: remove the `ticks` field and the `startTick` variable that only fed `ticksToString`.
- **`utils.gno` / `utils_test.gno`**: drop `ticksToString` and its tests — no longer referenced.

### Staker (`contract/r/gnoswap/staker/v1/reward_calculation_tick.gno`)

- Emit one `StakerTickCross` per crossed tick inside `processBatchedTickCrosses`, mirroring the pool-side split.
  - `poolPath`, `tick` (JSON via `NewTickCrossEventInfo(...).ToString()`)
- Rename the per-swap summary event `SwapTickCross` → `BatchStakerTickCross` and drop the embedded `crossedTicks` array (now redundant with per-cross events).
- `prevAddr` / `prevRealm` are intentionally omitted from `PoolTickCross` and `StakerTickCross` to minimize per-event size; the parent `Swap` / `BatchStakerTickCross` events already carry that context within the same transaction.

### `PoolTickCross` (pool/v1)

| Key | Type | Notes |
|-----|------|-------|
| `poolPath` | string | |
| `tick` | JSON | `{t, lg, ln, fg0, fg1, tco, spl, so, i}` — post-cross snapshot |

### `StakerTickCross` (staker/v1)

| Key | Type | Notes |
|-----|------|-------|
| `poolPath` | string | |
| `tick` | JSON | staker-side tick cross info (id, stakedLiquidityGross, stakedLiquidityDelta, currentOutsideAccumulation) |

### `BatchStakerTickCross` (staker/v1) — renamed from `SwapTickCross`

| Key | Type | Notes |
|-----|------|-------|
| `prevAddr`, `prevRealm` | string | |
| `poolPath` | string | |
| `blockTimestamp` | int64 | |
| `stakedLiquidity` | uint256 string | |
| `globalRewardRatioAccX128` | uint256 string | |


junghoon-vans (approved):
Everything looks fine!

notJoon (approved):

PR #1291, opened 2026-05-21
fix(staker): scale external incentive rewardPerSecond by 2^128 to reduce dust loss

jinoosss:
## Summary

External incentive `rewardPerSecond` previously suffered substantial precision loss from int64 integer division: `rewardAmount / duration` discarded the remainder, locking the dust in the contract. For small `rewardAmount` relative to `duration` the loss could approach 100% of the deposit. This PR scales `rewardPerSecond` by 2^128 so that sub-integer rates survive across the lifetime of an incentive.

- Store reward per second as a Q128 fixed-point `*u256.Uint` (`rewardPerSecondX128 = (rewardAmount << 128) / duration`)
- External incentive reward calculation now multiplies by the Q128 rate and floors back with `>> 128` per `MulDiv` (one floor per collect call, ≤ 1 wei loss each)
- Internal reward path is **unchanged** — still uses the int64 `rewardPerSecond` directly, no Q128 conversion
- `endExternalIncentive` refund formula updated to derive `distributable` from the Q128 rate
- Quantitative impact on the dust-refund scenario (`rewardAmount=10_000_000_007`, `duration=7_776_000 s`):
  | Quantity | Before (int64 rps) | After (Q128) |
  |---|---|---|
  | Refund (residual dust) | 70,437 wei | **6,431 wei** (≈11× smaller) |
  | Distributed to stakers | 7,974,942,592 | 7,974,993,637 (+51,045) |
  | Staker contract leftover | 3 wei | **2 wei** |

## Breaking changes

- `IStakerGetter.GetIncentiveRewardPerSecond` return type changed from `int64` to `*u256.Uint`. The returned value is now Q128-scaled — divide by `2^128` to recover the legacy integer rate. External consumers (indexers / dApps) must update.

## Implementation notes

- New helper `rewardPerWarmupX128(startTime, endTime, rps_x128)` in `reward_calculation_pool.gno` is used only by `updateExternalReward`. The original `rewardPerWarmup(startTime, endTime, rps int64)` is preserved as-is for the internal reward path to avoid expanding the overflow surface on every swap-hook reward calculation.
- `ExternalIncentive` struct field renamed `rewardPerSecond int64` → `rewardPerSecondX128 *u256.Uint`. A legacy `RewardPerSecond() int64` getter is kept on the struct, returning the floor of the Q128 value for observability only — internal callers must use `RewardPerSecondX128()`.
- `NewExternalIncentive` now computes the Q128 rate via `u256.MulDiv(rewardAmount, 2^128, duration)`; callers no longer need to call `SetRewardPerSecond` after construction (removed in tests and the canonical-env helper).

## Summary by CodeRabbit

* **Refactor**
  * Incentive reward math now uses higher-precision 128-bit fixed-point arithmetic for more accurate per-second rates and reward accruals
  * External and internal reward paths aligned to preserve precision across creation, distribution, and warmup segments
* **Tests**
  * Added tests validating precise external-incentive refund and reward calculations using the new fixed-point representation

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnoswap-labs/gnoswap/pull/1291?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/contract/r/gnoswap/staker/getter.gno
+++ b/contract/r/gnoswap/staker/getter.gno
@@ -181,8 +181,10 @@ func GetIncentiveRewardAmountAsString(poolPath string, incentiveId string) strin
 	return getImplementation().GetIncentiveRewardAmountAsString(poolPath, incentiveId)
 }
 
-// GetIncentiveRewardPerSecond returns the reward rate per second of an incentive.
-func GetIncentiveRewardPerSecond(poolPath string, incentiveId string) int64 {
+// GetIncentiveRewardPerSecond returns the reward rate per second of an incentive,
+// expressed as a Q128 fixed-point number (i.e. actual rate = value / 2^128).
+// Callers needing an integer rate can right-shift the result by 128.
+func GetIncentiveRewardPerSecond(poolPath string, incentiveId string) *u256.Uint {
 	return getImplementation().GetIncentiveRewardPerSecond(poolPath, incentiveId)
 }
 
--- a/contract/r/gnoswap/staker/getter.gno
+++ b/contract/r/gnoswap/staker/getter.gno
@@ -181,11 +181,12 @@ func GetIncentiveRewardAmountAsString(poolPath string, incentiveId string) strin
 	return getImplementation().GetIncentiveRewardAmountAsString(poolPath, incentiveId)
 }
 
-// GetIncentiveRewardPerSecond returns the reward rate per second of an incentive,
-// expressed as a Q128 fixed-point number (i.e. actual rate = value / 2^128).
-// Callers needing an integer rate can right-shift the result by 128.
-func GetIncentiveRewardPerSecond(poolPath string, incentiveId string) *u256.Uint {
-	return getImplementation().GetIncentiveRewardPerSecond(poolPath, incentiveId)
+// GetIncentiveRewardPerSecondX128 returns the reward rate per second of an
+// incentive, expressed as a Q128 fixed-point number (i.e. actual rate =
+// value / 2^128). Callers needing an integer rate can right-shift the result
+// by 128.
+func GetIncentiveRewardPerSecondX128(poolPath string, incentiveId string) *u256.Uint {
+	return getImplementation().GetIncentiveRewardPerSecondX128(poolPath, incentiveId).Clone()
 }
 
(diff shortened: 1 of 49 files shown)


contract/r/gnoswap/staker/getter.gno
```diff
@@ -181,8 +181,10 @@ func GetIncentiveRewardAmountAsString(poolPath string, incentiveId string) strin
 	return getImplementation().GetIncentiveRewardAmountAsString(poolPath, incentiveId)
 }
 
-// GetIncentiveRewardPerSecond returns the reward rate per second of an incentive.
-func GetIncentiveRewardPerSecond(poolPath string, incentiveId string) int64 {
+// GetIncentiveRewardPerSecond returns the reward rate per second of an incentive,
+// expressed as a Q128 fixed-point number (i.e. actual rate = value / 2^128).
+// Callers needing an integer rate can right-shift the result by 128.
+func GetIncentiveRewardPerSecond(poolPath string, incentiveId string) *u256.Uint {
 	return getImplementation().GetIncentiveRewardPerSecond(poolPath, incentiveId)
```
junghoon-vans: Please add `Clone` function calling

dongwon8247 (approved):

contract/r/gnoswap/staker/v1/external_incentive.gno
```diff
@@ -256,10 +256,14 @@ func (s *stakerV1) endExternalIncentive(resolver *PoolResolver, incentiveResolve
...
-	distributableX256 := u256.Zero().Mul(
+	distributableX256, overflow := u256.Zero().MulOverflow(
 		incentiveResolver.RewardPerSecondX128(),
 		u256.NewUintFromInt64(duration),
 	)
+	if overflow {
+		panic(errOverflow)
+	}
+
 	distributable := safeConvertToInt64(u256.Zero().Rsh(distributableX256, 128))
```
junghoon-vans: How about use u256.MulDiv

jinoosss: d9aa5bf1cc8f8898ec2b5ed2c6ff9251b7d89f11

junghoon-vans (approved):

PR #1293, opened 2026-05-26
refactor(interrealm): migrate inter-realm spec v2

jinoosss:
## Summary by CodeRabbit

* **Documentation**
  * Corrected terminology in deployment documentation for accuracy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnoswap-labs/gnoswap/pull/1293?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


contract/p/gnoswap/store/doc.gno
```diff
@@ -1,164 +1,175 @@
...
-//   - CastToInt64: Casts to int64 with error handling
-//   - CastToUint64: Casts to uint64 with error handling
-//   - CastToBool: Casts to bool with error handling
-//   - CastToString: Casts to string with error handling
-//   - CastToAddress: Casts to address with error handling
-//
-// These utilities are used internally by typed getters but can also be used directly.
+//   - `castToInt64`, `castToUint64`, `castToBool`, `castToString`,
+//     `castToAddress`, `castToBPTree` - Safe runtime casts used by the typed
+//     getters and reusable for custom flows.
```
jeronimoalbi: Mention in case you find it relevant, the "Type Casting Utilities" section documents private functions. They might end up as part of dev/user docs though package docs generation, for example in some Gnoweb's view I think.

contract/p/gnoswap/version_manager/doc.gno
```diff
@@ -119,7 +119,7 @@
 //	func init() {
 //	    // Register this version during package initialization
 //	    manager := protocol_fee.GetManager()
-//	    manager.RegisterInitializer(func(store any) any {
+//	    manager.RegisterInitializer(func(_ int, rlm realm, store any) any {
```
jeronimoalbi: It seems docs are missing the cross to be inline with the code:

```go
func init(cur realm) {
    //...
    manager.RegisterInitializer(cross(cur), func(_ int, rlm realm, store any) any
    // ...
}
```

contract/p/gnoswap/version_manager/README.md
```diff
@@ -71,9 +91,10 @@ type protocolFeeV1 struct {
 }
 
 func init() {
-    // Register this version during package initialization
-    manager := protocol_fee.GetManager()
-    manager.RegisterInitializer(func(store any) any {
+    // Register this version during package initialization.
+    // `cross` invokes the domain's crossing entry point, which threads the
+    // live realm token into the version manager.
+    protocol_fee.RegisterInitializer(cross, func(_ int, rlm realm, store any) any {
```
jeronimoalbi: README.md documentation is not yet inline with the code, is missing `init(cur realm)` and `cross(cur)` for example

contract/p/gnoswap/version_manager/version_manager.gno
```diff
@@ -234,13 +258,13 @@ func NewVersionManager(
 // makeInitializerSafe creates a safe copy of an initializer function.
 // Used to verify stored initializer types during inspection.
 func makeInitializerSafe(data any) any {
```
jeronimoalbi: Nothing important. I was curious to figure out the reason for the wrapping and noticed this function is only used in the test file, so it seems it could potentially be moved there 🤔

contract/r/gnoswap/protocol_fee/_helper_test.gno
```diff
@@ -17,20 +17,20 @@ var adminAddr = access.MustGetAddress(prbac.ROLE_ADMIN.String())
 func resetTestState(t *testing.T) {
 	testing.SetRealm(testing.NewCodeRealm("gno.land/r/gnoswap/protocol_fee"))
 
-	kvStore = store.NewKVStore(runtime.CurrentRealm().Address())
+	kvStore = store.NewKVStore(unsafe.CurrentRealm().Address())
```
jeronimoalbi: Another way that would work, so `unsafe` import is removed:

```go
func resetTestState(t *testing.T, rlm realm) {
	// ...
	kvStore = store.NewKVStore(rlm.Address())
	// ...
}
```

contract/r/gnoswap/protocol_fee/state.gno
```diff
@@ -11,16 +10,22 @@ import (
 )
 
 var (
-	currentAddress = runtime.CurrentRealm().Address()
-	domainPath     = runtime.CurrentRealm().PkgPath()
+	// Package-init reads use chain/runtime/unsafe.CurrentRealm because there
+	// is no `cur realm` in scope at package-initialization time. The values
+	// captured here describe the protocol_fee realm itself, not any caller.
```
jeronimoalbi: Seems that it could be removed:

```suggestion
```

jeronimoalbi (commented):
Is looking good 👍 

Leaving minor comments in case some of them make sense to be addressed.

contract/r/gnoswap/protocol_fee/_helper_test.gno
junghoon-vans: That's an important point. If we do that, we won't need to use `unsafe` in the tests either.
I have made the changes as you requested.

contract/p/gnoswap/store/kv_store.gno
```diff
@@ -187,24 +194,28 @@ func (k *kvStore) IsWriteAuthorized(caller address) bool {
 // GetAuthorizedCallers returns all authorized callers and their permissions
 func (k *kvStore) GetAuthorizedCallers() (map[address]Permission, error) {
 	if k.authorizedCallers == nil {
-		return make(map[address]Permission), ErrAuthorizedCallerNotFound
+		return make(map[address]Permission), errors.New(ErrAuthorizedCallerNotFound)
 	}
 
 	return k.authorizedCallers, nil
```
moul: **ACL map returned by reference.** Any holder of the `KVStore` interface can mutate the internal `authorizedCallers` map directly, bypassing the `rlm.IsCurrent()` and domain-address checks enforced by `AddAuthorizedCaller` / `UpdateAuthorizedCaller` / `RemoveAuthorizedCaller`. Suggested fix: return a defensive copy.
```go
copy := make(map[address]Permission, len(k.authorizedCallers))
for k, v := range k.authorizedCallers {
    copy[k] = v
}
return copy, nil
```

contract/p/gnoswap/store/kv_store.gno
```diff
@@ -127,38 +127,45 @@ func (k *kvStore) GetBPTree(key string) (*bptree.BPTree, error) {
...
+// bypass the ACL so deploy and test flows can populate state without prior
+// registration.
+func (k *kvStore) Set(_ int, rlm realm, key string, value any) error {
+	if !rlm.IsCurrent() {
+		return errors.New(ErrSpoofedRealm)
+	}
 
-	if currentRealm.IsCode() && !k.IsWriteAuthorized(currentRealm.Address()) {
-		return ErrWritePermissionDenied
+	if rlm.IsCode() && !k.IsWriteAuthorized(rlm.Address()) {
```
moul, line 140: **EOA bypass in `Set` is asymmetric with `Delete`.** When `!rlm.IsCode()` (user/EOA caller), the ACL check is skipped entirely. `Delete` at line 163 does not have this bypass, with the comment explaining "Delete is destructive" — but `Set` is equally destructive (silent overwrite of any key). If a domain realm has a crossing function reachable by a user, that user can write arbitrary keys. Either remove the bypass and require explicit ACL registration for all callers, or at minimum add a prominent warning in the `NewKVStore` doc comment that this bypass exists and must be accounted for by every consuming realm.

moul (commented):
Review of the Interrealm v2 migration. The migration direction is solid — realm threading is consistent, `rlm.IsCurrent()` spoof guards are in the right places, and the upgrade scenario tests give real confidence against readonly-taint regressions. A few mutable-reference escapes remain from the earlier review pass, plus two new findings. Items below are the ones verified against the current HEAD.

contract/p/gnoswap/store/kv_store.gno
moul: test

moul (commented):
test

contract/p/gnoswap/store/kv_store.gno
moul, line 140: test

moul (commented):
test

moul (commented):
test

contract/r/gnoswap/rbac/consts.gno
```diff
@@ -1,38 +1,54 @@
...
 	// DEV_OPS is the initial DevOps address for operational tasks.
 	DEV_OPS address = "g1mjvd83nnjee3z2g7683er55me9f09688pd4mj9"
-)
 
-// Derived addresses for GnoSwap protocol packages.
-var (
 	// EMISSION_ADDR is the derived address for the emission package.
-	EMISSION_ADDR address = chain.PackageAddress("gno.land/r/gnoswap/emission")
+	// EMISSION_ADDR is chain.PackageAddress("gno.land/r/gnoswap/emission")
+	EMISSION_ADDR address = "g177jkk4xx79uledh9xedqkq4ht4ef9t6amxwjeq"
```
moul: test

contract/r/gnoswap/staker/store.gno
```diff
@@ -176,8 +195,45 @@ func (s *stakerStore) GetAllowedTokens() []string {
 	return tokens
```
moul: test

moul (commented):
test

contract/r/gnoswap/staker/store.gno
```diff
@@ -478,8 +620,12 @@ func (s *stakerStore) GetWarmupTemplate() []Warmup {
 	return warmups
```
moul: test

moul (commented):
test

contract/r/gnoswap/pool/pool.gno
```diff
@@ -89,11 +89,11 @@ func (p *Pool) SetSlot0(slot0 Slot0) {
 }
 
 func (p *Pool) SetFeeGrowthGlobal0X128(feeGrowthGlobal0X128 *u256.Uint) {
-	p.feeGrowthGlobal0X128 = feeGrowthGlobal0X128
+	p.feeGrowthGlobal0X128 = u256.Zero().Set(feeGrowthGlobal0X128)
```
moul, line 92: test

moul (commented):
test

moul (commented):
test

contract/r/gnoswap/pool/store.gno
```diff
@@ -165,8 +182,12 @@ func (s *poolStore) GetPendingProtocolFees() map[string]int64 {
 	return pendingProtocolFees
```
moul: test

contract/p/gnoswap/store/kv_store.gno
moul: **ACL map returned by reference.**

Any holder of the `KVStore` interface can mutate the internal `authorizedCallers` map directly, bypassing the `rlm.IsCurrent()` and domain-address checks on `Add/Update/RemoveAuthorizedCaller`.

```go
copy := make(map[address]Permission, len(k.authorizedCallers))
for addr, perm := range k.authorizedCallers {
    copy[addr] = perm
}
return copy, nil
```

contract/p/gnoswap/store/kv_store.gno
moul, line 140: **`Set` EOA bypass is asymmetric with `Delete`.**

When `!rlm.IsCode()` (user/EOA caller), the ACL check is skipped entirely. `Delete` (line 163) does not have this bypass — the comment explains "Delete is destructive," but `Set` silently overwrites any existing key too.

If a domain realm exposes a crossing function reachable by a user that eventually calls `Set`, that user can write arbitrary keys without being in the ACL. Either remove the bypass and require explicit registration for all callers, or add a clear warning in `NewKVStore`'s doc comment so consuming realms know to guard against this.

contract/r/gnoswap/rbac/consts.gno
moul: **Hardcoded derived package addresses.**

All 10 role addresses are now string literals with only a comment referencing the `chain.PackageAddress(...)` call they came from. A single transcription error silently breaks role-based ACL with no compile-time or test-time signal.

Options:
1. Restore `chain.PackageAddress("gno.land/r/gnoswap/emission")` etc. (computed at deploy time, always consistent with the package path).
2. If the motivation is to avoid the `chain` import, add an integration test asserting each constant matches the computed value:
```go
if EMISSION_ADDR != chain.PackageAddress("gno.land/r/gnoswap/emission") {
    panic("EMISSION_ADDR mismatch")
}
```

contract/r/gnoswap/staker/store.gno
moul: **`GetAllowedTokens` returns the stored slice by reference.**

`AddAllowedToken` / `RemoveAllowedToken` happen to be safe because `append` may create a new backing array. But a direct caller who assigns into the returned slice (`tokens[0] = "..."`) mutates the realm-owned value and bypasses `SetAllowedTokens`.

```go
return append([]string(nil), tokens...)
```

contract/r/gnoswap/staker/store.gno
moul: **`GetWarmupTemplate` returns the stored `[]Warmup` by reference.**

Callers get the live slice. Since `Warmup` contains only value types, element-level mutation won't alias, but replacing or appending elements via the returned slice bypasses `SetWarmupTemplate`.

```go
return append([]Warmup(nil), warmups...)
```

contract/r/gnoswap/pool/pool.gno
moul, line 92: Good — you added `.Clone()` here and on `SetLiquidity` (line 112). The same fix is still needed for `Slot0.SetSqrtPriceX96` (line 283) and `Slot0.SqrtPriceX96()` (line 278): the setter stores the caller's pointer directly (no clone), and the getter returns the live internal pointer. `NewSlot0` (line 295) already clones on construction — make the setter consistent:

```go
func (s *Slot0) SetSqrtPriceX96(sqrtPriceX96 *u256.Uint) {
    s.sqrtPriceX96 = sqrtPriceX96.Clone()
}
```

contract/r/gnoswap/pool/store.gno
```diff
@@ -91,12 +96,16 @@ func (s *poolStore) GetFeeAmountTickSpacing() map[uint32]int32 {
 }
 
 // SetFeeAmountTickSpacing stores the mapping between fee amounts and tick spacing.
-func (s *poolStore) SetFeeAmountTickSpacing(feeAmountTickSpacing map[uint32]int32) error {
+func (s *poolStore) SetFeeAmountTickSpacing(_ int, rlm realm, feeAmountTickSpacing map[uint32]int32) error {
```
moul, line 99: (Anchored on `SetFeeAmountTickSpacing` — the getter `GetFeeAmountTickSpacing` ends just above at line 96, which falls one line before this diff hunk.)

**`GetFeeAmountTickSpacing` returns the internal map by reference.**

A caller in the same realm can mutate the stored `map[uint32]int32` without going through `SetFeeAmountTickSpacing` and its `rlm.IsCurrent()` check. Fix: return a copy.

```go
result := make(map[uint32]int32, len(feeAmountTickSpacing))
for k, v := range feeAmountTickSpacing {
    result[k] = v
}
return result
```

contract/r/gnoswap/pool/store.gno
moul: **`GetPendingProtocolFees` returns the internal map by reference.**

Same pattern as `GetFeeAmountTickSpacing`: callers can mutate the stored `map[string]int64` directly. A caller that reads, modifies, and does *not* call `SetPendingProtocolFees` will still observe the mutation on the next read. Fix: return a defensive copy before returning.

moul (commented):
Review of the Interrealm v2 migration (current HEAD `d703816e`). The direction is solid — realm threading is consistent, `rlm.IsCurrent()` spoof guards are in the right places, and the upgrade scenario tests give real confidence against readonly-taint regressions. Several mutable-reference escapes remain and are noted inline. Two additional issues are in the body below.

**Not in diff — noted here:**

`contract/r/gnoswap/staker/deposit.gno:209-214` — `Warmups()` returns the internal slice directly (`return d.warmups`), and `SetWarmups(warmups []Warmup)` stores it without copying (`d.warmups = warmups`). `Clone()` correctly calls `cloneWarmups()`, but callers of `Warmups()` get the live slice. Fix: `return append([]Warmup(nil), d.warmups...)` in the getter; `d.warmups = append([]Warmup(nil), warmups...)` in the setter.

`contract/r/gnoswap/pool/pool.gno:278,283` — `Slot0.SqrtPriceX96()` returns the raw internal pointer; `SetSqrtPriceX96(v *u256.Uint)` stores it without cloning. `NewSlot0` (line 295) correctly clones on construction. Same fix needed on the getter and setter as was applied to `SetFeeGrowthGlobal0X128` and `SetLiquidity`.

contract/p/gnoswap/store/kv_store.gno
jinoosss: It appears that this issue has been resolved by this PR.
https://github.com/gnoswap-labs/gnoswap/pull/1339

contract/p/gnoswap/store/kv_store.gno
jinoosss: It appears that this issue has been resolved by this PR.
https://github.com/gnoswap-labs/gnoswap/pull/1339

jinoosss: It appears that this issue has been resolved by this PR.
https://github.com/gnoswap-labs/gnoswap/pull/1339

junghoon-vans (approved):

PR #1294, opened 2026-05-26
refactor(p/store): migrate KV store to Interrealm v2

jinoosss:
## Summary

- Migrate `p/gnoswap/store` from Gno Interrealm Spec v1 to v2 (fresh deployment, no backward compatibility).
- All mutating methods now thread an explicit `realm` token through a leading `_ int, rlm realm` parameter pair, replacing the deprecated `runtime.CurrentRealm()` / `runtime.PreviousRealm()` lookups.
- Updates tests, interface, errors, and package documentation in lockstep.

## Motivation

Interrealm v2 removes `runtime.CurrentRealm()` / `runtime.PreviousRealm()` and requires a `realm` capability token to be passed explicitly through every crossing call. The KV store is consumed by every gnoswap realm, so this is the prerequisite for migrating the rest of the codebase.

### `kv_store.gno`

- `Set(_ int, rlm realm, key, value)` — adds `rlm.IsCurrent()` spoof guard, keeps the `IsCode` gate, authorizes against `rlm.Previous().Address()` (the caller).
- `Delete(_ int, rlm realm, key)` — adds spoof guard, unconditional ACL check on `rlm.Previous().Address()` (no `IsCode` bypass — Delete is destructive and should not have the EOA escape hatch).
- `AddAuthorizedCaller` / `UpdateAuthorizedCaller` / `RemoveAuthorizedCaller` — gated on `rlm.Address()` against the domain address.
- `isUpdatableAuthorizedCaller` now takes an `address` parameter so the realm identity is passed in explicitly rather than fetched via `runtime.CurrentRealm()`.
- Removes the `chain/runtime` import.

### `types.gno`

- `KVStore` interface updated to match the new mutating-method signatures.

### `errors.gno`

- Removes `ErrReadPermissionDenied` — there is no read-permission system; all reads are public within the package.
- Adds `errSpoofedRealm`, raised by the new `IsCurrent()` guards.

### `doc.gno`

- Rewritten for v2:
  - Drops the obsolete "None / ReadOnly / Write" three-level permission text.
  - Removes `std` imports from the example.
  - Documents the v1 → v2 identity mapping and the per-method authorization policy explicitly.

### `kv_store_test.gno`

- Top-level test functions: `func TestX(cur realm, t *testing.T)`.
- `t.Run` subtest closures: `func(cur realm, t *testing.T)` — each subtest receives its own realm token, so mid-test `SetRealm` calls are reflected when the subtest writes to the store.
- All write call sites migrated to the `0, cur, ...` form.

## Identity mapping

The mechanical v1 → v2 mapping used throughout the migration:

| v1 | v2 | Meaning |
|---|---|---|
| `runtime.CurrentRealm().Address()` | `rlm.Address()` | Current realm |
| `runtime.PreviousRealm().Address()` | `rlm.Previous().Address()` | Caller realm |

Which one each method authorizes against is **domain logic** and decided per call site. The store's particular choices (Set/Delete check the caller; management ops check the current realm) reflect this store's policy and should not be blindly copied to other modules during their own migrations.

## Summary by CodeRabbit

* **Documentation**
  * Updated KV store documentation to reflect write-only access control model and realm-threading semantics.

* **Refactor**
  * Updated KV store method signatures to include realm context for storage and authorization operations.
  * Removed read permission error handling from access control.

* **Tests**
  * Updated test suite to align with new realm-aware API signatures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gnoswap-labs/gnoswap/pull/1294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/contract/p/gnoswap/store/errors.gno
+++ b/contract/p/gnoswap/store/errors.gno
@@ -11,7 +11,7 @@ var (
 	// Data access errors
 	ErrKeyNotFound            = errors.New("key not found")
 	ErrWritePermissionDenied  = errors.New("write permission denied")
-	ErrReadPermissionDenied   = errors.New("read permission denied")
 	ErrUpdatePermissionDenied = errors.New("update permission denied")
 	ErrFailedCast             = errors.New("failed to cast")
+	errSpoofedRealm           = errors.New("rlm does not match the current crossing frame")
 )
--- a/contract/p/gnoswap/store/errors.gno
+++ b/contract/p/gnoswap/store/errors.gno
@@ -13,5 +13,5 @@ var (
 	ErrWritePermissionDenied  = errors.New("write permission denied")
 	ErrUpdatePermissionDenied = errors.New("update permission denied")
 	ErrFailedCast             = errors.New("failed to cast")
-	errSpoofedRealm           = errors.New("rlm does not match the current crossing frame")
+	ErrSpoofedRealm           = errors.New("rlm does not match the current crossing frame")
 )
--- a/contract/p/gnoswap/store/kv_store.gno
+++ b/contract/p/gnoswap/store/kv_store.gno
@@ -134,7 +134,7 @@ func (k *kvStore) GetBPTree(key string) (*bptree.BPTree, error) {
 // registration.
 func (k *kvStore) Set(_ int, rlm realm, key string, value any) error {
 	if !rlm.IsCurrent() {
-		return errSpoofedRealm
+		return ErrSpoofedRealm
 	}
 
 	if rlm.IsCode() && !k.IsWriteAuthorized(rlm.Previous().Address()) {
@@ -153,7 +153,7 @@ func (k *kvStore) Set(_ int, rlm realm, key string, value any) error {
 // offers for EOA / test callers is intentionally not extended here.
 func (k *kvStore) Delete(_ int, rlm realm, key string) error {
 	if !rlm.IsCurrent() {
-		return errSpoofedRealm
+		return ErrSpoofedRealm
 	}
 
 	caller := rlm.Previous().Address()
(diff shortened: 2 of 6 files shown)


tests/README.md
```diff
@@ -12,7 +12,7 @@ tests/
 │   │   ├── dev.mk       # Development environment configuration
 │   │   ├── staging.mk   # Staging environment configuration
 │   │   └── production.mk # Production environment configuration
-│   ├── deploy.mk        # Deployment scripts
+│   ├── deploy.mk        # Deployment script
```
junghoon-vans: This seems like an unnecessary change. Could you please remove it?

contract/p/gnoswap/store/errors.gno
```diff
@@ -11,7 +11,7 @@ var (
 	// Data access errors
 	ErrKeyNotFound            = errors.New("key not found")
 	ErrWritePermissionDenied  = errors.New("write permission denied")
-	ErrReadPermissionDenied   = errors.New("read permission denied")
 	ErrUpdatePermissionDenied = errors.New("update permission denied")
 	ErrFailedCast             = errors.New("failed to cast")
+	errSpoofedRealm           = errors.New("rlm does not match the current crossing frame")
```
junghoon-vans: Please capitalize the first letter to conform to convention.

contract/p/gnoswap/store/kv_store.gno
```diff
@@ -127,12 +125,19 @@ func (k *kvStore) GetBPTree(key string) (*bptree.BPTree, error) {
...
+// the authorized writers set. EOA / user-realm callers (IsCode == false)
+// bypass the ACL so deploy and test flows can populate state without prior
+// registration.
+func (k *kvStore) Set(_ int, rlm realm, key string, value any) error {
+	if !rlm.IsCurrent() {
+		return ErrSpoofedRealm
+	}
 
-	if currentRealm.IsCode() && !k.IsWriteAuthorized(currentRealm.Address()) {
+	if rlm.IsCode() && !k.IsWriteAuthorized(rlm.Previous().Address()) {
```
junghoon-vans: Since it is a non-cross function, the realm does not change. Why are you checking the previous realm? Previously, you checked the current realm

jinoosss: dc167363e0ddfe8c55c0c58328eda21c00da7eb0

junghoon-vans (approved):

PR #1295, opened 2026-05-26
refactor(p/rbac): remove unused new func

junghoon-vans:


jinoosss (approved):
In addition, it seems we need to provide a default constructor or rename an existing function.
It would be best to proceed with further refactoring.

PR #1296, opened 2026-05-26
refactor(r/access): migrate to interrealm v2

junghoon-vans:
## Summary

Migrate the `r/gnoswap/access` realm to the Interrealm v2 calling model.

This PR is split out from the broader Interrealm v2 migration and focuses only on the access-control realm surface.

## Changes

- Replace `runtime.PreviousRealm()` usage in RBAC-only entrypoints with `cur.Previous()`.
- Thread the current realm through RBAC-to-access calls with `cross(cur)`.
- Preserve the existing role-address registry behavior for setting, removing, looking up, and authorizing roles.
- Keep emitted event fields aligned with the previous behavior while sourcing caller information from the Interrealm v2 `cur` realm.
- Update access tests to pass `cur realm` through test functions, subtests, and panic/abort assertion helpers.

## Scope

- `contract/r/gnoswap/access/access.gno`
- `contract/r/gnoswap/access/access_test.gno`
- `contract/r/gnoswap/access/assert_test.gno`

### DIFF
--- a/contract/r/gnoswap/access/access.gno
+++ b/contract/r/gnoswap/access/access.gno
@@ -2,7 +2,6 @@ package access
 
 import (
     "chain"
-	"chain/runtime"
 	"strings"
 
 	ufmt "gno.land/p/nt/ufmt/v0"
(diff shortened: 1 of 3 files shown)


jinoosss (approved):

jinoosss:
The changes appear to be simple and clear. 
Since there are many dependent contracts, I will merge them as a priority.

PR #1299, opened 2026-05-26
refactor(common): migrate to interrealm v2

junghoon-vans:
## Summary
- Migrates common helper realm calls to Interrealm v2 conventions.
- Updates common package tests to use realm-aware test and uassert signatures.
- Refreshes GRC20 Teller wrapper calls to pass the current realm explicitly.

## Changes
- Replaced native coin assertion dependency with the Interrealm v2 unsafe runtime API.
- Updated common GRC20 helper wrappers to call Teller write methods with `(0, cur, ...)`.
- Converted common tests and subtests to pass `cur realm`, including `cross(cur)` call sites and realm-aware `uassert` helpers.

## Verification
- `make test PKG=gno.land/r/gnoswap/common`

### DIFF
--- a/contract/r/gnoswap/common/grc20reg_helper.gno
+++ b/contract/r/gnoswap/common/grc20reg_helper.gno
@@ -104,7 +104,7 @@ func Allowance(path string, owner, spender address) int64 {
 //
 // Returns an error if the transfer fails, nil otherwise.
 func Transfer(cur realm, path string, to address, amount int64) error {
-	return GetTokenTeller(path).Transfer(to, amount)
+	return GetTokenTeller(path).Transfer(0, cur, to, amount)
 }
 
 // TransferFrom transfers tokens from one address to another using grc20.Teller.TransferFrom.
@@ -118,7 +118,7 @@ func Transfer(cur realm, path string, to address, amount int64) error {
 //
 // Returns an error if the transfer fails, nil otherwise.
 func TransferFrom(cur realm, path string, from, to address, amount int64) error {
-	return GetTokenTeller(path).TransferFrom(from, to, amount)
+	return GetTokenTeller(path).TransferFrom(0, cur, from, to, amount)
 }
 
 // Approve approves tokens for the specified spender using grc20.Teller.Approve.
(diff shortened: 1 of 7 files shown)


contract/r/gnoswap/common/grc20reg_helper.gno
jinoosss, line 1: When using services like Transfer, please verify who is handling your assets.

junghoon-vans: Thanks, verified. This path intentionally uses `CallerTeller`, so the asset actor is derived from the crossing realm rather than from a user-supplied address.

Upstream GRC20 evidence:
- `CallerTeller` resolves the actor as `rlm.Previous().Address()`: https://github.com/gnolang/gno/blob/d8a351e7509ce945267b0d929ead44b981b4061b/examples/gno.land/p/demo/tokens/grc20/tellers.gno#L7-L19
- The write methods first require `rlm.IsCurrent()`, then use that resolved caller/spender for `Transfer`, `Approve`, and `TransferFrom`: https://github.com/gnolang/gno/blob/d8a351e7509ce945267b0d929ead44b981b4061b/examples/gno.land/p/demo/tokens/grc20/tellers.gno#L115-L145
- The `Teller` interface documents that these write calls take the caller captured `cur`: https://github.com/gnolang/gno/blob/d8a351e7509ce945267b0d929ead44b981b4061b/examples/gno.land/p/demo/tokens/grc20/types.gno#L43-L73

So these common helpers intentionally stay as crossing adapters: `common.Transfer(cross(cur), ...)` creates the common crossing frame, and forwarding that live `cur` into `Transfer(0, cur, ...)` makes `CallerTeller` see `cur.Previous()` as the caller realm whose assets should move/approve. I strengthened the existing `TestGrc20regHelper_Transfer`, `TestGrc20regHelper_Approve`, and `TestGrc20regHelper_TransferFrom` cases to assert the actual balance/allowance changes, then re-ran `make test PKG=gno.land/r/gnoswap/common`.

jinoosss (approved):

PR #1300, opened 2026-05-27
refactor(gns): migrate to interrealm v2

junghoon-vans:
## Summary
- Migrate the GNS realm and package tests to Interrealm v2 cur realm / cross(cur) call patterns.
- Update GNS token/emission entrypoints to use cur.Previous() and v2 GRC20 APIs.
- Keep abort expectations unchanged while adjusting assertion callback shape for v2 caller frames.

## Dependencies
- Depends on https://github.com/gnoswap-labs/gnoswap/pull/1296 for access Interrealm v2 migration.
- Depends on https://github.com/gnoswap-labs/gnoswap/pull/1297 for rbac Interrealm v2 migration.

## Validation
- make test PKG=gno.land/r/gnoswap/gns passed with local direct dependency migrations applied.
- GNS-only PR scope intentionally excludes contract/r/gnoswap/access and contract/r/gnoswap/rbac; without those dependency migrations, make test PKG=gno.land/r/gnoswap/gns currently fails at access.gno with name PreviousRealm not declared.

### DIFF
--- a/contract/r/gnoswap/gns/_helper_test.gno
+++ b/contract/r/gnoswap/gns/_helper_test.gno
@@ -16,17 +16,18 @@ import (
 
 var adminAddr = access.MustGetAddress(prabc.ROLE_ADMIN.String())
 
-func resetObject(t *testing.T) {
+func resetObject(cur realm, t *testing.T) {
 	t.Helper()
 
-	resetGnsTokenObject(t)
+	resetGnsTokenObject(cur, t)
 	resetEmissionState(t)
 }
 
-func resetGnsTokenObject(t *testing.T) {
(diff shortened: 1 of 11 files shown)


jinoosss (approved):

junghoon-vans:
I had previously saved the `userTeller` but wasn't using it. I modified the code to use the saved value directly, rather than going through the `token.CallerTeller()`. [2de2847](https://github.com/gnoswap-labs/gnoswap/pull/1300/commits/2de2847e4e89d7bacf00c99191dbf4317721e7ec) @jinoosss

jinoosss (approved):

PR #1305, opened 2026-05-27
refactor(r/gnft): migrate to interrealm v2

junghoon-vans:
## Summary
- Migrate GNFT realm entrypoints and tests to Interrealm v2.
- Keep GNFT transfer and approval behavior aligned with GRC721 owner/approved/approved-for-all checks.
- Refresh GNFT README examples to use `cross(cur)` and document staked-token movement rules.

## Verification
- `make test PKG=gno.land/r/gnoswap/gnft`
- `GIT_MASTER=1 git diff --check origin/refactor/interrealm-v2...HEAD`

### DIFF
--- a/contract/r/gnoswap/gnft/utils.gno
+++ b/contract/r/gnoswap/gnft/utils.gno
@@ -25,7 +25,7 @@ func checkErr(err error) {
 }
 
 // checkTransferErr wraps transfer errors with more specific context.
-func checkTransferErr(caller address, err error, from, to address, tid grc721.TokenID) {
+func checkTransferErr(err error, caller, from, to address, tid grc721.TokenID) {
 	if err == nil {
 		return
 	}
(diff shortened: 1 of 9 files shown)


contract/r/gnoswap/gnft/utils.gno
```diff
@@ -26,13 +25,11 @@ func checkErr(err error) {
 }
 
 // checkTransferErr wraps transfer errors with more specific context.
-func checkTransferErr(err error, from, to address, tid grc721.TokenID) {
```
jinoosss, line 29: It seems that the `caller` argument is in a different context from the function. If it's being added, it would be clearer to place it at the end.

junghoon-vans: [b57af8e](https://github.com/gnoswap-labs/gnoswap/pull/1305/commits/b57af8e64438f2b77e4dac3b2ec37377e08e1c65)

I think it would be better to use it as the second argument rather than passing it as the last one. By keeping the `err` argument as the main one, you can organize the signatures of similar functions in a similar order. 

```go
func SafeTransferFrom(cur realm, from, to address, tid grc721.TokenID) error {
	assertFromIsValidAddress(from)
	assertToIsValidAddress(to)
	caller := cur.Previous().Address()
	assertIsAllowedTransfer(caller, tid)

	err := nft.SafeTransferFrom(caller, from, to, tid)
	checkTransferErr(err, caller, from, to, tid)
	return nil
}
```

jinoosss (approved):

PR #1307, opened 2026-05-27
refactor(p/version_manager): migrate initializer to interrealm v2

jinoosss:
## Summary

Propagate the v2 interrealm spec into the inner-closure layer of `version_manager`. The realm-threading marker `(_ int, rlm realm, ...)` is now consistently applied to the `initializer` and `initializeDomainStoreFn` signatures, matching the already-migrated `RegisterInitializer` / `ChangeImplementation` outer surface.

## Why

`RegisterInitializer(_ int, rlm realm, ...)` and `ChangeImplementation(_ int, rlm realm, ...)` were already migrated to the v2 spec, but their inner callbacks still used the v1 signatures `func(store any) any` and `func(kvStore store.KVStore) any`. This inconsistency hid realm context from the storage-initialization layer and left a half-migrated boundary inside the package.

This PR threads the live realm token end-to-end:

- `version_manager.initializers` now stores `map[string]func(_ int, rlm realm, store any) any`
- `initializeDomainStoreFn` now takes `(_ int, rlm realm, kvStore store.KVStore) any`
- All inner closures inside `RegisterInitializer` and `ChangeImplementation` call sites now visibly thread `0, rlm`

The domain-level `RegisterInitializer` public surface is left untouched to keep `v1/init.gno` callers (and scenario test implementations) unaffected. Only the proxy → version_manager bridge closure is updated.

### `contract/p/gnoswap/version_manager`
- `types.gno`: `VersionManager` interface — `RegisterInitializer` and `GetInitializers` signatures updated to the v2 marker pattern.
- `version_manager.gno`: `versionManager` struct fields (`initializers`, `initializeDomainStoreFn`), `NewVersionManager` parameter, and `makeInitializerSafe` updated. Internal calls now pass `0, rlm`.
- `version_manager_test.gno`: `mockInitializer`, `mockInitializeDomainStoreFn`, map literals, and the `TestMakeInitializerSafe_InvalidType` valid-input case updated to the new signature.
- `doc.gno`: usage examples updated.

### 8 domain proxy realms (`contract/r/gnoswap/{protocol_fee,pool,position,staker,router,launchpad,gov/governance,gov/staker}`)
- `state.gno`: `initializeDomainStore` parameter list now `(_ int, rlm realm, kvStore store.KVStore)`.
- `upgrade.gno`: inner `initializerFunc` closure is now `func(_ int, rlm realm, domainStore any) any`. Public `RegisterInitializer` signatures (the domain-specific one consumed by `v1/init.gno`) are unchanged.
- `_helper_test.gno`: inline `NewVersionManager` closure updated.

### `contract/r/gnoswap/gov/staker/state_test.gno`
- `TestState_initializeDomainStore` updated to accept `cur realm` and pass `0, cur` to the now-realm-aware `initializeDomainStore`.

## Compatibility

- `v1/init.gno` files across all 8 domains are unaffected — they still call the domain-specific `RegisterInitializer(cur realm, initializer func(...) I...)`.
- Scenario test implementations under `contract/r/scenario/upgrade/implements/**` are unaffected for the same reason.
- No on-chain storage layout changes.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -425,6 +425,7 @@ No data migration required
 - Role-based storage permissions
 - Namespace isolation
 - Admin-only upgrade functions
+- Realm tokens are threaded explicitly into the version manager and validated via `rlm.IsCurrent()`, rejecting spoofed or stale crossing-frame tokens
 
 ### 3. Emergency Halt
 
@@ -459,14 +460,18 @@ func init() {
 // In pool/upgrade.gno
 func UpgradeImpl(cur realm, packagePath string) {
     caller := runtime.PreviousRealm().Address()
-    access.AssertIsAdmin(caller)
+    access.AssertIsAdminOrGovernance(caller)
 
-    if _, ok := initializers[packagePath]; !ok {
-        panic("Initializer not found")
+    // Thread the live crossing-frame token into the version manager. The
+    // leading 0 is the v2 interrealm sentinel; the manager validates
+    // rlm.IsCurrent() to reject spoofed/stale tokens before switching.
+    if err := versionManager.ChangeImplementation(0, cur, packagePath); err != nil {
+        panic(err)
     }
 
-    poolImpl = initializers[packagePath](NewPoolStore(kvStore))
-    // Update storage permissions...
+    if err := updateImplementation(); err != nil {
+        panic(err)
+    }
 }
 ```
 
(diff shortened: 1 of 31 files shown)


junghoon-vans (approved):

junghoon-vans:
The code changes look good. However, it would be best to update `ARCHITECTURE.md` and `version_manager/README.md` to reflect these changes as well.

junghoon-vans (approved):

PR #1308, opened 2026-05-27
refactor(r/protocol_fee): migrate to interrealm v2

jinoosss:
## Summary

- Migrates the `protocol_fee` proxy, v1 implementation, KV store layer, and tests to Interrealm v2 call semantics.
- Threads the crossing `realm` from proxy entry points through implementation methods, store writes, and GRC-20 transfers.
- Replaces `runtime.PreviousRealm()` / bare `cross` usage with `rlm.Previous()`, `cross(rlm)`, and explicit realm setup in tests.

### Proxy & realm initialization
- Switches package `init()` to `init(cur realm)` and captures `currentAddress` / `domainPath` from `cur` instead of `runtime.CurrentRealm()`.
- Documents that package-init paths without a crossing frame are the only place `chain/runtime/unsafe.CurrentRealm` is used (test reset helper included).

### Crossing API (`_ int, rlm realm`)
- Extends `IProtocolFee`, `IProtocolFeeStore`, and initializer signatures so mutating paths accept `_ int, rlm realm`.
- Proxy forwards `cur` into the active implementation as literal `0` + `cur` (e.g. `getImplementation().AddToProtocolFee(0, cur, ...)`).
- KV store writes go through `kvStore.Set(0, rlm, ...)` so mutations run under the proxy’s write ACL.

### v1 implementation & security
- Updates `DistributeProtocolFee`, `SetDevOpsPct`, `SetGovStakerPct`, and `AddToProtocolFee` to:
  - assert `rlm.IsCurrent()` (new `errSpoofedRealm` on mismatch),
  - resolve callers via `rlm.Previous()` for access checks and events,
  - pass `rlm` into state/store helpers and `common.SafeGRC20Transfer(From)(cross(rlm), ...)`.
- Threads the proxy realm through `RegisterInitializer` / `initStoreData` so v1 bootstrap writes use the proxy identity, not the v1 package path.
- `UpgradeImpl` uses `cur.Previous()` instead of `runtime.PreviousRealm()` for admin/governance checks.

### Tests
- Converts tests to Interrealm v2 style: top-level `func(cur realm, t *testing.T)`, subtests `func(cur realm, t *testing.T)`, and `cross(cur)` at external call sites.
- Refactors token funding helpers (`transferTokenTo`, `approveProtocolFeeSpender`, `fundRealmAndApproveProtocolFee`) to take `cur realm` and cross explicitly.
- Sets caller realms with `testing.SetRealm(...)` before exercising authorized/unauthorized paths.
- Restores overflow coverage for `AddToProtocolFee` via explicit crossing (`pf.AddToProtocolFee(0, cur, ...)` inside `cross(cur)`) and asserts the machine abort with `uassert.AbortsContains`.
- Updates proxy, store, upgrade, assert, getter, instance, and state unit tests and mocks to the new store/interface shapes.

### DIFF
--- a/contract/r/gnoswap/protocol_fee/store.gno
+++ b/contract/r/gnoswap/protocol_fee/store.gno
@@ -45,6 +45,10 @@ func (s *protocolFeeStore) HasDevOpsPctStoreKey() bool {
 }
 
 func (s *protocolFeeStore) InitializeDevOpsPct(_ int, rlm realm) error {
+	if rlm.IsCurrent() {
+		return s.kvStore.Set(0, rlm, StoreKeyDevOpsPct.String(), defaultDevOpsPct)
+	}
+
 	return s.kvStore.Set(0, rlm, StoreKeyDevOpsPct.String(), defaultDevOpsPct)
 }
 
@@ -58,6 +62,10 @@ func (s *protocolFeeStore) GetDevOpsPct() int64 {
 }
 
 func (s *protocolFeeStore) SetDevOpsPct(_ int, rlm realm, pct int64) error {
+	if rlm.IsCurrent() {
+		return s.kvStore.Set(0, rlm, StoreKeyDevOpsPct.String(), defaultDevOpsPct)
+	}
+
 	return s.kvStore.Set(0, rlm, StoreKeyDevOpsPct.String(), pct)
 }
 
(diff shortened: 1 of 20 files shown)


contract/r/gnoswap/protocol_fee/store.gno
```diff
@@ -44,8 +44,12 @@ func (s *protocolFeeStore) HasDevOpsPctStoreKey() bool {
...
 
-func (s *protocolFeeStore) InitializeDevOpsPct() error {
-	return s.kvStore.Set(StoreKeyDevOpsPct.String(), defaultDevOpsPct)
+func (s *protocolFeeStore) InitializeDevOpsPct(_ int, rlm realm) error {
+	if rlm.IsCurrent() {
+		return s.kvStore.Set(0, rlm, StoreKeyDevOpsPct.String(), defaultDevOpsPct)
+	}
+
+	return s.kvStore.Set(0, rlm, StoreKeyDevOpsPct.String(), defaultDevOpsPct)
 }
```
junghoon-vans, line 53: I don't think the branch is necessary.

junghoon-vans (approved):

PR #1315, opened 2026-05-28
refactor(r/gov/staker): migrate to interrealm v2

jinoosss:


jeronimoalbi (commented):
InterrealmV2 changes look good 👍

Not important; Some unit tests or sub-tests have `cur realm` definitions that are not needed, which could be removed, but I'm guessing they are defined to keep refactor simple by replacing matches within test files.

PR #1333, opened 2026-06-10
fix(staker): allowed token slice read-only taint

jinoosss:
## Summary
- Fix a `cannot append to readonly tainted slice` panic when `AddToken` / `RemoveToken` mutate the allowed-tokens list after a contract upgrade (e.g. v1 → v2_valid).
- Move allowed-token list read-modify-write logic into the staker store domain via `AddAllowedToken` and `RemoveAllowedToken`.
- Simplify `staker/v1` by removing the `modifyTokenList` validator/executor abstraction and keeping only default-token business rules at the v1 layer.
## Problem
`GetAllowedTokens()` returns a readonly-tainted slice from the domain store. The previous v1 implementation appended directly to that slice inside `addTokenExecutor`, which panics on the second add after a remove (observed in `staker_v2_valid_upgrade_filetest` scenario 8).
Typical failure path:
1. First `AddToken` — token already present, no append → succeeds
2. `RemoveToken` — subslice append creates a new slice → succeeds
3. Second `AddToken` — direct append on readonly slice → panic
## Solution
- Add `AddAllowedToken` / `RemoveAllowedToken` to `IStakerStore` and `stakerStore`, following the same RMW pattern as other domain setters (e.g. `SetTokenSpecificMinimumRewardItem`).
- Keep v1 `AddToken` / `RemoveToken` thin: admin/governance checks, default-token validation, event emission, and delegation to the store.
- Split default-token validation into `checkCannotAddDefaultToken` and `checkCannotRemoveDefaultToken` so remove uses `errDefaultExternalToken` instead of the add error.
- Expose `DefaultAllowedTokens()` from the staker package and use it during init.


junghoon-vans:
Blocking: `contract/r/gnoswap/staker/v1/init_test.gno:15` writes directly into the slice returned from the store:

```go
firstStoreAllowedTokens[0] = "gno.land/r/gnoswap/custom"
```

Under Interrealm v2 this object is readonly-tainted, so the package test fails before it can verify fresh defaults. I reproduced it locally with:

```bash
make test PKG=gno.land/r/gnoswap/staker/v1 RUN=TestInit_initStoreDataUsesFreshAllowedTokensPerStore
```

Failure:

```text
panic: cannot directly modify readonly tainted object (use a method or crossing function): firstStoreAllowedTokens...
```

Please rewrite this test so it does not mutate the returned slice directly. For example, mutate through a store-owned API, or use another taint-safe assertion to prove `initStoreData` allocates fresh default allowed-token state.

junghoon-vans (approved):

PR #1336, opened 2026-06-12
fix(pool): suppress DrySwap PoolTickCross events

junghoon-vans:
## Summary

- Prevent `DrySwap` simulations from emitting canonical `PoolTickCross` events when simulated price movement crosses initialized ticks.
- Preserve `PoolTickCross` emission and staker tick-cross hook dispatch for real `Swap` execution.
- Add integration coverage proving `DrySwap` emits no tick-cross events while a real swap across the same tick still dispatches the staker batch event.

## Context

`DrySwap` reuses the normal swap computation path against a cloned pool. Before this change, `tickCross` emitted `PoolTickCross` directly, so a simulation could leak swap-like canonical events even though no real pool settlement occurred.

The staker hook path was already gated by the optional tick-cross callback and was not the active issue. This change moves `PoolTickCross` emission behind the same real-`Swap` callback boundary, so simulated tick crossing no longer produces canonical tick-cross events.

## Changes

- Move `PoolTickCross` emission from `tickCross` into the `Swap`-only tick-cross callback.
- Keep `DrySwap` on the shared computation path with a nil tick-cross callback.
- Update the tick-cross callback signature so real swap event payloads are still built from the crossed pool tick state.
- Add a staker integration fixture that rejects `PoolTickCross`, `StakerTickCross`, and `BatchStakerTickCross` during `DrySwap`, then confirms real swap behavior still emits the expected staker batch event.

## Verification

- `git diff --check`
- `make test PKG=gno.land/r/gnoswap/pool/v1`
- `make integration-test-run TEST=staker_dryswap_does_not_dispatch_tick_cross_hook`

### DIFF
--- a/contract/r/gnoswap/pool/v1/swap.gno
+++ b/contract/r/gnoswap/pool/v1/swap.gno
@@ -239,12 +239,18 @@ func (i *poolV1) Swap(
 	}
 
 	var onTickCross tickCrossHookFn
+	var hook func(cur realm, poolPath string, tickId int32, zeroForOne bool, timestamp int64)
 	if i.store.HasTickCrossHook() {
-		hook := i.store.GetTickCrossHook()
+		hook = i.store.GetTickCrossHook()
+	}
+	onTickCross = func(pool *pl.Pool, tickId int32, zeroForOne bool, timestamp int64) {
+		chain.Emit(
+			"PoolTickCross",
+			"poolPath", pool.PoolPath(),
+			"tick", NewTickEventInfo(tickId, *mustGetTick(pool, tickId)).ToString(),
+		)
 		if hook != nil {
-			onTickCross = func(poolPath string, tickId int32, zeroForOne bool, timestamp int64) {
-				hook(cross(rlm), poolPath, tickId, zeroForOne, timestamp)
-			}
+			hook(cross(rlm), pool.PoolPath(), tickId, zeroForOne, timestamp)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


contract/r/gnoswap/pool/v1/swap.gno
```diff
@@ -239,12 +239,18 @@ func (i *poolV1) Swap(
 	}
 
 	var onTickCross tickCrossHookFn
+	var hook func(cur realm, poolPath string, tickId int32, zeroForOne bool, timestamp int64)
 	if i.store.HasTickCrossHook() {
-		hook := i.store.GetTickCrossHook()
+		hook = i.store.GetTickCrossHook()
+	}
+	onTickCross = func(pool *pl.Pool, tickId int32, zeroForOne bool, timestamp int64) {
```
jinoosss: Is there a reason to create a callback method even when the hook is nil?
Since we check for nil when actually using the hook, this could cause confusion.
https://github.com/gnoswap-labs/gnoswap/blob/6c16e608d4a33bb7a56d6ea7270c5df627e32941/contract/r/gnoswap/pool/v1/swap.gno#L779C6-L779C24

contract/r/gnoswap/pool/v1/swap.gno
```diff
@@ -239,12 +239,18 @@ func (i *poolV1) Swap(
...
+		chain.Emit(
+			"PoolTickCross",
+			"poolPath", pool.PoolPath(),
+			"tick", NewTickEventInfo(tickId, *mustGetTick(pool, tickId)).ToString(),
+		)
 		if hook != nil {
-			onTickCross = func(poolPath string, tickId int32, zeroForOne bool, timestamp int64) {
-				hook(cross(rlm), poolPath, tickId, zeroForOne, timestamp)
-			}
+			hook(cross(rlm), pool.PoolPath(), tickId, zeroForOne, timestamp)
```
jinoosss, line 252: Emit the event after the hook has executed. 
The event data will be output even if the state has not changed.

jinoosss (changes requested):

jinoosss (approved):

jinoosss (approved):

PR #1338, opened 2026-06-12
refactor(p/consts): define constants as constructors in the consts package

jinoosss:
## Summary

Eliminates package-level **mutable singleton** values (`*u256.Uint` / `*i256.Int`
created via `MustFromDecimal`, `Lsh`, `Sub`, etc.) across the math library and
core realms, replacing them with **constructor functions**. A new
`p/gnoswap/consts` package centralizes constants that were previously duplicated
or independently defined in multiple packages.

### Motivation

`*u256.Uint` and `*i256.Int` are pointers to mutable values whose methods can
mutate the receiver in place (e.g. `z[0] = ...`). Exposing them as package-level
`var`s means every caller shares the same backing pointer, so an accidental
in-place mutation — or using one as a `Set`/arithmetic source — silently
corrupts the value for all other callers and realms.

For `/p/` packages the VM blocks in-place mutation of immutable-package objects,
but we should not rely on that runtime guard: a new caller attempting an in-place
op, or feeding one of these into a mutating operation, would reintroduce the
`/p/uint256` PkgID-based panic. Returning a fresh instance per call removes the
shared-pointer hazard entirely.

### New package: `p/gnoswap/consts`

- 12 constructor functions for protocol-wide constants:
  `Q96`, `Q128`, `Q192`, `Max160`, `MaxUint128`, `MaxUint256`, `MaxInt256`,
  `MinSqrtRatio`, `MaxSqrtRatio`, `MaxInt128I256`, `MaxInt64I256`, `MinInt64I256`.
- Each returns a freshly allocated value, so callers never share a backing
  instance.
- Values are built from little-endian `[4]uint64` array literals (both `Uint`
  and `Int` are defined as `[4]uint64`) instead of `MustFromDecimal`, avoiding
  decimal-string parsing on every call.
- Includes `consts_test.gno` asserting each constructor equals its
  `MustFromDecimal` equivalent and that successive calls return distinct
  pointers.

### `p/gnoswap/gnsmath`

- `tick_math.gno`: converted the `ratioConstants` slice (a slice plus 19 exposed
  pointers — the "globally exposed mutable array" anti-pattern), `ratio0/1`, the
  MSB thresholds, the tick offsets, and `oneLsh32` into constructor functions.
  Shared constants (`MaxUint256`, `MinSqrtRatio`, `MaxSqrtRatio`) now come from
  `consts`. Removed now-redundant nil-initialization guards.
- `sqrt_price_math.gno`: `q96` / `max160` / `maxInt256` → `consts`;
  `MIN_SQRT_RATIO` / `MAX_SQRT_RATIO` changed from package vars to functions.
- `liquidity_math.gno`, `bit_math.gno`: `q96Uint`, `maxUint128`, and the
  `msbShifts` / `lsbShifts` tables (mutable pointers inside slices) replaced with
  constructors / `consts`.

### Realms

- **q128 (2^128) deduplication**: removed 5 independent `var q128 =
  MustFromDecimal(...)` definitions (`staker`, `staker/v1`, `position/v1`,
  `launchpad/v1`, `gov/staker/v1`), all now using `consts.Q128()`.
- **`pool/v1`**:
  - `swap.gno`: inlined the read-only `zero` / `zeroI256` singletons to
    `u256.Zero()` / `i256.Zero()`; `maxInt256` / `maxInt64` / `minInt64` /
    `fixedPointQ128` → `consts`.
  - `factory_param.gno` (`Q192`, sqrt-ratio bounds), `oracle.gno` (`max160`),
    `utils.gno` (`uint128Mask`, `maxInt128`/`maxUint128` decimals) → `consts`.
  - `tick.gno`: `calculateMaxLiquidityPerTick` previously returned a shared
    singleton pointer directly to callers — the highest-risk case for
    cross-caller corruption. The four `maxLiquidityPerTickSpacing*` values are
    now constructor functions returning fresh instances.


junghoon-vans (commented):
I found one deployment blocker around the new consts package gas configuration.

tests/scripts/deploy.mk
```diff
@@ -44,6 +44,11 @@ deploy-int256:
 	@echo "" | gnokey maketx addpkg -pkgdir $(ROOT_DIR)/contract/p/gnoswap/int256 -pkgpath gno.land/p/gnoswap/int256 -insecure-password-stdin=true -remote $(GNOLAND_RPC_URL) -broadcast=true -chainid $(CHAINID) -gas-fee 17924ugnot -gas-wanted 17924000 -memo "" gnoswap_admin
 	@echo
 
+deploy-consts:
+	$(info ************ deploy consts ************)
+	@echo "" | gnokey maketx addpkg -pkgdir $(ROOT_DIR)/contract/p/gnoswap/consts -pkgpath gno.land/p/gnoswap/consts -insecure-password-stdin=true -remote $(GNOLAND_RPC_URL) -broadcast=true -chainid $(CHAINID) -gas-fee 20000ugnot -gas-wanted 20000000 -memo "" gnoswap_admin
```
junghoon-vans: `deploy-consts` is configured with `-gas-wanted 20000000`, but deploying the new `gno.land/p/gnoswap/consts` package used 35,312,863 gas when I ran the deploy txtar path. This target can fail out-of-gas in the real deployment flow. Please raise `gas-wanted` above the observed usage with margin, and adjust the gas fee consistently with the other deploy targets.

PR #1343, opened 2026-07-15
Feature/performance metrics framework

notJoon:
wip


jinoosss:
@notJoon Is this PR valid?

PR #1349, opened 2026-07-22
feat(event): emit reward accumulation and pool-tier reward events for reconciliation

jinoosss:
## Summary

Emit authoritative reward data on-chain so the batch/indexer can reconcile rewards
directly instead of re-deriving the accumulator off-chain (which requires replicating
the contract's fixed-point math, halving schedule, distribution ratios, and rounding — a
recurring source of divergence).

The reward formula is unchanged (MasterChef style); the contract now emits the global
accumulator (and, for the pool staker, the per-tier emission rate) at every point it advances:

```
pending = (globalAccX128 − userRewardDebtX128) × userStake / 2^128
```

This PR covers three reward surfaces: the **pool staker** tier events, the **gov staker**
emission / protocol-fee accumulators, and the **launchpad** per-tier deposit accumulator.

### 1. Pool staker — tier reward rate on tier events
`contract/r/gnoswap/staker/v1/manage_pool_tier_and_warmup.gno`,
`reward_calculation_pool_tier.gno`

- `SetPoolTier`, `ChangePoolTier`, and `RemovePoolTier` events now include
  `rewardPerSecond` and per-tier `tier1RewardPerSecond` / `tier2RewardPerSecond` /
  `tier3RewardPerSecond`.
- `PoolTier.changeTier` now returns the per-tier reward map (`map[uint64]int64`), and the
  internal `setPoolTier` / `changePoolTier` / `removePoolTier` helpers return
  `currentEmission` + that map so the event sites can emit them.

### 2. Gov staker — emission & protocol-fee accumulation events
`contract/r/gnoswap/gov/staker/v1/state.gno`

- Emit `UpdateEmissionRewardAccumulation` and `UpdateProtocolFeeRewardAccumulation` on
  **every** path that advances the accumulator — add / remove / claim (6 sites) — via the
  helpers `emitUpdateEmissionRewardAccumulation` and
  `emitUpdateProtocolFeeRewardAccumulation`.
- Emission emits the **stored baseline** `GetDistributedAmount()` (not the live local value)
  so the emitted tuple stays consistent with the persisted state used for the next delta.
- Protocol fee handles a nil per-token accumulator safely (`"0"`).

Event → batch `governance_reward_manager` column mapping:

| Event field | Column |
|---|---|
| `accumulatedRewardX128PerStake` / `accumulatedProtocolFeeX128PerStake` | `accumulated_reward_per_delegation_x128` |
| `totalStakedAmount` | `total_delegated_amount` |
| `distributedAmount` / `protocolFeeAmount` | `total_distributed_amount` |
| `accumulatedTimestamp` | `accumulated_timestamp` |
| `tokenPath` (protocol fee) | `token_path` |

### 3. Launchpad — per-tier deposit accumulation events
`contract/r/gnoswap/launchpad/v1/reward_manager.gno` and call sites in
`launchpad_deposit.gno`, `launchpad_withdraw.gno`, `launchpad_reward.gno`,
`launchpad_project.gno`

- New `UpdateLaunchpadRewardAccumulation` event emitted on all 4 paths that call
  `updateRewardPerDepositX128` (deposit / withdraw / collect / admin-reclaim), via the
  helper `emitUpdateLaunchpadRewardAccumulation`.
- Launchpad uses a `rate × time ÷ deposit` model (not a cumulative distributed amount), so
  the event carries a distinct field set: `projectTierId`, `totalDepositAmount`,
  `accumulatedRewardPerDepositX128`, `distributeAmountPerSecondX128`, `accumulatedTime`.
  `projectTierId` is required because the reward manager is keyed per tier.

## Consumer notes (batch)

- Timestamps are **Unix seconds**, not block height.
- The accumulation event and the user stake-change event (`Delegate` / `DepositGns` / …)
  fire in the same tx — apply the accumulation update **first**, then settle the user and
  reset `rewardDebt = newAcc` (MasterChef order).
- `updateRewardPerDepositX128` may early-return without advancing the accumulator; the event
  still fires with unchanged values — consume idempotently.
- **Follow-up (not in this PR):** the launchpad project-reward path in the batch service has
  no table consuming `UpdateLaunchpadRewardAccumulation` yet; a `projectTierId`-keyed
  accumulator snapshot table + forward projection via `distributeAmountPerSecondX128` is
  required.

## Summary by CodeRabbit

* **New Features**
  * Enhanced pool-tier management events now include the current reward emission rate, plus per-tier reward breakdowns (Tier 1–3) when tiers are set, changed, or removed.
  * Added/expanded reward accumulation events for emission rewards, protocol fee rewards, and launchpad rewards during stake/deposit/withdraw/claim flows.
* **Bug Fixes**
  * Updated tick-crossing event data to include the correct `lastTick` value for consistent off-chain reward calculation.
* **Tests**
  * Adjusted pool-tier tests to match the updated return values.


junghoon-vans (approved):
I checked this against the batch flow and the reward pipeline docs. The governance and launchpad accumulator events look aligned, and the pool-tier reward event direction makes sense as well.

Only minor follow-up: please guard the no-op `changeTier` case so it cannot emit an empty/zero per-tier reward snapshot on same-tier updates. Other than that, looks good.

PR #1357, opened 2026-07-29
fix(pool): guard oracle lookback before epoch

junghoon-vans:
## Summary
- return the existing oracle-too-old error before observeSingle can create a negative int64 target
- make lte use signed-safe timestamp adjustment instead of u256 conversion
- add regression coverage for pre-epoch TWAP lookbacks and Uniswap-style wrap ordering

## Uniswap comparison
Uniswap V3 computes target as uint32 time - secondsAgo and relies on uint32 wrap-around plus the OLD guard in getSurroundingObservations. GnoSwap uses int64 Unix timestamps, so an overlarge secondsAgo creates a real negative target. This patch preserves the OLD-error semantics while avoiding the u256 negative conversion panic.

## Tests
- gno test -v -run 'TestOracleLTE|TestOracleObserve|TestOracle_TWAPConsistencyCases' ./gno.land/r/gnoswap/pool/v1
- gno test -v ./gno.land/r/gnoswap/pool/v1

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected oracle timestamp comparison logic to use natural `int64` ordering (avoiding unintended uint32-style wrap behavior).
  * Oracle queries now fail with clear errors when the requested window goes before the Unix epoch.
  * Added an early rejection for invalid lookbacks where the requested `secondsAgo` exceeds the current time.

* **Tests**
  * Expanded oracle tests for `lte` ordering across equality and signed values, including a negative wraparound case.
  * Added new oracle observe and TWAP consistency failure scenarios for invalid time-window requests.


contract/r/gnoswap/pool/v1/oracle.gno
```diff
@@ -176,23 +176,22 @@ func grow(os *pl.ObservationState, currentCardinality, nextCardinality uint16) (
...
 	}
 
-	if bAdjusted.Gt(timeUint256) {
-		bAdjusted = bAdjusted.Add(bAdjusted, u256.NewUintFromInt64(1<<32))
+	if b <= time {
+		bAdjusted += 1 << 32
 	}
 
-	return aAdjusted.Lte(bAdjusted)
+	return aAdjusted <= bAdjusted
```
jinoosss: It must be changed to match the int64 type. 
https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/Oracle.sol#L128-L140

junghoon-vans: Agreed. Since GnoSwap stores oracle timestamps as monotonic `int64` Unix timestamps, we do not need Uniswap's `uint32` wraparound adjustment/shifting here in the first place. I removed that logic and made `lte` use natural `int64` ordering instead, with a test that explicitly rejects the old uint32-wrap expectation.

junghoon-vans: Adding a bit more context after checking the upstream implementation.

In Uniswap V3, the EVM/native block timestamp is not inherently `uint32`; Uniswap intentionally truncates it to a 32-bit modular clock:

https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/UniswapV3Pool.sol#L119-L123

```solidity
/// @dev Returns the block timestamp truncated to 32 bits, i.e. mod 2**32. This method is overridden in tests.
function _blockTimestamp() internal view virtual returns (uint32) {
    return uint32(block.timestamp); // truncation is desired
}
```

That is why `Oracle.Observation.blockTimestamp` is also `uint32` in Uniswap:

https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/Oracle.sol#L13-L20

The `lte(time, a, b)` helper is therefore a comparator for this truncated `uint32` time domain, not a general timestamp comparator:

https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/Oracle.sol#L122-L140

The key assumption is in the comment: it is safe for `0 or 1 overflows`, and `a`/`b` must be chronologically before or equal to `time`. Under that assumption, values greater than `time` are interpreted as timestamps from the previous uint32 cycle, while values less than or equal to `time` are in the current cycle and get shifted by `2**32` for comparison.

Example:

```text
time = 5
a    = 4294967294  // before uint32 wrap
b    = 2           // after uint32 wrap
```

A plain numeric comparison would say `4294967294 <= 2 == false`, but chronologically `a` is older than `b`. Uniswap adjusts this to:

```text
aAdjusted = 4294967294
bAdjusted = 2 + 2**32 = 4294967298
```

so the chronological comparison becomes true.

This also matches how Uniswap computes the lookback target:

https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/Oracle.sol#L245-L286

```solidity
uint32 target = time - secondsAgo;
```

Since `time` and `target` are both `uint32`, the target can wrap modulo `2**32`, and the surrounding-observation search needs the `lte()` wrap-aware comparator.

GnoSwap is different. We store oracle timestamps as `int64` Unix timestamps, not as a truncated uint32 modular clock:

https://github.com/gnoswap-labs/gnoswap/blob/8d53d10e8e3c72fe5ce490cf5095f161725e5b53/contract/r/gnoswap/pool/oracle.gno#L38-L47

`currentTime` comes from `time.Now().Unix()`:

https://github.com/gnoswap-labs/gnoswap/blob/8d53d10e8e3c72fe5ce490cf5095f161725e5b53/contract/r/gnoswap/pool/v1/oracle.gno#L17-L30

and the user-controlled value is only `secondsAgo`, which is bounded before constructing `target`:

https://github.com/gnoswap-labs/gnoswap/blob/8d53d10e8e3c72fe5ce490cf5095f161725e5b53/contract/r/gnoswap/pool/v1/oracle.gno#L270-L274

```go
if int64(secondsAgo) > currentTime {
    return 0, "", errors.New(errObservationTooOld)
}

target := currentTime - int64(secondsAgo)
```

So after this guard, `target` is a normal non-negative `int64` timestamp. There is no `uint32(block.timestamp)` truncation and no `2**32` modular time domain to normalize. The correct comparator for GnoSwap is therefore just natural `int64` ordering:

https://github.com/gnoswap-labs/gnoswap/blob/8d53d10e8e3c72fe5ce490cf5095f161725e5b53/contract/r/gnoswap/pool/v1/oracle.gno#L178-L180

```go
func lte(a, b int64) bool {
    return a <= b
}
```

So the conclusion is: Uniswap needs the shifting because it deliberately converts native time into a `uint32` modulo clock; GnoSwap keeps native Unix time as `int64`, so carrying over the shifting would mix two different time models.

jinoosss (approved):

PR #1359, opened 2026-07-29
fix(pool): align event payloads with canonical state

junghoon-vans:
## Summary
- remove the unauthenticated `payer` attribute and obsolete `payer` parameter from pool `Swap`
- emit canonical `CreatePool` token paths and `sqrtPriceX96` from normalized pool state
- add regression coverage for reversed-token `CreatePool` events and update affected integration fixtures

## Tests
- `GIT_MASTER=1 git diff --check origin/main...HEAD`
- `make test WORKDIR=tmp PKG=gno.land/r/gnoswap/pool/v1`
- `COMPOSE_PROJECT_NAME=gnoswap_event_pr make integration-test-run TEST=pool_swap_wugnot_gns_tokens`
- `COMPOSE_PROJECT_NAME=gnoswap_event_pr_fix make integration-test-run TEST=pool_create_pool_and_mint`
- QA also passed `router_exact_out_swap_route_with_gnot` for Swap event fixture coverage

## Notes
- Pool `Swap` no longer accepts `payer`; direct callers now provide only the swap callback after `sqrtPriceLimitX96`.
- Router callback `payer` remains as internal settlement context for choosing router-balance transfer vs user `TransferFrom`.
- Downstream event consumers should remove `payer` handling rather than substituting `prevAddr` or `recipient` as payer.

## Summary by CodeRabbit

* **Breaking Changes**
  * Simplified pool swap calls by removing the separate payer argument.
  * Swap events no longer include payer information.

* **Bug Fixes**
  * Corrected pool error-code mappings and related failure messages.
  * Pool creation events now report finalized token ordering and pricing.
  * Updated swap and routing behavior to preserve payer handling through callbacks.

* **Tests**
  * Expanded coverage for reversed token ordering and canonical pool-creation events.
  * Updated integration and fuzz tests for revised swap signatures and event output.


contract/r/gnoswap/pool/v1/swap.gno
jinoosss: It looks like it would be better to delete it.

jinoosss (approved):

PR #1366, opened 2026-08-01
fix(pool): seed tick oracle outside accumulators

junghoon-vans:
## Summary

`tickUpdate` left oracle outside accumulators uninitialized for ticks at or below `tickCurrent`, so the first crossing of such a tick produced corrupted oracle state (uninitialized `secondsPerLiquidityOutsideX128` / `tickCumulativeOutside` / `secondsOutside`).

When a tick is first initialized and is **active** (`tick <= tickCurrent`), `tickUpdate` now seeds its outside accumulators from the current oracle values:

- `secondsPerLiquidityOutsideX128` ← `secondsPerLiquidityCumulativeX128`
- `tickCumulativeOutside` ← `tickCumulative`
- `secondsOutside` ← `blockTimestamp`

`updatePosition` reads these via `observeSingle` (existing oracle logic) and passes them into `tickUpdate`, so the seeding matches the pool's current observation state exactly.

## Uniswap v3 alignment

This fix mirrors Uniswap v3's flow. Reference permalinks:

- [`UniswapV3Pool._updatePosition`](https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/UniswapV3Pool.sol#L379-L440) — calls `observeSingle(time, 0, slot0.tick, slot0.observationIndex, liquidity, slot0.observationCardinality)` before updating ticks, then passes the observed values into `ticks.update`.
- [`Tick.update`](https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/Tick.sol#L132-L140) — when the tick flips from uninitialized and `tick <= tickCurrent`, seeds `secondsPerLiquidityOutsideX128`, `tickCumulativeOutside`, and `secondsOutside` from those arguments.

Alignments made to match the reference:

- `tickUpdate` signature now takes `secondsPerLiquidityCumulativeX128 *u256.Uint`, `tickCumulative int64`, `blockTimestamp int64` as **flat required parameters** — identical in ordering to `Tick.update` (dropped the temporary `tickUpdateOracleState` struct / variadic fallback, so omitting the seeding args is a compile error rather than a silent regression).
- Seeding block (`liquidityGrossBefore == 0 && tick <= tickCurrent`) mirrors `Tick.update`'s flip-to-initialized branch.
- `observeSingle` is invoked with the same argument positions as `_updatePosition` (`time, 0, tick, observationIndex, liquidity, cardinality`).

## Test plan

- [x] `make test PKG=gno.land/r/gnoswap/pool/v1` — full unit suite passes.
- [x] New regression test `TestTickUpdateSeedsOracleOutsideForActiveTick` verifies seeding at initialization and correct deltas after a crossing.
- [x] Existing `tickUpdate` call sites updated to the new required signature (17 test call sites + `updatePosition`).

### DIFF
--- a/contract/r/gnoswap/pool/v1/position.gno
+++ b/contract/r/gnoswap/pool/v1/position.gno
@@ -316,6 +316,23 @@ func updatePosition(p *pl.Pool, positionParams ModifyPositionParams, tick int32)
 
 	var flippedLower, flippedUpper bool
 	if !liquidityDelta.IsZero() {
+		blockTimestamp := time.Now().Unix()
+		observationState := p.ObservationState()
+		tickCumulative, secondsPerLiquidityStr, err := observeSingle(
+			observationState,
+			blockTimestamp,
+			0,
+			tick,
+			observationState.Index(),
+			p.Liquidity(),
+			observationState.Cardinality(),
+		)
+		if err != nil {
+			return pl.NewDefaultPositionInfo(), err
+		}
+
+		secondsPerLiquidityCumulativeX128 := u256.MustFromDecimal(secondsPerLiquidityStr)
+
 		flippedLower = tickUpdate(
 			p,
 			positionParams.tickLower,
@@ -323,6 +340,9 @@ func updatePosition(p *pl.Pool, positionParams ModifyPositionParams, tick int32)
 			liquidityDelta,
 			feeGrowthGlobal0X128,
 			feeGrowthGlobal1X128,
+			secondsPerLiquidityCumulativeX128,
+			tickCumulative,
+			blockTimestamp,
 			false,
 			calculateMaxLiquidityPerTick(p.TickSpacing()),
 		)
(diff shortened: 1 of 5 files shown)


contract/r/gnoswap/pool/v1/tick.gno
junghoon-vans, line 221: https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/Tick.sol#L132-L140

contract/r/gnoswap/pool/v1/position.gno
```diff
@@ -334,6 +354,9 @@ func updatePosition(p *pl.Pool, positionParams ModifyPositionParams, tick int32)
 			liquidityDelta,
 			feeGrowthGlobal0X128,
 			feeGrowthGlobal1X128,
+			secondsPerLiquidityCumulativeX128,
+			tickCumulative,
+			blockTimestamp,
 			true,
 			calculateMaxLiquidityPerTick(p.TickSpacing()),
 		)
```
junghoon-vans, line 339: https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/UniswapV3Pool.sol#L379-L440

jinoosss (approved):

PR #1367, opened 2026-08-02
test: pay tokens via native transfers instead of the grc20 registry hub

moul:
## What

Across `pool`, `position`, `router`, `launchpad`, `gov/staker`, `protocol_fee`, the fuzz suite and the position scenario filetests, test callbacks and helpers moved tokens by calling the generic grc20 registry helper (`common.SafeGRC20Transfer` / `common.SafeGRC20Approve`) while an EOA/admin realm was set:

```go
testing.SetRealm(adminRealm)
common.SafeGRC20Transfer(cross(cur), token0Path, poolAddr, amount0Delta)
```

Production never funds the pool that way. Real swap callbacks move funds via the **token's own realm** (the token owner authorizes the transfer) or via **allowance** (`TransferFrom`), and reward/refund flows enter as the **module's own realm**. This aligns the tests with production:

- swap-callback mocks and approvals now dispatch to each token realm's own `Transfer`/`Approve` (e.g. `bar.Transfer(cross(cur), ...)`, `foo.Approve(cross(cur), ...)`);
- module-owned transfer tests (staker reward, launchpad refund) now drive the flow as the module realm, exactly as the real `CollectReward` / `TransferLeftFromProjectByAdmin` entry points do.

## Why

Test mocks that pay through a generic registry hop as an EOA don't reflect how tokens actually move on-chain, and they quietly assume the registry hub can debit an arbitrary caller. Paying via the token's native transfer keeps the mocks faithful to production and to grc20's realm-scoped teller semantics.

## Scope

40 `*_test.gno` / `*_filetest.gno` files. **No product code and no assertions changed** — behavior and expected output are identical. Each touched package's test suite passes.

## Summary by CodeRabbit

* **Tests**
  * Updated token transfer and approval test setup to execute from the correct contract context.
  * Improved token-specific handling across swap, pool, position, router, launchpad, staking, protocol-fee, and fuzz tests.
  * Added validation for unsupported token paths during test transfers and approvals.
  * Preserved existing coverage for refunds, rewards, swaps, liquidity changes, fee collection, and fee distribution.

### DIFF
--- a/contract/r/gnoswap/gov/staker/v1/staker_reward_test.gno
+++ b/contract/r/gnoswap/gov/staker/v1/staker_reward_test.gno
@@ -535,7 +535,7 @@ func TestStakerReward_TokenTransferScenarios(cur realm, t *testing.T) {
 		{
 			name:        "success - normal token transfer",
 			tokenPath:   GNS_PATH,
-			from:        admin,
+			from:        govStakerAddr,
 			to:          testutils.TestAddress("to"),
 			amount:      100_000_000,
 			fromBalance: 500_000_000,
@@ -544,26 +544,28 @@ func TestStakerReward_TokenTransferScenarios(cur realm, t *testing.T) {
 		{
 			name:        "success - exact balance transfer",
 			tokenPath:   GNS_PATH,
-			from:        admin,
+			from:        govStakerAddr,
 			to:          testutils.TestAddress("to"),
-			amount:      100_000_000,
-			fromBalance: 100_000_000,
+			amount:      1_000_000_000,
+			fromBalance: 1_000_000_000,
 			shouldFail:  false,
 		},
 		{
 			name:          "fail - insufficient balance",
 			tokenPath:     GNS_PATH,
-			from:          testutils.TestAddress("from"), // not have balance
+			from:          govStakerAddr,
 			to:            testutils.TestAddress("to"),
-			amount:        200_000_000,
-			fromBalance:   100_000_000,
+			// setup re-runs per subtest, so the funded balance accumulates; pick an
+			// amount that exceeds it under any ordering.
+			amount:        1_000_000_000_000_000,
+			fromBalance:   1_000_000_000,
 			expectedError: "insufficient balance",
 			shouldFail:    true,
 		},
 		{
 			name:          "fail - invalid to address",
 			tokenPath:     GNS_PATH,
-			from:          admin,
+			from:          govStakerAddr,
 			to:            address(""),
 			amount:        100_000_000,
 			fromBalance:   500_000_000,
@@ -573,7 +575,7 @@ func TestStakerReward_TokenTransferScenarios(cur realm, t *testing.T) {
 		{
 			name:          "fail - negative amount",
 			tokenPath:     GNS_PATH,
-			from:          admin,
+			from:          govStakerAddr,
 			to:            testutils.TestAddress("to"),
 			amount:        -100_000_000,
 			fromBalance:   500_000_000,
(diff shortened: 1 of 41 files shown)


contract/r/gnoswap/gov/staker/v1/staker_reward_test.gno
```diff
@@ -615,7 +618,11 @@ func TestStakerReward_TokenTransferScenarios(cur realm, t *testing.T) {
 			cleanupStakerRewardTest(cur, t)
 			setupStakerRewardStakerTokenBalance(cur, t)
 
-			testing.SetRealm(adminRealm)
+			// Drive the flow as the real on-chain entry (CollectReward) does:
+			// transferToken debits the caller realm (Previous), whose tokens
+			// were minted to govStakerAddr above. Previous must be the staker
+			// code realm, not an EOA/origin.
+			testing.SetRealm(testing.NewCodeRealm("gno.land/r/gnoswap/gov/staker"))
```
junghoon-vans, line 627: **Please verify the debit source used by these scenarios.**

After changing the test realm to `gno.land/r/gnoswap/gov/staker`, `setupStakerRewardStakerTokenBalance()` still funds `govStakerAddr`, but several cases in this table keep `from` as `admin` (and `fromBalance` is not used to set up or assert the balance). This means the exact-balance and insufficient-balance cases may not exercise the actual staker debit source. Could you either set `from` to `govStakerAddr` or explicitly fund and assert the intended source?

moul:
Re the `staker_reward_test.gno` note on the transfer scenarios: agreed, and it is worth fixing rather than deferring.

The table rows set `from: admin` while `setupStakerRewardStakerTokenBalance()` funds `govStakerAddr`, and `transferToken` checks `common.BalanceOf(tokenPath, from)`. So the exact-balance and insufficient-balance cases assert against an address the setup never funds — they pass for the wrong reason, and would keep passing if the debit source changed underneath them. That is the same class of problem this PR is trying to remove elsewhere: a test whose setup does not match the path production takes.

Will set `from` to `govStakerAddr` so the cases exercise the real debit source, and keep `fromBalance` meaningful by funding and asserting it explicitly.

moul:
Pushed as `8effa2c`.

Every row that used `from: admin` now uses `govStakerAddr` — the address `setupStakerRewardStakerTokenBalance()` actually funds and the one the transfer actually debits. Two rows needed more than a rename:

- **exact balance**: was `amount: 100_000_000` against a `fromBalance` label of the same value, while the funded balance is `1_000_000_000`. Now transfers the full funded amount, so "exact" means exact.
- **insufficient balance**: was pointed at an unfunded throwaway address, so it exercised the guard rather than the real source. Now points at `govStakerAddr` with an amount above any funded balance. Note the setup re-runs per subtest so the balance accumulates across rows — the amount is chosen to exceed it under any ordering, rather than assuming a particular one.

Full `gov/staker/v1` package passes.

Worth flagging separately, since it is product code and outside this PR's scope: `transferToken` checks `common.BalanceOf(tokenPath, from)` but then debits via `common.SafeGRC20Transfer(cross(rlm), ...)`, which takes the funds from the **calling realm**, not from `from`. The guard and the debit are on two different addresses, so `from` is effectively an unvalidated label. The tests were papering over that; they now at least pass the same address for both. Whether `transferToken` should take a `from` parameter at all is a separate question for the staker owners.

moul:
Context for this PR, now that the wider shape is visible: #1379 and #1380 are the two candidate adaptations to an upstream `grc20` accessor change, and this PR is the test-side groundwork for both — mocks that stop paying through the generic `common` hub and use each token's own realm instead. It stands on its own regardless of which shape lands upstream, and it is not gated on either.

PR #1371, opened 2026-08-05
feat: distribute protocol fee rewards per token

junghoon-vans:
## Summary
- `protocol_fee`: add `DistributeProtocolFeeByTokenPath` to flush reserved fees for a single token, so per-token reward distribution no longer requires flushing all reserved tokens at once
- `gov/staker`: track protocol fee rewards per token path
  - manager: `updateAccumulatedProtocolFeeX128PerStakeForToken`, `SetAccumulatedProtocolFeeX128PerStakeForToken`, `SetProtocolFeeAmountForToken`
  - user state: `updateRewardDebtX128ForToken`, `claimRewardForToken`, per-token reward debt / accumulated / claimed setters
  - entry point: `claimRewardProtocolFeeRewardByTokenPath` (single-token claim without bulk timestamp gate)
  - harden proxy-layer setters with copy-on-write (`u256.Zero().Set` deep copy + map rebuild) to prevent shared mutable state / aliasing across structs
- `launchpad`: forward split protocol fee reward claims per token (`CollectProtocolFeeRewardFromLaunchPad` reuses the launchpad reward ID used by add/remove stake)
- tests: unit tests for per-token accumulation/claim math (incl. same-timestamp independent tokens, bulk/per-token interplay, no-amount preservation) and scenario filetests for multi-token reward delegation and launchpad fee forwarding

## Tests
- `cd tmp/gno/examples`
- `gno test -v ./gno.land/r/gnoswap/protocol_fee/v1`
- `gno test -v ./gno.land/r/gnoswap/gov/staker/v1`
- `gno test -v ./gno.land/r/gnoswap/launchpad/v1`
- `gno test -v ./gno.land/r/gnoswap/scenario/gov/staker`
- `gno test -v ./gno.land/r/gnoswap/scenario/launchpad`
- all unit + scenario tests pass (scenario: 26/26 gov/staker + 26/26 launchpad, 0 failures)

## Notes
- Per-token claim intentionally does not update the manager/user `accumulatedTimestamp`; subsequent bulk updates recompute delta for already-claimed tokens as 0, so no double claim.
- Branch is rebased onto the latest `main` (`8e1d2f8e`), 0 commits behind.

## Summary by CodeRabbit

* **New Features**
  * Added token-specific protocol-fee distribution, reward tracking, claiming, and transfers.
  * Enabled individual emission and Launchpad reward collection with authorization and project-wallet validation.
  * Added reserved-token management and support for withdrawal halt states.
* **Bug Fixes**
  * Improved reward calculations at identical timestamps and prevented duplicate claims.
  * Safely handle missing, zero-value, and non-distributed reward data.
  * Preserved unrelated token reservations and reward balances during updates.
* **Tests**
  * Added coverage for token isolation, authorization, halted withdrawals, no-staker periods, and reserved-token management.


contract/r/gnoswap/gov/staker/protocol_fee_reward_manager.gno
```diff
@@ -95,21 +95,32 @@ func (p *ProtocolFeeRewardManager) SetAccumulatedProtocolFeeX128PerStake(accumul
...
 func (p *ProtocolFeeRewardManager) SetProtocolFeeAmountForToken(token string, amount int64) {
-	if p.protocolFeeAmounts == nil {
-		p.protocolFeeAmounts = make(map[string]int64)
+	copied := make(map[string]int64, len(p.protocolFeeAmounts)+1)
+	for existingToken, existingAmount := range p.protocolFeeAmounts {
+		copied[existingToken] = existingAmount
 	}
-	p.protocolFeeAmounts[token] = amount
+	copied[token] = amount
+	p.protocolFeeAmounts = copied
```
jinoosss: It looks like we won't need to do a full scan.

contract/r/gnoswap/gov/staker/protocol_fee_reward_state.gno
```diff
@@ -116,24 +116,30 @@ func (p *ProtocolFeeRewardState) GetClaimedRewardForToken(token string) int64 {
...
 func (p *ProtocolFeeRewardState) SetClaimedRewardForToken(token string, value int64) {
-	if p.claimedRewards == nil {
-		p.claimedRewards = make(map[string]int64)
+	copied := make(map[string]int64, len(p.claimedRewards)+1)
+	for existingToken, existingValue := range p.claimedRewards {
+		copied[existingToken] = existingValue
 	}
-	p.claimedRewards[token] = value
+	copied[token] = value
+	p.claimedRewards = copied
```
jinoosss: It looks like we won't need to do a full scan.

junghoon-vans:
### Follow-up fixes for per-token protocol-fee rewards

This update addresses the three follow-up issues identified in the per-token protocol-fee reward flow:

1. **Same-timestamp reward settlement** (`729133c8`)
   The settlement guards now use a strict `>` timestamp comparison instead of `>=`. A stake change at the same timestamp as the latest accumulation now settles the existing stake against the current accumulator and refreshes reward debt instead of bypassing the update. A regression test covers a stake increase after same-timestamp accumulation.

2. **Zero-distribution token claims** (`25eb632d`)
   After protocol-fee distribution runs, a zero cumulative actual distributed amount returns before manager/user reward-state persistence and accumulation-event emission. This avoids materializing per-token reward state when no distribution occurred. A regression test verifies that the relevant state maps remain empty.

3. **In-place per-token reward updates** (`5bc415ff`)
   Per-token setters now lazily initialize their maps and update only the target entry, rather than copying the entire map on every write. `u256` values continue to be cloned before storage, and the bulk setters retain their existing copy-all behavior.

Validated with the targeted root, v1, and scenario staker test suites.

contract/r/gnoswap/gov/staker/protocol_fee_reward_state.gno
junghoon-vans: [5bc415f](https://github.com/gnoswap-labs/gnoswap/pull/1371/commits/5bc415ff17f53a9be97be9b6d4d22bd56eb0a6b0)

contract/r/gnoswap/gov/staker/protocol_fee_reward_manager.gno
junghoon-vans: [5bc415f](https://github.com/gnoswap-labs/gnoswap/pull/1371/commits/5bc415ff17f53a9be97be9b6d4d22bd56eb0a6b0)

contract/r/gnoswap/gov/staker/v1/staker_reward.gno
```diff
@@ -95,12 +95,70 @@ func (gs *govStakerV1) CollectReward(_ int, rlm realm) {
...
 }
 
 // CollectProtocolFeeReward collects accumulated protocol fee rewards for the provided token path.
 func (gs *govStakerV1) CollectProtocolFeeReward(_ int, rlm realm, tokenPath string) {
-	assertNotImplementYet()
+	if !rlm.IsCurrent() {
+		panic(errors.New(errSpoofedRealm))
+	}
+
+	halt.AssertIsNotHaltedWithdraw()
```
jinoosss, line 134: After verifying whether it is actually a protocol fee token, it seems best to either request an early return or panic.
If the token is registered on grc20reg, it may generate unnecessary data.

contract/r/gnoswap/protocol_fee/v1/protocol_fee.gno
```diff
@@ -84,9 +84,63 @@ func (pf *protocolFeeV1) DistributeProtocolFee(_ int, rlm realm) {
...
+
+	if err := pfs.removeReservedToken(0, rlm, tokenPath); err != nil {
+		panic(err)
+	}
+
+	chain.Emit(
+		"TransferProtocolFee",
+		"prevAddr", prev.Address().String(),
+		"prevRealm", prev.PkgPath(),
+	)
```
jinoosss, line 129: It would be good if it were emitted by an explicit event.

contract/r/gnoswap/protocol_fee/v1/protocol_fee.gno
```diff
@@ -84,9 +84,63 @@ func (pf *protocolFeeV1) DistributeProtocolFee(_ int, rlm realm) {
 	)
 }
 
-// DistributeProtocolFeeByTokenPath is retained for ABI compatibility while token-path distribution is deferred.
+// DistributeProtocolFeeByTokenPath distributes collected protocol fees for one token path.
 func (pf *protocolFeeV1) DistributeProtocolFeeByTokenPath(_ int, rlm realm, tokenPath string) {
-	assertNotImplementYet()
+	if !rlm.IsCurrent() {
```
jinoosss, line 54: It seems best to create a separate internal implementation for this function to align its logic with that of the existing `DistributeProtocolFee` function.

jinoosss (commented):
Overall, it’s well implemented.
Please take a look at these few comments.

jinoosss (approved):

contract/r/gnoswap/protocol_fee/v1/protocol_fee.gno
junghoon-vans: [b07a59c](https://github.com/gnoswap-labs/gnoswap/pull/1371/commits/b07a59cbe6a0df6ce88a859c53dde5b896d830b2)

PR #1372, opened 2026-08-05
perf(staker): scope external incentive discovery per pool

junghoon-vans:
## Summary
Fix external incentive discovery in the staker: it previously scanned a **global** creation-time index, so staking collection cost grew with the total number of incentives system-wide.

- `3b16654e` — Remove the global creation-time index; scope discovery to each pool's own incentive tree (per-pool iteration).
- `ac3a4de3` — Add a per-pool creation-time index (`Incentives.byCreationTime`) with `AddIncentiveByCreationTime` / `IterateIncentiveIdsByTime`, register incentives in `create()`, and switch `getExternalIncentiveIdsBy` to a bounded range lookup. Discovery cost now scales with the incentives of the queried pool only.
- `a04bf5d9` — Remove unused `bptree` import in `staker_test.gno`.
- `54f7272e` — Fix `TestGetExternalIncentiveIdsBy` seeding which constructed incentives with a zero-length duration (start=end=0), causing a `u256.MulDiv` "denominator must be greater than 0" panic before exercising `getExternalIncentiveIdsBy`.

`IterateIncentiveIdsByTime` uses `ReverseIterate` because it is inclusive on both ends, matching the previous `createdTimestamp >= startTime && createdTimestamp <= endTime` discovery semantics.

## Tests
- `make test PKG=gno.land/r/gnoswap/staker/v1` — full suite passes (304s)
- `make test PKG=gno.land/r/gnoswap/staker/v1 RUN=TestGetExternalIncentiveIdsBy` — passes
- `make test PKG=gno.land/r/gnoswap/staker/v1 RUN=TestCreateExternalIncentive_SameTimestamp` — passes

## Notes
- The `TestGetExternalIncentiveIdsBy` panic was verified to be independent of the index change (reproduced at `3b16654e` before the index was introduced); it was a test-seeding bug, not a contract bug.
- Production path `CreateExternalIncentive` already guards against zero duration via `assertIsValidIncentiveDuration`.


contract/r/gnoswap/staker/pool.gno
```diff
@@ -314,11 +323,46 @@ func (i *Incentives) IterateIncentives(fn func(incentiveId string, incentive *Ex
 	})
 }
 
+// AddIncentiveByStartTime registers an incentive ID under its start timestamp
+// in the per-pool start-time index. Multiple incentives starting at the same
+// timestamp are accumulated as a list.
+func (i *Incentives) AddIncentiveByStartTime(startTimestamp int64, incentiveId string) {
+	var incentiveIds []string
```
junghoon-vans, line 330: make

junghoon-vans: @jinoosss Explicit initialization is unnecessary here. var incentiveIds []string creates a nil slice, and append works with nil slices by allocating storage as needed. Using make([]string, 0) would be redundant for this use case. make would only be useful if we wanted to preallocate capacity or guarantee a non-nil slice.

jinoosss (approved):

PR #1373, opened 2026-08-05
perf(staker): bound external incentive refund scanning

junghoon-vans:
## Summary

`EndExternalIncentive` refund execution iterated the pool's **unbounded** `unclaimablePeriods` tree. A pool that had accumulated many zero-liquidity windows over its lifetime made refunds unexecutable — gas and iterations grow without bound as the tree grows, so `EndExternalIncentive` could eventually exceed the block gas limit.

This PR replaces the tree scan with an **O(1) per-incentive accumulator**:

- `ExternalIncentive` gains a `unclaimableSeconds` field, updated whenever an unclaimable period closes (`endUnclaimablePeriod`) for every non-refunded incentive whose window overlaps it.
- `calculateUnclaimableReward` reads the accumulator instead of scanning the tree. Only ongoing periods (`end == 0`) are resolved at refund time via a **bounded tail scan** — the highest period before the incentive window and the highest period within it, each visited at most once.
- Refund gas is now constant regardless of how many unclaimable periods the pool has accumulated.

## Scenario that triggered the issue

- LP positions are repeatedly staked/unstaked during an incentive window, leaving multiple disjoint zero-liquidity (unclaimable) gaps.
- Over time a pool accumulates many such periods. Any incentive creator's `EndExternalIncentive` had to scan **all** of them, so refunds could become unexecutable.
- After the fix, the refund covers the **union of all unclaimable gaps** overlapping the incentive, computed in O(1), and the accounting invariant `Total == Remaining + Distributed + Penalty` is preserved.

## Changes

| File | Change |
|------|--------|
| `contract/r/gnoswap/staker/pool.gno` | Add `unclaimableSeconds` field + getter/setter/clone/init |
| `contract/r/gnoswap/staker/v1/reward_calculation_incentives.gno` | Accumulate closed periods into every active incentive; read accumulator at refund time with bounded ongoing-period tail scans |
| `contract/r/gnoswap/staker/v1/reward_calculation_incentives_test.gno` | Unit tests for accumulator updates on `endUnclaimablePeriod`, incl. refunded-incentive exclusion |
| `contract/r/scenario/staker/external_incentive_refund_with_multiple_unclaimable_periods_filetest.gno` | Scenario regression test: 3 stake/unstake cycles → 18 days of disjoint gaps → refund = full union, invariant intact, refund tokens actually transferred, staked position still collects |

## Test

All scenario filetests pass (73 filetests in `scenario/staker`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reward calculations by accurately accumulating time from completed unclaimable periods.
  * Prevented refunded incentives from being included in reward calculations.
  * Avoided double-counting rewards across multiple unclaimable periods.
* **Tests**
  * Added coverage for accumulated unclaimable time across closed and multiple periods.

### DIFF
--- a/contract/r/gnoswap/staker/pool.gno
+++ b/contract/r/gnoswap/staker/pool.gno
@@ -344,6 +344,8 @@ type ExternalIncentive struct {
 	creator                  address    // creator address
 
 	refunded bool // whether incentive has been refunded (includes GNS deposit and unclaimed rewards)
+
+	unclaimableSeconds int64 // accumulated seconds of unclaimable periods overlapping the incentive window
 }
 
 // ExternalIncentive Getter/Setter methods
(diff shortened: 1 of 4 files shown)


junghoon-vans:
Gas Metric: https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/31

jinoosss (approved):

PR #1376, opened 2026-08-06
perf(staker): reduce append-only storage growth from redundant writes

jinoosss:
## Summary

Two independent write-reduction fixes for the staker realm's append-only trees, which the gas report identified as growing without bound (no prune path) and driving the gradual increase in gas usage over time (issue 3).

### 1. Skip redundant staked-liquidity tree writes
`modifyDeposit` appended a `StakedLiquidity` entry on every call. It now writes only when the value actually changes. A zero-net-delta update (e.g. a balanced tick cross) is a no-op for readers, which resolve the value via `CurrentStakedLiquidity` (latest entry ≤ t) with no dependency on the entry's timestamp.

The global reward ratio accumulation write is deliberately **not** guarded the same way: each of its entries doubles as a *time checkpoint* that `calculateGlobalRewardRatioAccumulation` uses to derive `timeDiff`. Suppressing a same-valued write there would fail to advance the checkpoint and make a later non-zero-liquidity interval over-accumulate across the skipped zero-liquidity span.

### 2. Stop redundant global-accumulation writes in `cacheInternalReward`
`cacheInternalReward` wrote a global reward ratio accumulation checkpoint on every call. Because it runs on every position reward calculation (via `cacheRewardForPool`) and every emission distribution across all tiered pools, it appended an entry on nearly every staker interaction — a dominant driver of the unbounded growth.

The write is redundant: the global accumulation is a function of staked liquidity only (emission is applied later, per-warmup), and staked liquidity changes exclusively through `modifyDeposit`, which already checkpoints the accumulation (with pre-change liquidity) at every change. Between changes the accumulation for any timestamp is derived on demand as `lastCheckpoint + (t - lastCheckpointTime) * q128 / liquidity`.

## Reward rounding behavior (precision change)

Removing the per-call global-accumulation write from the reward path changes the rounding of collected rewards by small amounts (observed ±1, up to a few units in cumulative sums). This is a precision change, not an accounting error. Details below.

### What changed numerically

Scenario filetests under `gno.land/r/gnoswap/scenario/staker` show `CollectReward` outputs shifting slightly upward, with the matching refund shifting down, e.g.:

```
creator receives all remaining reward token: 1512000003 -> 1512000002   (+1 distributed to stakers)
frequent collections total (10 blocks):      40132690    -> 40132693     (closer to the single-collect 40132695)
```

The direction is consistent: slightly **more** is distributed and **less** dust is left behind.

### Why

The global reward ratio accumulation is built as a sum of floored terms in `calculateGlobalRewardRatioAccumulation`:

```
acc(t) = oldAcc + MulDiv(timeDiff, q128, stakedLiquidity)   // MulDiv floors
```

Previously an accumulation checkpoint was written on nearly every interaction (every position reward calculation / emission distribution), so the interval was split into many small pieces and `MulDiv` floored many times. The accumulation is now checkpointed only when staked liquidity actually changes (in `modifyDeposit`), so the same span is covered by fewer, larger `MulDiv` steps and is floored fewer times. The value is therefore slightly higher and closer to the mathematical value.

Invariant that holds in both cases: every term is floored, so `acc <= true value` regardless of checkpoint density. This change only reduces the undercount; it never makes `acc` exceed the true value.

### Conservation across multiple positions

For in-range positions the per-position inside-accumulation deltas telescope:

```
Σ (insideAccDelta_i · liquidity_i) = globalAccDelta · stakedLiquidity
```

so, with `rewardPerWarmup` computing `floor(insideAccDelta · liquidity · rewardPerSecond / q128)` per position:

```
Σ reward ≤ (rewardPerSecond / q128) · globalAccDelta · stakedLiquidity
         ≤ rewardPerSecond · Δt = emission
```

because `globalAccDelta ≤ true`. The total distributed can therefore never exceed the emission (internal) or the incentive budget (external), so no realm balance can be drained. This is confirmed by `end_external_incentive_reward_amount_invariant_filetest`, which passes.

### Ledger safety

- **Internal (emission):** staker balance = minted − collected ≥ 0, since collected ≤ emission = minted.
- **External incentives:** `CollectReward` already guards each transfer with the incentive's remaining `RewardAmount`, and the end-of-incentive refund stays ≥ 0 (`1512000002 > 0`).
- **Collection-order independence:** the accumulation now depends only on liquidity-changing events (`modifyDeposit`), not on when/who collects. Previously each collect wrote a checkpoint, so the accumulation — and its rounding — depended on collection timing/order. Rewards are now deterministic with respect to liquidity events alone.

### DIFF
--- a/contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
+++ b/contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
@@ -236,6 +236,12 @@ func (self *PoolResolver) calculateGlobalRewardRatioAccumulation(currentTime int
 }
 
 // updateGlobalRewardRatioAccumulation updates the global reward ratio accumulation and returns the new accumulation.
+//
+// NOTE: the write below cannot be skipped even when the accumulation value is unchanged (e.g. zero staked
+// liquidity or zero elapsed time). Each entry doubles as a *time checkpoint*: calculateGlobalRewardRatioAccumulation
+// derives timeDiff from the latest entry's timestamp. Omitting a same-valued write would fail to advance that
+// checkpoint, so a later non-zero-liquidity interval would over-accumulate across the skipped (zero-liquidity)
+// span. Bounding this append-only tree must therefore be done via watermark pruning, not by suppressing writes.
 func (self *PoolResolver) updateGlobalRewardRatioAccumulation(currentTime int64, currentStakedLiquidity *u256.Uint) *u256.Uint {
 	newAcc := self.calculateGlobalRewardRatioAccumulation(currentTime, currentStakedLiquidity)
 
@@ -546,7 +552,14 @@ func (self *PoolResolver) modifyDeposit(delta *i256.Int, currentTime int64, next
 		}
 	}
 
-	self.Pool.SetStakedLiquidityAt(currentTime, deltaApplied)
+	// Only append a staked-liquidity entry when the value actually changes (e.g. a tick cross whose
+	// net delta is zero leaves it unchanged). Unlike the global reward ratio accumulation, this tree
+	// carries no time-checkpoint semantics: it is read purely as a point-in-time value via
+	// CurrentStakedLiquidity (latest entry <= t), so omitting a duplicate-valued entry preserves
+	// behavior while keeping this append-only tree from growing on no-op updates.
+	if !lastStakedLiquidity.Eq(deltaApplied) {
+		self.Pool.SetStakedLiquidityAt(currentTime, deltaApplied)
+	}
 
 	return result
 }
--- a/contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
+++ b/contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
@@ -204,12 +204,23 @@ func (self *PoolResolver) cacheReward(currentTime int64, currentTierReward int64
 	}
 }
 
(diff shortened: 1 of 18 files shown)


contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
```diff
@@ -204,12 +204,23 @@ func (self *PoolResolver) cacheReward(currentTime int64, currentTierReward int64
...
+// of the staker realm's unbounded storage growth (gas report issue 3).
+//
+// INVARIANT: if a future code path mutates staked liquidity outside modifyDeposit, it MUST checkpoint the
+// accumulation there (as modifyDeposit does) or on-demand reads will over-accumulate across the change.
 func (self *PoolResolver) cacheInternalReward(currentTime int64, currentEmission int64) {
 	self.cacheReward(currentTime, currentEmission)
-
-	currentStakedLiquidity := self.CurrentStakedLiquidity(currentTime)
-	self.updateGlobalRewardRatioAccumulation(currentTime, currentStakedLiquidity)
 }
```
junghoon-vans: Please avoid redefining the same function.

jinoosss: e648194428c0d89f661f72b32b2d3c3c3e051f72

junghoon-vans (approved):

PR #1377, opened 2026-08-06
perf(gov/staker): avoid protocol-fee work for tokens with no new fees

jinoosss:
## Summary

Addresses the gas report's issue 1: per-transaction protocol-fee processing cost scaled with the number of *registered* tokens rather than the number that actually received fees ("even zero-reward tokens are fully iterated"). This change stops doing work for tokens whose newly distributed amount is zero.

## Changes

- `updateAccumulatedProtocolFeeX128PerStake` persists only the tokens whose accumulator/amount actually changed, instead of deep-copying and re-persisting the whole per-token maps on every add/remove/claim.
- `calculateClaimableRewards` skips the 256-bit `MulDiv` for tokens whose per-stake accumulator is unchanged since the staker's reward debt (zero newly accrued): the reward is exactly zero, so only the expensive computation is avoided while the result is preserved (the zero is still recorded).

Both are behavior-preserving: unchanged tokens keep identical values, registered protocol-fee tokens are monotonic (never removed), and `distributedAmounts` always covers the full set.

## Testing

- New table-driven tests (incl. error handling): `TestUpdateAccumulatedProtocolFee`, `TestUpdateAccumulatedProtocolFee_UnchangedTokenNotRewritten`, `TestUpdateAccumulatedProtocolFee_PastTimestampNoOp`, `TestProtocolFeeStake_AmountErrors`, `TestProtocolFeeReward_ClaimSkipsZeroDeltaTokens`.
- The full `gov/staker/v1` suite passes.

### DIFF
--- a/contract/r/gnoswap/gov/staker/v1/protocol_fee_reward_accumulation_test.gno
+++ b/contract/r/gnoswap/gov/staker/v1/protocol_fee_reward_accumulation_test.gno
@@ -1,0 +1,231 @@
+package staker
+
+import (
+	"testing"
+
+	uassert "gno.land/p/nt/uassert/v0"
+
+	"gno.land/r/gnoswap/gov/staker"
+)
+
+// updateAccumulatedProtocolFeeX128PerStake must persist only the tokens whose value actually changed,
+// while always advancing the accumulated timestamp (gas report issue 1 / §1.2).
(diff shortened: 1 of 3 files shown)


jinoosss:
depends on #1371

junghoon-vans (approved):

PR #1378, opened 2026-08-06
refactor(staker): separate read-only reward calculation from state update

jinoosss:
## Summary

The `Collectable*` view getters shared `calcPositionReward` with `CollectReward`, so querying a collectable amount mutated persisted state (reward-cache materialization, lazy pool creation, deposit incentive-index updates). This separates the pure calculation from the state update so views never write.

## Changes

- `calculatePositionReward` is pure and returns a `positionRewardUpdate` describing the would-be writes.
- `updatePositionReward` applies that update; only `CollectReward` calls it.
- `Collectable*` getters run the calculation and discard the update.

Unify the internal reward calculator: the per-second emission-rate schedule is resolved up-front by `resolveInternalRewardSegments` (persisted cache entries for the historical portion + on-the-fly halving splits for the uncached tail, using the current tier ratio/count), and a single `calculateInternalReward` consumes that schedule. This removes the reward-cache-materialization dependency from the calculation, so the calculation needs no writes. A schedule resolved from the emission halving schedule matches one resolved from a materialized cache bit-for-bit.

## Testing

- New table-driven tests (incl. edge cases): `TestResolveInternalRewardSegments`, `TestResolveInternalRewardSegments_EmissionMatchesMaterializedCache`, `TestCalculateInternalReward`.
- Validated against the canonical tick-cross emulator suite; the full `staker/v1` suite passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reward queries are now read-only and no longer modify cached rewards, pool data, deposit incentives, or external incentive tracking.
  * Reward calculations correctly handle emission-rate changes, halving boundaries, warmup periods, and multiple reward segments.
  * Reward collection continues to apply necessary updates and transfer calculated rewards correctly.

* **Tests**
  * Added coverage for reward calculations, cache immutability, historical equivalence, and read-only getter behavior.


junghoon-vans:
I added a test that uses MsgRun for verification. @jinoosss

junghoon-vans (approved):

PR #1381, opened 2026-08-07
fix(staker): emit reward ratio accumulation at event time, not last checkpoint

jinoosss:
## Summary

Follow-up to #1376. The staker's event payloads carry a global reward ratio accumulation that is no longer the accumulation *at the event timestamp*, which breaks off-chain reward accounting.

On-chain reward math is **not** affected — it always derives the accumulation on demand. Only event emission reads the raw stored checkpoint.

## The problem

Four emit sites read the accumulator through `CurrentGlobalRewardRatioAccumulation`, which returns the latest stored checkpoint (`<= currentTime`), not the accumulation at `currentTime`:

| Site | Event | `modifyDeposit` at the same timestamp? |
|---|---|---|
| `staker.gno:324` | `StakeToken` | only when the position is in range |
| `staker.gno:542` | `CollectReward` (external) | never |
| `staker.gno:665` | `CollectReward` (internal) | never |
| `staker.gno:755` | `UnStakeToken` | only when the position is in range |

Before #1376 this was accurate by accident: every one of these paths ran `cacheInternalReward` (via `PoolTier.cacheReward` / `cacheRewardForPool`) beforehand, which wrote a checkpoint at `currentTime`. #1376 removed that write — correctly, since it was redundant for reward calculation — so checkpoints are now written exclusively by `modifyDeposit`.

The emitted value therefore lags by:

```
(eventTime - lastCheckpointTime) * q128 / stakedLiquidity
```

`lastCheckpointTime` only advances on an in-range stake/unstake or a swap that crosses a tick with staked gross liquidity. For a pool with stable price and no staking activity the checkpoint can be hours or days old, and every `CollectReward` in that window emits the same frozen value while `blockTimestamp` keeps advancing.

Indexers consume the emitted accumulator as authoritative at the event timestamp and overwrite both the value and its timestamp. A lagging value makes the accumulation over `[lastCheckpointTime, eventTime]` unrecoverable, and because the accumulator is monotonically non-decreasing by construction, a regressing value can pin an indexer's accumulator in place.

## Fix

Emit the derived accumulation instead of the stored checkpoint:

```go
func (self *PoolResolver) globalRewardRatioAccumulationAt(currentTime int64) *u256.Uint {
	return self.calculateGlobalRewardRatioAccumulation(currentTime, self.CurrentStakedLiquidity(currentTime))
}
```

This is the exact derivation `CalculateRawRewardForPosition` already uses for reward calculation, so the emitted value now reproduces the on-chain reward math. It is read-only: no storage write is added and the gas reduction from #1376 is preserved.

`CurrentGlobalRewardRatioAccumulation` is unchanged and still used by `calculateGlobalRewardRatioAccumulation` to locate the checkpoint base.

## Tests

New `global_reward_ratio_accumulation_at_test.gno`:

- `TestGlobalRewardRatioAccumulationAt` — table-driven over the checkpoint/derivation boundary: constant liquidity past a stale checkpoint, query exactly on the checkpoint timestamp, a liquidity change rebasing the derivation, zero staked liquidity, and liquidity fully withdrawn (derivation freezes). Each case asserts the stored checkpoint and the derived value separately, so the two staying in sync would not hide a regression.
- `TestGlobalRewardRatioAccumulationAt_MatchesRewardCalculationBase` — the derivation agrees with the reward-calculation base across a range of elapsed times.

Note that no scenario filetest asserts `globalRewardRatioAccX128`, so the existing suite offered no coverage here.

```
make test PKG=gno.land/r/gnoswap/staker/v1              # ok (65.00s)
make test PKG=gno.land/r/gnoswap/scenario/staker        # ok (29.25s)
make test PKG=gno.land/r/gnoswap/scenario/upgrade       # ok (4.40s)
```

## Compatibility

Event schema is unchanged — only the value's freshness. Consumers that already treat the field as authoritative at the event timestamp become correct without any change on their side.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reward accumulator accuracy for staking, reward collection, and unstaking events.
  - Reward calculations now reflect the correct value at each event timestamp, including after liquidity changes, full withdrawals, and periods without liquidity.
  - Resolved discrepancies caused by stale checkpoint values.

- **Tests**
  - Added coverage for timestamp boundaries, stale checkpoints, liquidity changes, zero-liquidity periods, withdrawals, and reward calculation consistency.


junghoon-vans (commented):
The implementation direction looks correct, and the CI status is green. I have one non-blocking test-coverage suggestion below.

contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
```diff
@@ -236,6 +236,20 @@ func (self *PoolResolver) calculateGlobalRewardRatioAccumulation(currentTime int
...
 }
 
+// globalRewardRatioAccumulationAt returns the global reward ratio accumulation *at* currentTime.
+//
+// CurrentGlobalRewardRatioAccumulation returns the latest stored checkpoint (<= currentTime), which is
+// only equal to the accumulation at currentTime when a checkpoint was written at that very timestamp.
+// Checkpoints are written exclusively by modifyDeposit (staked liquidity changes), so on any other path
+// the stored value lags by (currentTime - lastCheckpointTime) * q128 / stakedLiquidity.
+//
+// Reward calculation never has this problem because it derives the accumulation on demand
```
junghoon-vans, line 246: Could we avoid doing the `CurrentStakedLiquidity` lookup inside this helper, or return/reuse the value at the call sites? Each of the four updated event paths immediately calls `CurrentStakedLiquidity(currentTime)` again to populate `stakedLiquidity` for the same event, so this traverses the staked-liquidity tree twice. That adds avoidable gas to `StakeToken`, both `CollectReward` emissions, and `UnStakeToken`; please compute it once and pass it into the derivation, or return both values from the helper.

junghoon-vans (commented):
The event-payload correctness fix looks good overall. I have one non-blocking performance concern in the inline comment below.

contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
jinoosss: Since the remaining comments have an O(1) time complexity for each lookup, I believe this represents a significant trade-off compared to maintaining consistency and refactoring the function.

In addition to the comments above, I have improved performance through the following changes:
I consolidated the duplicate lookups for each reward within the `CollectReward` function into a single lookup.
f9d60cf00e56589dbd1cbbd1514ce7e7b619813c

junghoon-vans: Thanks for consolidating the per-reward lookups in `CollectReward`; that is a good improvement.

One non-blocking consideration for a follow-up: `CurrentStakedLiquidity` uses a B+ tree predecessor lookup, so its complexity is `O(log N)` with respect to the number of checkpoints (it stops after the first match, so it is not `O(N)`, but it is not strictly `O(1)` either). The current helper still performs this lookup internally, and each call site reads it again.

If useful, please consider a snapshot-oriented interface so both values can reuse the same lookup:

```go
type GlobalRewardRatioSnapshot struct {
    Accumulation    *u256.Uint
    StakedLiquidity *u256.Uint
}

func (self *PoolResolver) globalRewardRatioSnapshotAt(currentTime int64) *GlobalRewardRatioSnapshot {
    stakedLiquidity := self.CurrentStakedLiquidity(currentTime)
    return &GlobalRewardRatioSnapshot{
        Accumulation:    self.calculateGlobalRewardRatioAccumulation(currentTime, stakedLiquidity),
        StakedLiquidity: stakedLiquidity,
    }
}
```

This is only a non-blocking interface suggestion; the current fix looks correct.

junghoon-vans (approved):
The event-time accumulator fix looks correct, and the lookup consolidation in f9d60cf is a good improvement. The remaining snapshot-oriented interface suggestion is non-blocking. Approving.

contract/r/gnoswap/staker/v1/reward_calculation_pool.gno
jinoosss: aa5ab590acb43d4010b04197115e4fb7337d81ea

PR #1390, opened 2026-08-18
feat(staker): collect emission and external incentive rewards separately

jinoosss:
## Summary

`CollectReward` harvested the GNS emission reward and every external incentive reward of a position in one call. A staker could not collect one reward source without the other, and could not collect one incentive without every other incentive of the deposit.

This PR adds two entry points next to it:

| Entry point | Collects |
|---|---|
| `CollectEmissionReward(cur, positionId)` | GNS emission only |
| `CollectExternalIncentiveReward(cur, positionId, incentiveId)` | one external incentive |
| `CollectReward(cur, positionId)` | everything, unchanged |

The getters were already split this way (`CollectableEmissionReward` / `CollectableExternalIncentiveReward`), so the collect API now matches them.

## Why the split is safe

The two reward sources already track their collection cursors independently:

- emission: one `lastCollectTime` on the deposit
- external: one `lastCollectTime` per incentive id

Collecting one source, or one incentive, therefore leaves everything else fully accruable. This is the same property the existing code relies on when it skips an incentive without advancing its cursor because the incentive is unfunded, has not started, or yields zero.

Warm-up ratios are read-only during a collect (`SetWarmups` is only called on stake), so narrowing a collect cannot change how warm-up is applied.

## Implementation

Calculation, state update and payout are now separate, and the three entry points compose them:

**Calculation** (read-only, shared with the view getters):

- `calculateInternalPositionReward` — emission
- `calculateExternalPositionReward(param, incentiveId)` — one incentive
- `calculatePositionReward` — both

**State update** (collect only):

- `updateInternalPositionReward` — lazy pool persist + reward cache
- `updateExternalPositionReward` — lazy pool persist + deposit incentive index
- `updatePositionReward` — all of them

**Payout:**

- `collectContext` — the deposit, timestamps, caller identity and the accumulator snapshot every delivery in one call reads
- `deliverEmissionReward` — fee, unclaimable settlement, `totalEmissionSent`, cursor advance, GNS transfers, events
- `deliverExternalIncentiveReward` — incentive debit, cursor advance, fee, reward token transfer, events

`CollectReward` still performs exactly one calculation, one state update and one accumulator snapshot for both sources, then delivers each incentive before the emission reward — the order the two paths ran in before the split. Its cost and its event stream are unchanged, and `UnStakeToken` is unaffected.

Incentive discovery still runs over every incentive even when the calculation is narrowed to one, so the deposit's incentive index stays complete regardless of which incentive the caller asked for. The update now also records the incentive ids the calculation produced a reward for, in calculation order, so delivery walks a slice instead of iterating a map.

`assertIsIncentiveOfDeposit` rejects an incentive id that does not target the pool the position is staked in, so a wrong id aborts instead of silently collecting nothing.

## Behaviour notes

Two things change for callers that use the new entry points. Neither affects `CollectReward`.

1. **`MintAndDistributeGns` runs on the emission path only.** A staker who only collects external incentives no longer triggers it. External reward calculation does not depend on emission state, so amounts are unaffected.
2. **Unclaimable settlement and reward-cache updates run on the emission path only.** External-only collects no longer settle the pool's unclaimable GNS to the community pool, nor advance the pool reward cache. Nothing is lost — the accumulation stays and is settled by the next emission collect — but off-chain consumers that read the reward cache should be aware the cache advances less often.

Also fixes the `CollectReward` proxy documentation, which described the return values as pool path / staking details / internal rewards / external rewards while the implementation returns the emission amount sent to the user, the emission penalty, and the per-token external reward and penalty maps.

## Commits

1. `refactor(staker)` — split position reward calculation per reward source
2. `feat(staker)` — collect emission and a single external incentive separately
3. `feat(staker)` — expose the split collect entry points through the proxy
4. `test(staker)` — cover independent emission and external reward collection

Each commit builds on its own.

## Testing

New tests in `staker/v1` cover the collection cursors rather than the payout amounts, because the package-level test environment does not run GNS emission (the payout itself stays covered by the scenario tests):

- an emission-only collect moves no reward token and does not advance any per-incentive cursor
- a single-incentive collect moves no GNS and does not advance the deposit's internal cursor
- a single-incentive collect leaves the deposit's other incentive fully collectable
- an incentive belonging to another pool is rejected

Passing: `r/gnoswap/staker`, `r/gnoswap/staker/v1`, `r/scenario/upgrade`, `r/scenario/staker`, `r/scenario/position`, `r/scenario/emission`, `r/scenario/halt`, `r/scenario/getter`. The scenario golden outputs are unchanged, which is the evidence that existing `CollectReward` behaviour is preserved.

## Notes for reviewers

- The new test file is named `staker_collect_reward_split_test.gno` so it sorts after `protocol_fee_unstaking_test.gno`. The `staker/v1` package tests share realm state and a collect performed here leaves GNS on the staker realm, which the insufficient-balance case of `TestHandleStakingRewardFee` relies on being empty. This is a pre-existing ordering dependency in the suite, not something introduced here.
- `CollectExternalIncentiveReward` returns the collected reward amount **before** the staking reward fee, matching what `CollectReward`'s external map has always contained.

## Summary by CodeRabbit

* **New Features**
  * Added separate collection options for GNS emission rewards and individual external incentives.
  * Reward collection now reports emission amounts, penalties, external incentive payouts, and related penalties.
  * External incentives can be collected independently from other reward types and incentives.

* **Bug Fixes**
  * Added validation to prevent collecting incentives associated with a different liquidity pool.
  * Improved handling of calculation and collection errors.
  * Preserved collection eligibility for zero-reward warmup scenarios and temporarily underfunded incentives.

* **Tests**
  * Added coverage for independent collection, multiple incentives, invalid pool scenarios, and zero-reward warmup cases.


contract/r/gnoswap/staker/v1/staker_collect_reward_split_test.gno
```diff
@@ -0,0 +1,259 @@
...
+
+	mockNFTMint(cur, addr01, positionId)
+	mockNFTApprove(cur, stakerAddr, positionId)
+	testing.SetRealm(testing.NewUserRealm(addr01))
+	sr.StakeToken(cross(cur), positionId, "")
+
+	testing.SetRealm(testing.NewUserRealm(addr01))
+	gnsApprove(cur, t, addr01, stakerAddr, maxApprove)
+
+	startTimestamp := int64(1234569600)
```
junghoon-vans: Could we avoid hardcoding `1234569600` here? In the test runtime this timestamp is only shortly after the current chain time, while `CreateExternalIncentive` requires the start time to be at least 24 hours in the future. The setup can therefore abort before exercising the collection logic. Please derive this from the test clock and advance enough blocks for the incentive to start.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -440,161 +449,357 @@ func (s *stakerV1) transferDeposit(_ int, rlm realm, positionId uint64, owner, c
...
+	rlm realm,
+	ctx *collectContext,
+	incentiveId string,
+	rewardAmount int64,
+	externalPenalty int64,
+) (string, int64, int64) {
+	// Skip when user reward is zero.
+	// Do not update last collect time so the reward accrues until
+	// the next collection where a non-zero amount can be delivered.
+	if rewardAmount == 0 {
```
junghoon-vans, line 676: Could we handle the case where `rewardAmount == 0` but `externalPenalty > 0`? The warmup calculation can round the user reward down to zero while still producing a penalty. This early return skips the penalty accounting and cursor advancement for this call. Please clarify the intended behavior and add a regression test, or explicitly rule out this state.

jinoosss: https://github.com/gnoswap-labs/gnoswap/pull/1390/changes/9ffaf24a1648239891358c14be07880a2152c5c5

junghoon-vans (approved):

junghoon-vans (approved):

PR #1392, opened 2026-08-18
refactor(protocol_fee): settle pending fees per token and accrue rewards on collection

jinoosss:
## Summary
- Settle the pending protocol fee of only the token being handled, instead of forwarding the whole pending map in one transaction.
- Base gov/staker reward accrual on the amount reserved when a fee reaches the protocol fee realm, so a stake change performs no cross-realm call at all.
- Walk only the tokens that collected a fee since the last accrual, instead of every token that has ever collected one.
- Persist only changed reward-debt entries per staker, instead of replacing the whole map on every stake change and claim.

The set of tokens that have ever collected a protocol fee never shrinks, while the set that collected one since the last accrual is normally one to three tokens. Most of the saving below comes from replacing the former with the latter.

### Fee collection — `HandleWithdrawalFee`, `CreatePool`, `handleSwapFee`, `handleStakingRewardFee`

Before:
1. `addPendingProtocolFee(token, amount)` copies the pending map, updates one key, stores it.
2. `addToProtocolFee()` iterates every pending entry.
3. Each entry costs a `SafeGRC20Approve` and a `pf.AddToProtocolFee`, so every pending token adds two cross-realm calls.
4. It rebuilds the map from the entries that could not be settled and stores it.

Note that steps 1 and 4 cancel out for a fee that is forwarded successfully: the entry is written and then deleted in the same call, at the cost of two store writes and two freshly allocated maps. `HandleWithdrawalFee` paid that four times over for its two tokens.

After:
1. `settleProtocolFee(token, amount)` forwards the amount plus anything already pending, in one `SafeGRC20Approve` and one `pf.AddToProtocolFee`.
2. A pending entry is written only when the fee cannot be forwarded, so the common path performs no store write at all.
3. A token that does carry a backlog settles it together with the new fee, in a single cross-realm call rather than two.

**Optimized:** a fee-bearing call makes two cross-realm calls in total, rather than two for every pending token, and no longer stages the fee in storage on the way. This is also the correctness fix — a panic crossing a realm boundary cannot be contained, so previously one bad token aborted the transaction of every other token.

### `Delegate` / `Undelegate`

Before:
1. `pf.DistributeProtocolFee()` iterates every reserved token awaiting distribution; for each one it computes the split, calls `removeReservedToken` (a linear scan plus a slice rewrite), and performs up to two `SafeGRC20Transfer`.
2. `pf.GetActualDistributedToGovStaker()` walks the full bptree into a map holding one entry per token ever seen, which the proxy then clones again.
3. `updateAccumulatedProtocolFeeX128PerStake` iterates that whole map and builds two more maps of the same size.
4. `addStake` calls `updateRewardDebtX128`, which **replaces the staker's whole reward-debt map**, allocating a fresh map plus one `u256` per token, and writes an accumulated-reward entry for every token even when the reward is zero.
5. Stores the manager.

After:
1. `pf.ConsumeAccrualPendingProtocolFees()` reads the reserved share of the tokens that collected a fee since the last accrual, and clears that list. **No transfers.**
2. `updateAccumulatedProtocolFeeX128PerStake` iterates only those tokens.
3. `addStake` still reads every accumulator entry, but **writes only the ones whose value changed**.
4. Stores the manager.

**Optimized:** token transfers drop to zero; the accrual loop covers only the tokens that just collected a fee instead of every token ever seen; and the per-staker reward debt is written entry by entry instead of having its whole map replaced. What remains proportional to the total token count is a read loop that performs no storage writes.

### `CollectReward`

Step 1 is unchanged: the payout leaves gov/staker's own balance, so `pf.DistributeProtocolFee()` still pulls the reserved fees in first. The remaining steps get the same reductions as `Delegate` — the accrual loop shrinks to the tokens that just collected a fee, and the reward-debt map is no longer replaced.

### `CollectProtocolFeeReward(tokenPath)`

Before and after are the same shape — `DistributeProtocolFeeByTokenPath` for one token, a single-token accumulator update, a single-token claim, one transfer. Only the basis changed, from `GetActualDistributedToGovStakerByTokenPath` to `GetAccuTransferToGovStakerByTokenPath`. This path already handled one token at a time and was already isolated from the others.

### Summary of per-transaction cost

| Path | Before | After |
| --- | --- | --- |
| Fee collection | two cross-realm calls per pending token; two store writes per fee | two cross-realm calls in total; no store write unless the fee cannot be forwarded |
| `Delegate` / `Undelegate` | up to two transfers per reserved token; one pass over every token ever seen; the staker's whole reward-debt map replaced | no transfers; one pass over the tokens that just collected a fee; only changed reward-debt entries written |
| `CollectReward` | the distribution loop, plus the same accrual and reward-debt costs as above | the distribution loop is kept; accrual and reward-debt as above |
| `CollectProtocolFeeReward` | one token | unchanged |

`CollectReward` keeps its distribution loop over every reserved token deliberately: it is the convenience path that pays out all tokens at once, and `CollectProtocolFeeReward` is the isolated per-token alternative.

## Behavior changes
- A fee accrues when it is collected rather than when it is moved, so it stays claimable while withdrawals are halted, even though it still cannot be withdrawn.
- A pending entry now only survives while protocol fee collection is halted. Every fee path calls `addToProtocolFee` even when the fee is zero, so the next swap, withdrawal or reward collection touching that token settles the leftover. `GetPendingProtocolFees` exposes what is still waiting.
- `ConsumeAccrualPendingProtocolFees` is restricted to gov/staker rather than admin as well, since clearing the pending list without folding the fees into an accumulator strands them until that token collects another fee.
- Two filetests are updated for the accrual change. The launchpad golden values move because tokens now sit in the protocol fee realm until a claim, with the settlement amounts unchanged; the safe-mode test captures its baseline after the pending fee is collected, so both readings observe the same state.

## Testing
- `make test PKG=gno.land/r/gnoswap/pool/v1` (and `pool`, `router`, `router/v1`, `staker`, `staker/v1`, `position/v1`, `emission`)
- `make test PKG=gno.land/r/gnoswap/protocol_fee/v1` (and `protocol_fee`)
- `make test PKG=gno.land/r/gnoswap/gov/staker/v1` (and `gov/staker`, `gov/governance/v1`, `launchpad/v1`)
- `make test PKG=gno.land/r/gnoswap/scenario/gov/staker` (and `scenario/{pool,router,staker,position,upgrade,emission,launchpad,halt,protocol_fee,getter,gns,rbac,gov/governance}`)
- Each commit was checked out and tested on its own, so the series is bisectable.
- `gno.land/r/gnoswap/test/fuzz` was not run.

## Summary by CodeRabbit

* **New Features**
  * Added visibility into pending protocol fees by token across pools, routers, and staking.
  * Added support for consuming pending fee accruals globally or for a specific token.
  * Protocol fees now settle individually, including previously pending amounts, while preserving fees during halts or forwarding failures.

* **Bug Fixes**
  * Corrected reward calculations to use accrued rather than already distributed protocol fees.
  * Preserved previously accrued token rewards and reward state during updates.

### DIFF
--- a/contract/r/gnoswap/gov/staker/v1/state.gno
+++ b/contract/r/gnoswap/gov/staker/v1/state.gno
@@ -516,7 +516,7 @@ func (g *govStakerV1) claimAccruedProtocolFeeRewardByTokenPath(address string, t
 func (g *govStakerV1) claimRewardProtocolFeeRewardByTokenPath(_ int, rlm realm, address string, tokenPath string, currentTimestamp int64) (int64, error) {
 	pf.DistributeProtocolFeeByTokenPath(cross(rlm), tokenPath)
 
-	accruedAmount := pf.GetAccuTransferToGovStakerByTokenPath(tokenPath)
+	accruedAmount := pf.ConsumeAccrualPendingProtocolFeeByTokenPath(cross(rlm), tokenPath)
 	protocolFeeRewardManager, reward, processed, err := g.claimAccruedProtocolFeeRewardByTokenPath(address, tokenPath, accruedAmount, currentTimestamp)
 	if err != nil {
 		return 0, err
--- a/contract/r/gnoswap/gov/staker/v1/state.gno
+++ b/contract/r/gnoswap/gov/staker/v1/state.gno
@@ -397,14 +397,12 @@ func (g *govStakerV1) removeLaunchpadProjectDeposit(ownerAddress string) bool {
 //   - amount: amount of stake to add
 //   - currentTimestamp: current timestamp
 func (g *govStakerV1) addStakeProtocolFeeReward(_ int, rlm realm, address string, amount int64, currentTimestamp int64) error {
-	pf.DistributeProtocolFee(cross(rlm))
-
-	distributedAmounts := g.getDistributedProtocolFees()
+	accruedAmounts := g.getAccruedProtocolFees()
 
 	protocolFeeRewardManager := g.store.GetProtocolFeeRewardManager()
 	resolvedManager := NewProtocolFeeRewardManagerResolver(protocolFeeRewardManager)
 
-	if err := resolvedManager.updateAccumulatedProtocolFeeX128PerStake(distributedAmounts, currentTimestamp); err != nil {
+	if err := resolvedManager.updateAccumulatedProtocolFeeX128PerStake(accruedAmounts, currentTimestamp); err != nil {
 		return err
 	}
 
@@ -418,7 +416,7 @@ func (g *govStakerV1) addStakeProtocolFeeReward(_ int, rlm realm, address string
 	}
 
 	// Emit the protocol fee reward accumulation
-	emitUpdateProtocolFeeRewardAccumulation(protocolFeeRewardManager, distributedAmounts)
+	emitUpdateProtocolFeeRewardAccumulation(protocolFeeRewardManager, accruedAmounts)
 
 	return nil
 }
@@ -431,14 +429,12 @@ func (g *govStakerV1) addStakeProtocolFeeReward(_ int, rlm realm, address string
 //   - amount: amount of stake to remove
 //   - currentTimestamp: current timestamp
 func (g *govStakerV1) removeStakeProtocolFeeReward(_ int, rlm realm, address string, amount int64, currentTimestamp int64) error {
-	pf.DistributeProtocolFee(cross(rlm))
-
-	distributedAmounts := g.getDistributedProtocolFees()
+	accruedAmounts := g.getAccruedProtocolFees()
 
 	protocolFeeRewardManager := g.store.GetProtocolFeeRewardManager()
 	resolvedManager := NewProtocolFeeRewardManagerResolver(protocolFeeRewardManager)
 
-	if err := resolvedManager.updateAccumulatedProtocolFeeX128PerStake(distributedAmounts, currentTimestamp); err != nil {
+	if err := resolvedManager.updateAccumulatedProtocolFeeX128PerStake(accruedAmounts, currentTimestamp); err != nil {
 		return err
 	}
 
@@ -451,7 +447,7 @@ func (g *govStakerV1) removeStakeProtocolFeeReward(_ int, rlm realm, address str
 	}
 
 	// Emit the protocol fee reward accumulation
-	emitUpdateProtocolFeeRewardAccumulation(protocolFeeRewardManager, distributedAmounts)
+	emitUpdateProtocolFeeRewardAccumulation(protocolFeeRewardManager, accruedAmounts)
 
 	return nil
 }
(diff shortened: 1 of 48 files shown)


contract/r/gnoswap/gov/staker/v1/state.gno
```diff
@@ -518,8 +516,8 @@ func (g *govStakerV1) claimDistributedProtocolFeeRewardByTokenPath(address strin
 func (g *govStakerV1) claimRewardProtocolFeeRewardByTokenPath(_ int, rlm realm, address string, tokenPath string, currentTimestamp int64) (int64, error) {
 	pf.DistributeProtocolFeeByTokenPath(cross(rlm), tokenPath)
 
-	distributedAmount := pf.GetActualDistributedToGovStakerByTokenPath(tokenPath)
-	protocolFeeRewardManager, reward, processed, err := g.claimDistributedProtocolFeeRewardByTokenPath(address, tokenPath, distributedAmount, currentTimestamp)
+	accruedAmount := pf.GetAccuTransferToGovStakerByTokenPath(tokenPath)
+	protocolFeeRewardManager, reward, processed, err := g.claimAccruedProtocolFeeRewardByTokenPath(address, tokenPath, accruedAmount, currentTimestamp)
```
junghoon-vans, line 520: Clear this token from accrualPendingTokens after a token-scoped claim

This path updates the reward manager using the cumulative amount and emits the per-token event, but it never consumes or removes tokenPath from protocol_fee's accrualPendingTokens. If token A is claimed here and a later Delegate/UnDelegate runs, ConsumeAccrualPendingProtocolFees still returns token A, so it is reprocessed with a zero delta, causing unnecessary cross-realm reads, accumulator comparisons, storage work, and potentially a no-op accumulation event.

Please add a token-scoped consume/remove operation that preserves other pending tokens, or otherwise clear only tokenPath after this claim succeeds.

jinoosss: Good catch — done in f94345839.

Added `ConsumeAccrualPendingProtocolFeeByTokenPath`, the single-token counterpart of the batch consume, and the token-scoped claim path now uses it, so a token it has already folded into the accumulator is no longer handed back by a later consume.

Two things came out of following the suggestion:

- `RemoveAccrualPendingToken` writes only when the token was actually listed, so settling a token that collected no fee costs no store write. (`RemoveReservedToken` next to it always writes; left alone here.)
- The list was inaccurate for a second reason: `reserveCollectedProtocolFee` listed a token regardless of `toGovStakerAmount`, so with `devOpsPct` at 10000 every fee listed a token whose accumulator can never move. It is now listed only when gov/staker actually has a share to accrue — without this, the redundant reprocessing and the no-op event would still happen through that path.

Beyond the reported symptoms, the stale entry also made `GetAccrualPendingProtocolFees` report a token as pending accrual after it had been claimed, which is a wrong answer rather than just wasted work. Worth noting that the storage cost you listed did not actually occur: the write guards added in 307ad0abc already skipped both entries at a zero delta. The extra read and the no-op event did.

While there, guarded the writes in `updateAccumulatedProtocolFeeX128PerStakeForToken`, which stored both entries unconditionally even at a zero delta — the all-token path already had that guard, the single-token one was missed.

Covered by two tests in `protocol_fee/v1`: one asserts a single-token consume leaves the other pending tokens alone and that a later batch consume no longer returns the claimed one, the other that a fee leaving gov/staker nothing is never listed.

junghoon-vans (approved):
Reviewed the current head f94345839d8a09ac1575f115df3a457eb2ab0d16. The pending protocol-fee settlement and token-scoped accrual changes are correct, and the focused protocol_fee/v1 and gov/staker/v1 tests pass. The remaining full-map reward-debt traversal is a non-blocking optimization follow-up; no blocking issues found.

PR #1393, opened 2026-08-19
test(scenario): full cross-function upgrade scenario and data-validating upgrade filetests

jinoosss:
## Summary

Adds a single full-lifecycle upgrade scenario that exercises every cross (crossing) entry point of the eight upgradeable contracts, adds the four missing per-contract `upgrade_with_validate_data` filetests, and turns the existing validate-data filetests into real assertions instead of golden-output prints.

## Why

The upgrade filetests we had covered one contract at a time and mostly *printed* getter output into the golden file. A data-layout regression in a new implementation would therefore show up as a golden diff to be re-blessed, not as a failing assertion — and no test drove the whole protocol across an upgrade of every contract at once.

### 1. Full-lifecycle cross-function upgrade scenario (new)

`contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno`

20 sections, with `SkipHeights` sized to the behaviour under test rather than to "some blocks":

| Sections | Content | Time advanced |
|---|---|---|
| 1–3 | Activate v1 on all eight contracts, emission start time, canonical pool, every admin setter + getter assertion | 1 block |
| 4–8 | Pool creation (the 100 GNS fee proves `AddToProtocolFee`), observation cardinality, mint / increase / decrease / collect fee / reposition | 1 block each |
| 6 | Router exact-in and exact-out, single and multi hop (covers `pool.Swap` and `router.SwapCallback`) | 1 block each |
| 9–10 | Pool tier, external incentive, stake, then collect across two warm-up tiers | 10 days |
| 11 | Protocol fee distribution, per token and global | 2 blocks |
| 12 | Delegate / redelegate / undelegate / collect, including the lockup window | lockup + 1 day |
| 13 | Three proposals: text → cancel, community-pool spend → execute, parameter change → execute | smoothing + start delay + voting period + execution delay, twice |
| 14 | Launchpad project, deposit, reward collection, recipient-side collection | 7 days + 4 days |
| **15** | **Snapshot 29 getter values while v1 is active** | — |
| 16 | Upgrade all eight implementations to `v2_valid` | 1 block |
| **17** | **Assert every snapshot field is unchanged** | — |
| 18 | Re-run the same operations through `v2_valid` | 1 day |
| 19–20 | End incentive, unstake, remove tier, sweep protocol fees and launchpad leftovers, final invariants | 91 days + dynamic skip to project end |

Role-gated entry points that an EOA cannot reach are covered through their only legitimate caller: `pool.Mint` / `Burn` / `Collect` / `HandleWithdrawalFee` via position, `pool.Swap` via router, the pool swap hooks via the staker initializer, `router.SwapCallback` via pool, `position.SetPositionOperator` via `StakeToken`, `protocol_fee.AddToProtocolFee` via pool/position/router/staker, and the launchpad-only `gov/staker` entry points via the project recipient.

Amounts are asserted, not just printed. A few that pin real arithmetic:

- launchpad: a sole depositor collects `13,333,719` on day 4 of a 30-day tier funded with `100,000,000` (= `100,000,000 × 4/30`)
- protocol fee 40/60 split: `6,023 / 9,036` on the same accrual
- warm-up: 39.8% of the internal reward reaches the user after 10 days (30% for 5 days, then 50%)
- community-pool spend proposal transfers exactly `1,000` to the recipient
- the parameter-change proposal moves `staker.GetUnstakingFee()` from 100 to 150, proving the governance → staker permission path

### 2. `upgrade_with_validate_data` filetests for the remaining contracts (new)

`protocol_fee`, `launchpad`, `gov/governance` and `gov/staker` had none. Each writes state with v1, snapshots the getters, walks v1 → `v2_invalid` (asserting the calls abort) → `v3_valid`, asserts the snapshot is unchanged, and keeps operating on the v1-created data.

`launchpad` goes v1 → `v2_valid` only — see the note below.

### 3. Existing validate-data filetests strengthened

`pool`, `position` (collect fee), `router` (swap) and `staker` (collect reward) now snapshot the getters under v1 and assert them field by field after the upgrade, instead of only printing them:

- **pool** — 30 fields: tick liquidity gross/net, the pool-side position entry, balances, fee growth, protocol fees, plus a check that a post-upgrade mint only adds liquidity to the shared tick range and leaves position 1 untouched
- **position** — fee growth inside last, tokens owed, unclaimed fee, and the invariant that a post-upgrade swap accrues on top of the v1 values rather than resetting them
- **router** — swap fee, pool balance direction, position identity
- **staker** — 15 deposit identity fields, pool tier and ratio, and reward ledgers moving forward only

## Summary by CodeRabbit

* **Documentation**
  * Clarified liquidity reduction and fee collection behavior.
  * Documented that accrued swap fees are collected first, returned fees reflect applicable withdrawal charges, and principal is returned separately without that charge.
  * Clarified the distinction between net fees paid out and fees collected before deductions.

### DIFF
--- a/contract/r/gnoswap/position/proxy.gno
+++ b/contract/r/gnoswap/position/proxy.gno
@@ -79,11 +79,11 @@ func IncreaseLiquidity(
 // Returns:
 //   - uint64: position ID
 //   - string: removed liquidity amount
-//   - string: amount of token0 removed
-//   - string: amount of token1 removed
+//   - string: collected fee amount of token0 (after the withdrawal fee)
+//   - string: collected fee amount of token1 (after the withdrawal fee)
+//   - string: amount of token0 returned to the caller
+//   - string: amount of token1 returned to the caller
 //   - string: pool path
-//   - string: net amount of token0 after fees
-//   - string: net amount of token1 after fees
 func DecreaseLiquidity(
 	cur realm,
 	positionId uint64,
--- a/contract/r/gnoswap/position/proxy.gno
+++ b/contract/r/gnoswap/position/proxy.gno
@@ -76,13 +76,16 @@ func IncreaseLiquidity(
 //   - amount1MinStr: minimum amount of token1
 //   - deadline: transaction deadline
 //
+// The accrued swap fees are collected first, so the returned fee amounts are
+// net of the withdrawal fee while the returned principal is not subject to it.
+//
 // Returns:
 //   - uint64: position ID
 //   - string: removed liquidity amount
-//   - string: collected fee amount of token0 (after the withdrawal fee)
-//   - string: collected fee amount of token1 (after the withdrawal fee)
-//   - string: amount of token0 returned to the caller
-//   - string: amount of token1 returned to the caller
+//   - string: token0 fees collected, net of the withdrawal fee
+//   - string: token1 fees collected, net of the withdrawal fee
+//   - string: principal amount of token0 returned to the caller
+//   - string: principal amount of token1 returned to the caller
 //   - string: pool path
 func DecreaseLiquidity(
 	cur realm,
@@ -135,11 +138,11 @@ func Reposition(
 //
 // Returns:
 //   - uint64: position ID
-//   - string: amount of token0 collected
-//   - string: amount of token1 collected
+//   - string: token0 fees paid out, net of the withdrawal fee
+//   - string: token1 fees paid out, net of the withdrawal fee
 //   - string: pool path
-//   - string: token0 path
-//   - string: token1 path
+//   - string: token0 fees collected, before the withdrawal fee
+//   - string: token1 fees collected, before the withdrawal fee
 func CollectFee(
 	cur realm,
 	positionId uint64,
--- a/contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
+++ b/contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
@@ -211,11 +211,13 @@ type upgradeSnapshot struct {
 	stakerUnstakingFee uint64
 	stakerDepositGns   int64
 	stakerAllowedToken int
+	stakerAllowedList  []string
 	depositLiquidity   string
 	depositStakeTime   int64
 	depositTickLower   int32
 	depositTickUpper   int32
 	depositWarmupCount int
+	depositWarmups     []staker.Warmup
 	depositCollected   int64
 	totalEmissionSent  int64
 	incentiveTotal     int64
@@ -1139,11 +1141,13 @@ func captureSnapshot() upgradeSnapshot {
 		stakerUnstakingFee: staker.GetUnstakingFee(),
 		stakerDepositGns:   staker.GetDepositGnsAmount(),
 		stakerAllowedToken: len(staker.GetAllowedTokens()),
+		stakerAllowedList:  staker.GetAllowedTokens(),
 		depositLiquidity:   staker.GetDepositLiquidityAsString(1),
 		depositStakeTime:   staker.GetDepositStakeTime(1),
 		depositTickLower:   staker.GetDepositTickLower(1),
 		depositTickUpper:   staker.GetDepositTickUpper(1),
 		depositWarmupCount: len(staker.GetDepositWarmUp(1)),
+		depositWarmups:     staker.GetDepositWarmUp(1),
 		depositCollected:   staker.GetDepositCollectedInternalReward(1),
 		totalEmissionSent:  staker.GetTotalEmissionSent(),
 		incentiveTotal:     staker.GetIncentiveTotalRewardAmount(barFoo500PoolPath, externalIncentiveId),
@@ -1199,6 +1203,13 @@ func printSnapshot(s upgradeSnapshot) {
 	println("[EXPECTED] proposal ids:", len(s.proposalIDs), "delegation ids:", len(s.delegationIDs))
 	println("[EXPECTED] param proposal yea / nay / status:", s.paramProposalYea, s.paramProposalNay, s.paramProposalState)
 	println("[EXPECTED] launchpad project active:", s.projectActive)
+	for _, tokenPath := range s.stakerAllowedList {
+		println("[EXPECTED] staker allowed token:", tokenPath)
+	}
+	for i, warmup := range s.depositWarmups {
+		println("[EXPECTED] deposit warmup", i, "duration / nextWarmupTime / ratio:",
+			warmup.TimeDuration, warmup.NextWarmupTime, warmup.WarmupRatio)
+	}
 }
 
 // ---------------------------------------------------------------------------
@@ -1314,6 +1325,18 @@ func validateSnapshotUnchanged() {
 	uassert.Equal(t, len(snap.proposalIDs), len(after.proposalIDs))
 	for i, id := range snap.proposalIDs {
 		uassert.Equal(t, id, after.proposalIDs[i])
+	}
+	uassert.Equal(t, len(snap.stakerAllowedList), len(after.stakerAllowedList))
+	for i, tokenPath := range snap.stakerAllowedList {
+		uassert.Equal(t, tokenPath, after.stakerAllowedList[i])
+	}
+	// warm-up records carry the reward ratio schedule, so every field of every
+	// tier is compared, not just how many tiers there are
+	uassert.Equal(t, len(snap.depositWarmups), len(after.depositWarmups))
+	for i, warmup := range snap.depositWarmups {
+		uassert.Equal(t, warmup.TimeDuration, after.depositWarmups[i].TimeDuration)
+		uassert.Equal(t, warmup.NextWarmupTime, after.depositWarmups[i].NextWarmupTime)
+		uassert.Equal(t, warmup.WarmupRatio, after.depositWarmups[i].WarmupRatio)
 	}
 
 	println("[EXPECTED] every snapshot field is identical after the upgrade")
@@ -1679,6 +1702,13 @@ func finalInvariants(cur realm) {
 // [EXPECTED] proposal ids: 3 delegation ids: 3
 // [EXPECTED] param proposal yea / nay / status: 6500000000 0 executed
 // [EXPECTED] launchpad project active: true
+// [EXPECTED] staker allowed token: gno.land/r/gnoswap/gns.GNS
+// [EXPECTED] staker allowed token: gno.land/r/gnoland/wugnot.wugnot
+// [EXPECTED] staker allowed token: gno.land/r/onbloc/bar.BAR
+// [EXPECTED] deposit warmup 0 duration / nextWarmupTime / ratio: 432000 1234999970 30
+// [EXPECTED] deposit warmup 1 duration / nextWarmupTime / ratio: 864000 1235863970 50
+// [EXPECTED] deposit warmup 2 duration / nextWarmupTime / ratio: 2592000 1238455970 70
+// [EXPECTED] deposit warmup 3 duration / nextWarmupTime / ratio: 9223372036854775807 9223372036854775807 100
 //
 // [SCENARIO] 16. Upgrade all eight implementations to v2_valid
 // [EXPECTED] pool impl: gno.land/r/gnoswap/pool/v2_valid
--- a/contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
+++ b/contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
@@ -197,20 +197,57 @@ type upgradeSnapshot struct {
 	positionTickLower  int32
 	positionTickUpper  int32
 	positionOwner      address
+	positionOperator   address
+	positionFeeGrowth0 string
+	positionFeeGrowth1 string
+	positionOwed0      int64
+	positionOwed1      int64
+	positionUnclaimed0 string
+	positionUnclaimed1 string
+	positionBurned     bool
+	positionInRange    bool
 	stakerPoolTier     uint64
+	stakerPoolTierRat  uint64
 	stakerUnstakingFee uint64
 	stakerDepositGns   int64
+	stakerAllowedToken int
 	depositLiquidity   string
+	depositStakeTime   int64
+	depositTickLower   int32
+	depositTickUpper   int32
+	depositWarmupCount int
+	depositCollected   int64
+	totalEmissionSent  int64
+	incentiveTotal     int64
+	incentiveDistrib   int64
+	incentiveRefunded  bool
 	devOpsPct          int64
 	govStakerPct       int64
 	accuGovStakerBar   int64
+	accuDevOpsBar      int64
+	reservedTokens     []string
 	totalDelegated     int64
 	totalXGnsSupply    int64
+	totalLocked        int64
 	delegationCount    int
+	delegationIDs      []int64
 	proposalCount      int
+	proposalIDs        []int64
+	paramProposalYea   int64
+	paramProposalNay   int64
+	paramProposalState string
+	spendProposalState string
+	textProposalState  string
+	adminVoteWeight    int64
 	configVersion      int64
+	configVotingPeriod int64
 	projectCount       int
+	projectActive      bool
+	projectRecipient   address
+	projectTier30Total int64
 	launchpadDeposits  int64
+	launchpadDepositID int64
+	depositAmount      int64
 }
 
 func main(cur realm) {
@@ -999,6 +1036,19 @@ func launchpadLifecycle(cur realm) {
 func captureSnapshot() upgradeSnapshot {
 	feeGrowth0, feeGrowth1 := pool.GetFeeGrowthGlobalX128(barFoo500PoolPath)
 	protocolFee0, protocolFee1 := pool.GetProtocolFeesTokens(barFoo500PoolPath)
+	positionFeeGrowth0, positionFeeGrowth1 := position.GetPositionFeeGrowthInsideLastX128(1)
+	positionOwed0, positionOwed1 := position.GetPositionTokensOwed(1)
+	positionUnclaimed0, positionUnclaimed1 := position.GetUnclaimedFee(1)
+	paramStatus, _ := govGovernance.GetProposalStatusByProposalId(paramProposalId)
+	spendStatus, _ := govGovernance.GetProposalStatusByProposalId(spendProposalId)
+	textStatus, _ := govGovernance.GetProposalStatusByProposalId(textProposalId)
+	paramYea, _ := govGovernance.GetYeaByProposalId(paramProposalId)
+	paramNay, _ := govGovernance.GetNayByProposalId(paramProposalId)
+	adminVoteWeight, _ := govGovernance.GetVoteWeight(paramProposalId, adminAddr)
+	projectActive, _ := launchpad.GetProjectActiveStatus(launchpadProjectId)
+	projectRecipient, _ := launchpad.GetProjectRecipient(launchpadProjectId)
+	projectTier30Total, _ := launchpad.GetProjectTierTotalDistributeAmount(launchpadProjectId, 30)
+	launchpadDepositAmount, _ := launchpad.GetDepositAmount(launchpadDepositId)
 
 	return upgradeSnapshot{
 		poolLiquidity:      pool.GetLiquidity(barFoo500PoolPath),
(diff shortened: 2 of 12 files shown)


contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
```diff
@@ -0,0 +1,1523 @@
...
+
+	// snapshot captured right before the v2_valid upgrade
+	snap upgradeSnapshot
+)
+
+// upgradeSnapshot holds the state observed through the public getters right
+// before the upgrade. Every field must be identical after the upgrade: the KV
+// store is shared across implementations, so a data-layout regression in the
+// new implementation shows up here.
+type upgradeSnapshot struct {
```
junghoon-vans, line 184: This snapshot covers only 29 getter values. Important persisted values such as position fee-growth/tokens-owed/unclaimed data, governance proposal IDs/voting info, launchpad active/deposit identity, and staker reward ledgers are not included in the full-lifecycle snapshot. The domain-specific tests cover some of these, but the full set is not compared exactly immediately after the upgrade. Please either document the intentional exclusions or include these getters in the snapshot/assertions.

contract/r/scenario/upgrade/position_collectfee_upgrade_with_validate_data_filetest.gno
```diff
@@ -98,6 +124,69 @@ func main(cur realm) {
...
+	uassert.Equal(t, v1PositionTickLower, position.GetPositionTickLower(positionId))
+	uassert.Equal(t, v1PositionTickUpper, position.GetPositionTickUpper(positionId))
+	// no mint / burn happened, so the in-range liquidity is unchanged; the tick
+	// may have moved because of the post-upgrade swap
+	uassert.Equal(t, v1PoolLiquidity, pool.GetLiquidity(barFoo500PoolPath))
+	uassert.False(t, position.IsBurned(positionId))
+
+	// the post-upgrade swap must keep accruing on top of the v1 values, never
+	// reset them
+	uassert.True(t, feeGrowth0 >= v1PoolFeeGrowth0)
```
junghoon-vans, line 216: This validation runs only after the post-upgrade swap/collect and uses >= for fee growth. If v1 fee growth were reset during the upgrade and the new swap rebuilt the value, this could still pass. The snapshot fields v1PositionFeeGrowth* and v1PositionTokensOwed* are also never compared here. Please validate exact equality immediately after the upgrade, then assert post-upgrade accrual separately.

contract/r/scenario/upgrade/staker_collect_reward_upgrade_with_validate_data_filetest.gno
```diff
@@ -120,6 +141,74 @@ func main(cur realm) {
...
+	uassert.Equal(t, v1PoolTier, staker.GetPoolTier(barFoo500PoolPath))
+	uassert.Equal(t, v1PoolTierRatio, staker.GetPoolTierRatio(barFoo500PoolPath))
+	uassert.Equal(t, v1PoolStakedLiq, staker.GetPoolStakedLiquidity(barFoo500PoolPath))
+	uassert.Equal(t, v1UnstakingFee, staker.GetUnstakingFee())
+	uassert.Equal(t, v1DepositGnsAmount, staker.GetDepositGnsAmount())
+	uassert.Equal(t, v1AllowedTokenCount, len(staker.GetAllowedTokens()))
+	uassert.True(t, staker.IsStaked(positionId))
+
+	// reward ledgers only ever move forward
+	uassert.True(t, staker.GetDepositCollectedInternalReward(positionId) > v1CollectedInternal)
```
junghoon-vans, line 235: After upgrading to v3_valid, the test calls collectReward before checking >. If the existing reward ledgers were reset to zero, a new collection could still make the values greater than the snapshot. Please compare GetDepositCollectedInternalReward and GetTotalEmissionSent exactly immediately after the upgrade, then assert the expected increase after collection separately.

contract/r/scenario/upgrade/router_swap_upgrade_with_validate_data_filetest.gno
```diff
@@ -85,6 +109,66 @@ func main(cur realm) {
...
+	uassert.Equal(t, v1PoolLiquidity, pool.GetLiquidity(barFoo500PoolPath))
+	uassert.Equal(t, v1PositionLiquidity, position.GetPositionLiquidity(positionId))
+	uassert.Equal(t, v1PositionPoolKey, position.GetPositionPoolKey(positionId))
+	uassert.Equal(t, v1PositionTickLower, position.GetPositionTickLower(positionId))
+	uassert.Equal(t, v1PositionTickUpper, position.GetPositionTickUpper(positionId))
+	uassert.False(t, position.IsBurned(positionId))
+
+	// the swap routed through the upgraded router keeps accruing on top of the
+	// v1 values instead of resetting them
+	uassert.True(t, feeGrowth0 >= v1PoolFeeGrowth0)
```
junghoon-vans, line 195: After upgrading to v2_mock, the test performs a second swap before checking >=. A reset fee-growth/protocol-fee state could be repopulated by that swap and still pass; v1PoolProtocolFee* and v1PositionUnclaimed0 are not asserted either. Please add exact snapshot validation immediately after the upgrade and keep second-swap assertions separate.

contract/r/scenario/upgrade/launchpad_upgrade_with_validate_data_filetest.gno
```diff
@@ -0,0 +1,470 @@
...
+	v1Tier30Start, _ = launchpad.GetProjectTierStartTime(projectId, 30)
+	v1Tier30End, _ = launchpad.GetProjectTierEndTime(projectId, 30)
+	v1DepositCount = launchpad.GetDepositCount()
+	v1DepositAmount, _ = launchpad.GetDepositAmount(depositId)
+	v1DepositTier, _ = launchpad.GetDepositTier(depositId)
+	v1DepositProjectID, _ = launchpad.GetDepositProjectID(depositId)
+	v1TotalGNSStaked = launchpad.GetTotalGNSStakedAmount()
+	v1RewardManagers = launchpad.GetProjectTierRewardManagerCount()
+	v1RewardTotalAmount, _ = launchpad.GetProjectTierRewardTotalDistributeAmount(tier30Id)
+	v1ProjectActive, _ = launchpad.GetProjectActiveStatus(projectId)
```
junghoon-vans, line 283: v1ProjectActive is captured but never compared in validateSnapshot. The later deposit-count/total-staked assertions do not directly verify preservation of the active flag. Please add uassert.Equal(t, v1ProjectActive, launchpad.GetProjectActiveStatus(projectId)) to the immediate post-upgrade validation.

contract/r/scenario/upgrade/gov_governance_upgrade_with_validate_data_filetest.gno
```diff
@@ -0,0 +1,396 @@
...
+	nay, _ := govGovernance.GetNayByProposalId(paramProposalId)
+	quorumAmount, _ := govGovernance.GetQuorumAmountByProposalId(paramProposalId)
+	status, _ := govGovernance.GetProposalStatusByProposalId(paramProposalId)
+	configVer, _ := govGovernance.GetConfigVersionByProposalId(paramProposalId)
+	voteWeight, _ := govGovernance.GetVoteWeight(paramProposalId, adminAddr)
+	votedAt, _ := govGovernance.GetVotedAt(paramProposalId, adminAddr)
+	textStatus, _ := govGovernance.GetProposalStatusByProposalId(textProposalId)
+
+	uassert.Equal(t, v1ProposalCount, govGovernance.GetProposalCount())
+	uassert.Equal(t, len(v1ProposalIDs), len(govGovernance.GetProposalIDs(0, 10)))
```
junghoon-vans: GetProposalIDs is compared only by length, not by the actual IDs or ordering. A list with the same number of incorrect proposal IDs would pass. Please compare every returned ID with v1ProposalIDs.

contract/r/scenario/upgrade/protocol_fee_upgrade_with_validate_data_filetest.gno
```diff
@@ -0,0 +1,289 @@
...
+	testing.SetRealm(adminRealm)
+	protocolFee.UpgradeImpl(cross(cur), packagePath)
+	println("[EXPECTED] implementation:", protocolFee.GetImplementationPackagePath())
+	uassert.Equal(t, packagePath, protocolFee.GetImplementationPackagePath())
+}
+
+func validateSnapshot() {
+	uassert.Equal(t, v1DevOpsPct, protocolFee.GetDevOpsPct())
+	uassert.Equal(t, v1GovStakerPct, protocolFee.GetGovStakerPct())
+	uassert.Equal(t, len(v1ReservedTokens), len(protocolFee.GetReservedTokens()))
```
junghoon-vans: GetReservedTokens is also compared only by length. A reserved-token list with changed contents or ordering would pass as long as its length is unchanged. Because this list drives the distribution iteration, please compare each token path.

contract/r/scenario/upgrade/launchpad_upgrade_with_validate_data_filetest.gno
```diff
@@ -0,0 +1,470 @@
...
+	println("[SCENARIO] 4. Collect the tier 30 reward with v1")
+	collectRewardWithV1(cur)
+	println()
+
+	println("[SCENARIO] 5. Snapshot v1 data through the getters")
+	captureV1Snapshot()
+	println()
+
+	println("[SCENARIO] 6. Upgrade to v2_valid and validate the v1 data")
+	upgradeTo(cur, launchpadV2ValidPath)
```
junghoon-vans, line 137: This data-validation test covers only the v1 -> v2_valid success path and does not exercise a v2_invalid failure/permission path. If that is not intentionally covered by the existing generic tests, please add a launchpad v2_invalid fixture and verify that mutation/getter calls are blocked in this lifecycle as well.

contract/r/scenario/upgrade/gov_governance_upgrade_with_validate_data_filetest.gno
```diff
@@ -0,0 +1,396 @@
...
+
+func upgradeTo(cur realm, packagePath string) {
+	testing.SetRealm(adminRealm)
+	govGovernance.UpgradeImpl(cross(cur), packagePath)
+	println("[EXPECTED] implementation:", govGovernance.GetImplementationPackagePath())
+	uassert.Equal(t, packagePath, govGovernance.GetImplementationPackagePath())
+}
+
+func validateSnapshot() {
+	proposer, _ := govGovernance.GetProposerByProposalId(paramProposalId)
```
junghoon-vans: The validation getters discard all errors, so a getter failure is converted into a zero value and can make the failure mode unclear or create a false negative for fields that are not otherwise checked. Since this test is intended to validate data through the public getters, please assert err == nil (or use a checked-getter helper) for each getter call.

contract/r/gnoswap/position/proxy.gno
```diff
@@ -79,11 +79,11 @@ func IncreaseLiquidity(
 // Returns:
 //   - uint64: position ID
 //   - string: removed liquidity amount
-//   - string: amount of token0 removed
-//   - string: amount of token1 removed
+//   - string: collected fee amount of token0 (after the withdrawal fee)
+//   - string: collected fee amount of token1 (after the withdrawal fee)
+//   - string: amount of token0 returned to the caller
+//   - string: amount of token1 returned to the caller
```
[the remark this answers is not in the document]
jinoosss: Adopted the principal-versus-fee wording in 47752b185, with one correction: the withdrawal fee is not applied to the collected principal.

`decreaseLiquidity` calls `collectFee` first, and `collectFee` returns `amount0WithoutFee` / `amount1WithoutFee` (i.e. fees already net of the withdrawal fee). The `amount0` / `amount1` values come straight from `pl.Collect(...)` on the burned liquidity and no withdrawal fee is charged on them — the filetest confirms this: decreasing liquidity with zero accrued fees returns the full principal and adds nothing to the protocol fee accumulator.

So the note now reads "the returned fee amounts are net of the withdrawal fee while the returned principal is not subject to it".

While verifying this I found the same class of error on `CollectFee` right below: its last two return values were documented as token paths, but the implementation returns the gross fee amounts before the withdrawal fee. Fixed in the same commit.

contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
jinoosss: Included them rather than documenting the exclusion — the snapshot went from 29 to 60 fields in 55a569da5.

Added: position fee growth inside last, tokens owed, unclaimed fee, burned/in-range flags and operator; staker deposit stake time, ticks, warm-up tier count, collected internal reward, total emission sent and the external incentive ledger (total / distributed / refunded); governance proposal ids, per-proposal statuses (text, spend, param), yea/nay, admin vote weight and the config voting period; launchpad project active flag, recipient, tier-30 total, current deposit id and deposit amount; protocol fee devops accrual and the reserved-token list; gov staker total locked and delegation ids.

The comparison still runs in section 17, immediately after the upgrade and before any post-upgrade transaction, so every field is checked for exact equality. List-valued fields are compared element by element.

contract/r/scenario/upgrade/position_collectfee_upgrade_with_validate_data_filetest.gno
jinoosss: Fixed in 55a569da5. Agreed — the `>=` check after a fresh swap could not distinguish "preserved" from "wiped and rebuilt".

Scenario 9-1 now runs `validateExactAfterUpgrade()` immediately after the v3_valid upgrade, before the post-upgrade swap, and asserts exact equality on every snapshot field including `v1PositionFeeGrowth0/1`, `v1PositionTokensOwed0/1`, `v1PositionUnclaimed0/1`, the pool fee growth, the pool protocol fees and the pool tick.

Scenario 12 keeps the accrual assertions (`>=` on fee growth, unclaimed back to 0 after collecting) as a separate step.

contract/r/scenario/upgrade/staker_collect_reward_upgrade_with_validate_data_filetest.gno
jinoosss: Fixed in 55a569da5, same shape as the position test.

Scenario 11-1 runs `validateExactAfterUpgrade()` right after the v3_valid upgrade and before `collectReward`, asserting `GetDepositCollectedInternalReward` and `GetTotalEmissionSent` are exactly equal to the snapshot, along with the deposit identity, pool tier / ratio, staked liquidity and config fields.

Scenario 15 keeps the "ledgers only move forward" assertions for after the post-upgrade collection.

contract/r/scenario/upgrade/router_swap_upgrade_with_validate_data_filetest.gno
jinoosss: Fixed in 55a569da5.

Scenario 6-2 now runs `validateExactAfterUpgrade()` between the v2_mock upgrade and the second swap, asserting exact equality on the swap fee, pool liquidity, tick, fee growth, both protocol fee values, both balances, the position identity and `v1PositionUnclaimed0/1` — the two fields you flagged as never asserted are now covered.

Scenario 10 keeps the post-swap accrual assertions (`>=` on fee growth, balance direction).

contract/r/scenario/upgrade/launchpad_upgrade_with_validate_data_filetest.gno
jinoosss: Added in e84cb9739: `validateSnapshot()` now asserts `uassert.Equal(t, v1ProjectActive, projectActive)` against `launchpad.GetProjectActiveStatus(projectId)`, and the getter error is asserted with `uassert.NoError` instead of being discarded.

contract/r/scenario/upgrade/gov_governance_upgrade_with_validate_data_filetest.gno
jinoosss: Fixed in 55a569da5. `validateSnapshot()` now compares the ids themselves:

```go
proposalIDs := govGovernance.GetProposalIDs(0, 10)
uassert.Equal(t, len(v1ProposalIDs), len(proposalIDs))
for i, id := range v1ProposalIDs {
    uassert.Equal(t, id, proposalIDs[i])
}
```

The same element-wise comparison was applied to the delegation ids and the reserved-token list in the full-lifecycle snapshot.

contract/r/scenario/upgrade/protocol_fee_upgrade_with_validate_data_filetest.gno
jinoosss: Fixed in 55a569da5 — agreed that this list matters more than most, since `DistributeProtocolFee` iterates it.

`validateSnapshot()` now compares every token path positionally, and both the snapshot and the post-upgrade state print each entry so the golden output shows the ordering too.

contract/r/scenario/upgrade/launchpad_upgrade_with_validate_data_filetest.gno
jinoosss: It was not covered elsewhere — launchpad simply had no `v2_invalid` fixture, so I added one in e84cb9739.

`implements/v2_invalid/launchpad/test_impl.gno` follows the same shape as the other `v2_invalid` implementations: it wraps `v1.NewLaunchpadV1(store)` and every one of the 57 interface methods panics with `test implementation: <method> not supported` unless it has been activated through `SetActiveMethod`.

The lifecycle is now v1 -> v2_invalid -> v2_valid, and scenario 6 asserts that `DepositGns`, `CollectRewardByDepositId` and the getters (`GetTotalGNSStakedAmount`) all abort while the invalid implementation is active, before scenario 6-1 upgrades to `v2_valid` and validates the snapshot.

contract/r/scenario/upgrade/gov_governance_upgrade_with_validate_data_filetest.gno
jinoosss: Fixed in 55a569da5. Every getter call in the governance `validateSnapshot()` now binds `err` and asserts `uassert.NoError(t, err)` — proposer, title, description, type, yea, nay, quorum, status, config version, vote weight, voted-at, the text proposal status and `GetConfig(v1LatestConfigVer)`.

The launchpad test got the same treatment for its two error-returning getters in e84cb9739.

jinoosss:
## Review feedback addressed

Thanks — all ten comments were actionable and are now fixed across three commits.

| Commit | Content |
|---|---|
| 47752b185 | `docs(position)`: principal-versus-fee wording, plus the same class of error found on `CollectFee` |
| e84cb9739 | `test(scenario)`: new launchpad `v2_invalid` fixture and the blocked-path lifecycle |
| 55a569da5 | `test(scenario)`: tighter post-upgrade validation (timing, list comparison, getter errors, wider snapshot) |

### Validate before the next transaction, not after it

The strongest point in the review: `position`, `router` and `staker` only compared their ledgers once a post-upgrade swap or collection had already run, using `>=`. A value the upgrade wiped and the new transaction rebuilt would have passed.

Each of those tests now has an exact-equality step that runs immediately after the upgrade and before anything else happens:

- position → scenario 9-1, covering `feeGrowthInsideLast`, `tokensOwed`, `unclaimedFee`, pool fee growth, pool protocol fees, pool tick
- staker → scenario 11-1, covering `GetDepositCollectedInternalReward` and `GetTotalEmissionSent` plus the deposit identity and tier config
- router → scenario 6-2, covering swap fee, both protocol fee values, both balances and `unclaimedFee`

The accrual assertions stayed, but as a clearly separate later step.

### List-valued getters compared element by element

`GetReservedTokens`, `GetProposalIDs` and the full-lifecycle delegation ids were compared by length only. All three now compare positionally, and the reserved-token list prints each entry into the golden output so ordering is visible — it drives the `DistributeProtocolFee` iteration.

### Getter errors asserted

The governance validation discarded every error, turning a failed lookup into a zero value. All thirteen getter calls there now assert `uassert.NoError`, and the launchpad test does the same for its error-returning getters.

### Full-lifecycle snapshot: 29 → 60 fields

Rather than documenting the exclusions, the missing getters were added: position fee growth / tokens owed / unclaimed / burned / in-range / operator; staker stake time, ticks, warm-up tiers, collected reward, total emission sent and the external incentive ledger; governance proposal ids, per-proposal statuses, vote tallies, admin vote weight, config voting period; launchpad active flag, recipient, tier totals, deposit identity; protocol fee reserved tokens and devops accrual; gov staker total locked and delegation ids.

### Launchpad `v2_invalid` fixture

It genuinely was not covered — launchpad was the only upgradeable contract without a `v2_invalid` fixture. Added one following the existing pattern (wraps v1, every one of the 57 interface methods panics unless activated). The launchpad lifecycle is now v1 → v2_invalid → v2_valid, asserting that `DepositGns`, `CollectRewardByDepositId` and the getters are all blocked while the invalid implementation is active.

### One correction to the bot's suggestion

The suggested note "Applies withdrawal fee on collected amounts" would have been wrong for the principal. `DecreaseLiquidity` collects the fees first — those are net of the withdrawal fee — while the principal returned from `pl.Collect(...)` is not charged one. The filetest confirms it: decreasing liquidity with no accrued fees returns the full principal and adds nothing to the protocol fee accumulator. The doc now states both halves explicitly.

### Verification

- `gno test ./gno.land/r/gnoswap/scenario/upgrade -v` → 38/38 pass (golden outputs regenerated for the seven touched files)
- `gno test ./gno.land/r/gnoswap/launchpad ./gno.land/r/gnoswap/position` → pass
- `gofumpt` clean; still no production logic changed, only the two doc comments

jinoosss:
## Position lifecycle now accrues real swap fees (39d82e0ca)

The position lifecycle section was decreasing liquidity on a position that had never earned a fee, so `DecreaseLiquidity` collected zero and the withdrawal fee printed as `0` — the code path ran, but nothing about it was verified.

It now trades 2,000,000 in each direction before the decrease, and the assertions pin the entire split:

```
[EXPECTED] fee-accrual swap bar->foo in/out: 2000000 -1976553
[EXPECTED] fee-accrual swap foo->bar in/out: 2000000 -1981465
[EXPECTED] pool tick after the fee-accrual swaps: 0
[EXPECTED] fees accrued before decrease (gross): 597 597
[EXPECTED] principal returned: 29999012 30000987
[EXPECTED] fees returned with the principal (net): 592 592
[EXPECTED] withdrawal fee accrued (BAR / FOO): 5 5
[EXPECTED] withdrawal fee expected from the gross fee: 5 5
```

The numbers are checked, not just printed:

- **gross fee 597** — each 2,000,000 swap pays the 0.05% pool fee (1,000), the protocol takes its 1/4 share (250), the LPs split the remaining 750, and position 1 holds 1,280,302,477 of the 1,608,058,052 in-range liquidity → 750 × 79.6% = 597
- **withdrawal fee 5** — asserted equal to `floor(gross × pool.GetWithdrawalFee() / 10000)`, measured as the delta on the protocol fee accumulator strictly around the `DecreaseLiquidity` call so the router fees from the swaps stay outside the window
- **net 592** — asserted equal to `gross − withdrawal fee`
- the **principal** is asserted separately, confirming no withdrawal fee is charged on it (which is the distinction the doc fix in 47752b185 was about)

Golden output regenerated — the two extra swaps shift the downstream amounts in the later sections. Full suite still green: `gno test ./gno.land/r/gnoswap/scenario/upgrade` → 38/38.

contract/r/scenario/upgrade/full_lifecycle_v2_valid_upgrade_filetest.gno
```diff
@@ -0,0 +1,1739 @@
...
+		positionOwed1:      positionOwed1,
+		positionUnclaimed0: positionUnclaimed0,
+		positionUnclaimed1: positionUnclaimed1,
+		positionBurned:     position.IsBurned(1),
+		positionInRange:    position.IsInRange(1),
+		stakerPoolTier:     staker.GetPoolTier(barFoo500PoolPath),
+		stakerPoolTierRat:  staker.GetPoolTierRatio(barFoo500PoolPath),
+		stakerUnstakingFee: staker.GetUnstakingFee(),
+		stakerDepositGns:   staker.GetDepositGnsAmount(),
+		stakerAllowedToken: len(staker.GetAllowedTokens()),
```
junghoon-vans, line 1143: **Suggestion — Blocker: no**

Could we snapshot and compare the elements of `GetAllowedTokens()` and the `Warmup` records returned by `GetDepositWarmUp(1)`, rather than only their lengths? A reset, reordering, or changed token/warmup value with the same count would still pass `validateSnapshotUnchanged()` at lines 1268 and 1273. Comparing each token path and `Warmup` field would make the upgrade-preservation assertion detect content changes as well. This is non-blocking, but it would strengthen the full-lifecycle snapshot.

jinoosss: Good catch — done in a927b9351.

Both lists are now snapshotted by value and compared entry by entry in `validateSnapshotUnchanged()`:

```go
uassert.Equal(t, len(snap.stakerAllowedList), len(after.stakerAllowedList))
for i, tokenPath := range snap.stakerAllowedList {
    uassert.Equal(t, tokenPath, after.stakerAllowedList[i])
}
uassert.Equal(t, len(snap.depositWarmups), len(after.depositWarmups))
for i, warmup := range snap.depositWarmups {
    uassert.Equal(t, warmup.TimeDuration, after.depositWarmups[i].TimeDuration)
    uassert.Equal(t, warmup.NextWarmupTime, after.depositWarmups[i].NextWarmupTime)
    uassert.Equal(t, warmup.WarmupRatio, after.depositWarmups[i].WarmupRatio)
}
```

Every `Warmup` field is compared, not just the tier count — those records carry the reward ratio schedule, so a silent change there would move payouts.

Both are printed into the golden output too, which pins the ladder itself:

```
[EXPECTED] staker allowed token: gno.land/r/gnoswap/gns.GNS
[EXPECTED] staker allowed token: gno.land/r/gnoland/wugnot.wugnot
[EXPECTED] staker allowed token: gno.land/r/onbloc/bar.BAR
[EXPECTED] deposit warmup 0 duration / nextWarmupTime / ratio: 432000 1234999970 30
[EXPECTED] deposit warmup 1 duration / nextWarmupTime / ratio: 864000 1235863970 50
[EXPECTED] deposit warmup 2 duration / nextWarmupTime / ratio: 2592000 1238455970 70
[EXPECTED] deposit warmup 3 duration / nextWarmupTime / ratio: 9223372036854775807 9223372036854775807 100
```

`gno test ./gno.land/r/gnoswap/scenario/upgrade` → 38/38 pass.

junghoon-vans (approved):

PR #1409, opened 2026-08-26
refactor(pool): pay collectors directly instead of approving position

jinoosss:
## Summary

`pool.Collect` updated the pool's ledger but left the tokens in place, handing the position contract an ERC20-style allowance to pull them out afterwards. Only one of the two collect paths ever spent that allowance, so the other left an unspent approval sitting on the token ledger until the next `Collect` overwrote it.

Two commits, reviewable independently.

## Change 1 — withdraw the protocol fee approval when settlement fails

`settleProtocolFee` approves the protocol fee realm for the amount it is about to forward, then returns early if `AddToProtocolFee` reports an error. The amount is kept pending and retried later, but the approval stays behind — an unspent allowance from the paying realm to the protocol fee realm, live until some later settlement of the same token overwrites it.

The block is copy-pasted into **pool**, **staker** and **router**; all three now withdraw the approval on that branch.

The error branch is currently unreachable — `AddToProtocolFee` only fails for a halted realm, which `settleProtocolFee` checks first — so this is hardening against the guard above it changing, not a live leak. It ships first because it stands on its own.

## Change 2 — the pool pays collectors directly

`Collect` now transfers to `recipient` itself instead of approving the position contract, the way `UniswapV3Pool.collect` does:

```solidity
// UniswapV3Pool.collect
if (amount0 > 0) {
    position.tokensOwed0 -= amount0;
    TransferHelper.safeTransfer(token0, recipient, amount0);
}
```

The ledger write and the transfer now happen in one call under one pool lock. Previously they were split across two realm calls with the lock released in between, so the pool's internal accounting and its actual token balance were briefly out of step.

- `SafeGRC20Approve` in `Collect` — removed. No allowance towards the position contract is created at all any more.
- `SafeGRC20TransferFrom` in `position/v1/burn.gno` — removed. That path no longer touches tokens; the file lost its `common`, `access` and `prabc` imports as a result.
- `HandleWithdrawalFee` — removed, folded into the collect path.
- The `recipient` parameter, previously validated and then ignored, now does what its name says.

`Collect` pays the full amount; `CollectSwapFee` withholds the withdrawal fee and settles it to the protocol fee realm. Both share one internal `collect()`, so the cap, the CEI ordering and the lock discipline live in a single place.

Which entry point applies the withdrawal fee is the caller's choice, because the pool cannot tell accrued fees from returned principal — both land in the same `tokensOwed`, exactly as they do in Uniswap, where `Position.update()` credits fees and `burn()` credits principal into one field. The position contract keeps them apart by collecting fees before `Burn` credits any principal. Splitting the two in pool storage would let the pool decide on its own, but that means changing a persisted schema, so it is deliberately left out of this PR.

## Behaviour

Unchanged. A fee collection debits the pool by the gross amount, pays the collector the amount net of the withdrawal fee, and credits the protocol fee realm with the fee — same numbers before and after.

```
[SCENARIO] 2. CollectFee routes through pool.CollectSwapFee
[INFO] collected gross 14999, net 14850
[EXPECTED] BAR delta - alice 14850, pool -14999, position 0, protocol_fee 149
[EXPECTED] after CollectFee: pool->position allowance BAR 0, FOO 0

[SCENARIO] 3. DecreaseLiquidity routes through pool.Collect
[EXPECTED] BAR delta - alice 5635709, position 0
[EXPECTED] after DecreaseLiquidity: pool->position allowance BAR 0, FOO 0
```

## Tests

- `position_collect_no_allowance_filetest` — drives both position operations end to end, pinning the allowance at zero throughout and tracing where the money lands. Verified to fail (`allowance 14999` instead of `0`) against the unfixed code.
- `TestDeductWithdrawalFee` — fee arithmetic; the fee and the remainder must always sum back to the original amount.
- `TestCollectAccessControl` — both entry points remain position-only.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated swap-fee collection for positions, reporting collected amounts and withheld withdrawal fees.
  * Updated standard collection to pay owed tokens without withdrawal fees.
* **Bug Fixes**
  * Failed protocol-fee settlements now revoke token approvals, preventing lingering allowances.
  * Zero-rate swap-fee collection still settles pending protocol fees, while standard collection leaves them untouched.
  * Preserved access-control and swap-lock protections during fee collection.
* **Refactor**
  * Simplified liquidity-burn and fee-collection flows by consolidating withdrawal-fee handling.


contract/r/gnoswap/pool/v1/pool.gno
```diff
@@ -244,10 +319,8 @@ func (i *poolV1) Collect(
 		))
 	}
 
-	amount0 := minRequestedAmount(amount0Req, position.TokensOwed0())
-	amount1 := minRequestedAmount(amount1Req, position.TokensOwed1())
-
-	positionAddr := access.MustGetAddress(prabc.ROLE_POSITION.String())
+	amount0 = minRequestedAmount(amount0Req, position.TokensOwed0())
```
junghoon-vans, line 325: The approval calls below still use `positionAddr`, but this change removes its declaration, so the current head does not compile. Please remove the remaining `SafeGRC20Approve` calls (and the matching Position-side `TransferFrom`) rather than restoring `positionAddr`; this path is supposed to pay the recipient directly and leave the Pool allowance at zero. The remaining old helper call signatures also need to be migrated to the current RealmTeller API.

contract/r/gnoswap/pool/v1/pool.gno
```diff
@@ -278,7 +351,25 @@ func (i *poolV1) Collect(
 		panic(err)
 	}
 
-	return utils.FormatInt(amount0), utils.FormatInt(amount1)
+	// Effects are persisted above; everything below leaves the realm.
+	fee0, amount0AfterFee := deductWithdrawalFee(amount0, withdrawalFeeBPS)
+	fee1, amount1AfterFee := deductWithdrawalFee(amount1, withdrawalFeeBPS)
+
+	if withdrawalFeeBPS != ZeroBps {
```
junghoon-vans: This guard skips `settleProtocolFee` whenever the current withdrawal-fee rate is zero. A zero rate can follow a halted period that left a pending protocol fee; the next `CollectSwapFee` then bypasses settlement entirely and the backlog can remain indefinitely. Please invoke settlement on the fee-bearing path even when the calculated fee is zero, keep the fee-free `Collect` path separate, and add a regression test for non-zero fee -> halt/pending -> fee set to zero -> resumed collection.

junghoon-vans (commented):
I found two issues in the current head; details are inline.

contract/r/gnoswap/pool/v1/pool.gno
jinoosss: 5383702e6d985be2c939cdbe58abb300ebac6098

junghoon-vans (approved):

PR #1411, opened 2026-08-28
fix(gnsmath): check fallback denominator overflow

junghoon-vans:
## Summary

- use `AddOverflow` for the fallback denominator in `getNextPriceAmount0Add`
- panic with `errSafeMathOverflow` on overflow, matching Uniswap V3 `LowGasSafeMath.add` behavior
- add a regression test and document when critical `Add` operations require overflow handling

## Background

The previous fallback denominator used wrapping `uint256.Add`:

```text
(liquidity << 96) / sqrtPrice + amount
```

Uniswap V3 evaluates the equivalent expression with `LowGasSafeMath.add`, which reverts on `uint256` overflow. This change restores that behavior.

This is a defensive consistency fix, not a currently exploitable protocol path. In the production token0 exact-input path:

- `amountToAdd` originates from a positive `int256`, so it is less than `2^255`.
- active pool liquidity is constrained to `uint128`, so `liquidity << 96` and therefore its division by a non-zero sqrt price are less than `2^224`.
- consequently, the fallback denominator is less than `2^255 + 2^224`, which is below `2^256`.

The explicit check prevents silent divergence if these input bounds or invariants change, or if the helper is reused under wider ranges.

References:

- [Uniswap V3 `SqrtPriceMath.sol` fallback calculation](https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/SqrtPriceMath.sol#L34-L50)
- [Uniswap V3 `LowGasSafeMath.add(uint256,uint256)`](https://github.com/Uniswap/v3-core/blob/d0831dc6b8a318df3872b6d68f6de135c9f3ec29/contracts/libraries/LowGasSafeMath.sol#L5-L11)

## Verification

- `make test WORKDIR=tmp-fallback-add PKG=gno.land/p/gnoswap/gnsmath RUN='TestSqrtPriceMath_getNextPriceAmount0Add'`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented silent arithmetic overflow during fallback price calculations.
  * Added an explicit safety error when critical denominator calculations exceed supported limits.

* **Documentation**
  * Updated math guidance to require overflow-checked addition for values affecting pricing, accounting, state, or denominators.

* **Tests**
  * Added coverage verifying that denominator overflow is detected and reports the expected safety error.


jinoosss (approved):

contract/p/gnoswap/gnsmath/sqrt_price_math.gno
jinoosss, line 67: This is fine ?

notJoon (approved):

PR #1415, opened 2026-08-28
fix(staker): cache rewards before tick crosses

junghoon-vans:
## Summary

Fixes internal GNS reward accounting when a staked position becomes in-range by crossing a tick after one or more elapsed annual GNS halving boundaries.

`processTickCrossImmediate` and `processBatchedTickCrosses` now materialize the pool's reward-cache history **before** `modifyDeposit` changes staked liquidity. This closes and restarts an open zero-liquidity (`unclaimable`) interval at every elapsed halving boundary using the rate valid for each interval.

## Root cause

When a tiered pool had zero staked liquidity at a halving boundary `hv`, then a later tick cross changed liquidity from zero to positive:

1. The pre-fix tick-cross path called `modifyDeposit` first.
2. `modifyDeposit` closed the open unclaimable interval at the tick-cross timestamp.
3. Its reward rate was resolved from the interval start, before `hv`.
4. A later cache update replayed the halving after the interval was already closed.

The `hv -> tickCross` interval was therefore charged at the pre-halving rate. The excess was transferred to the community pool rather than directly to the crossing position, but it increased `TotalEmissionSent` and could drain GNS reserved for later staker rewards.

## Fix

- In the normal `Swap` path, tick crosses accumulate in the active batch. For each non-empty batch, `processBatchedTickCrosses` calls `cacheRewardForPool(timestamp, ...)` once before applying the batch's net liquidity delta, regardless of how many ticks it contains.
- The immediate path is a fallback for tick-cross callers without an active swap batch. It calls `cacheRewardForPool(timestamp, ...)` before `modifyDeposit` for each individual cross.
- The cache replay writes the halving checkpoint while pre-cross liquidity is still zero, so `unclaimable` is split at `hv`.

The fix preserves normal historical cache replay; it does not reject historical timestamps or replace the rate schedule with a special-case path.

## Before / after

The strengthened filetest keeps a tiered pool at zero staked liquidity through two halving boundaries, then uses a router swap to cross the position into range and calls `CollectReward`.

| Measurement | `main` (pre-fix) | PR head (fixed) | Difference |
| --- | ---: | ---: | ---: |
| Staker reward | 6,688,779 | 6,688,779 | 0 |
| Community unclaimable / penalty | 295,312,439,226,149 | 210,937,430,489,752 | 84,375,008,736,397 excess on `main` |
| Protocol fee | 0 | 0 | 0 |
| Total emission delta | 295,312,445,914,928 | 210,937,437,178,531 | 84,375,008,736,397 excess on `main` |

The difference is the two zero-liquidity intervals spanning the halving boundaries before the tick cross. Position reward and protocol fee remain unchanged.

### Production filetest

`halving_tick_cross_unclaimable_filetest.gno` exercises the production path: router swap, pool swap hooks, batched tick-cross processing, and `CollectReward`. It verifies that the pool remains at zero staked liquidity through the first and second halving, then asserts the complete internal GNS distribution:

```text
[SCENARIO] two-halving tick-cross unclaimable
[EXPECTED] first halving crossed
[EXPECTED] second halving crossed
[EXPECTED] swap crossed into staked range
[RESULT] internal GNS distribution:
  - Staker reward: 6688779
  - Community unclaimable/penalty: 210937430489752
  - Protocol fee: 0
  - Total emission delta: 210937437178531
[EXPECTED] recipients match emission accounting: true
```

The same filetest, temporarily applied to `main` at `b084fa89`, fails only its golden output: its community-pool and total-emission values are each 84,375,008,736,397 higher.

### Immediate fallback unit test

`TestProcessTickCrossImmediate_ReplaysHalvingBeforeDeposit` in `reward_calculation_tick_immediate_test.gno` isolates the currently unused immediate fallback. It starts a zero-liquidity pool at rate `100`, crosses a halving boundary to `50`, then crosses a tick at time `10`; the expected unclaimable reward is `650`. This keeps the fallback's `cacheRewardForPool → modifyDeposit` ordering covered without duplicating the production batch-path filetest.

## Verification

Passed on this PR head:

```sh
make test WORKDIR=/tmp/gnoswap-pr1415-review \
  PKG=gno.land/r/gnoswap/scenario/staker

make test WORKDIR=/tmp/gnoswap-pr1415-review \
  PKG=gno.land/r/gnoswap/staker/v1 \
  RUN=TestProcessTickCrossImmediate_ReplaysHalvingBeforeDeposit
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected reward calculation during tick crossings so halving boundaries are applied before liquidity changes.
  * Fixed reward accounting for immediate, unbatched tick crossings when liquidity is added after a reward-rate halving.

* **Tests**
  * Added coverage for reward calculations across halving boundaries before the first liquidity deposit.


jinoosss (approved):

contract/r/gnoswap/staker/v1/reward_calculation_tick.gno
```diff
@@ -156,15 +155,17 @@ func (s *stakerV1) tickCrossHook(_ int, rlm realm, poolPath string, tickId int32
...
 }
 
-// processTickCrossImmediate processes a single tick cross immediately
-// This function handles individual tick crosses for non-swap operations
-// where batch processing is not applicable (e.g., position modifications, liquidations)
+// processTickCrossImmediate processes one unbatched tick cross.
+// It materializes elapsed halving boundaries while the pre-cross liquidity is
+// still effective, before modifying the deposit for that individual cross.
 func (s *stakerV1) processTickCrossImmediate(pool *sr.Pool, tick *sr.Tick, tickId int32, zeroForOne bool, timestamp int64) {
+
```
notJoon: newly added new lines seems unnecessary
```suggestion
```

notJoon (approved):
left a minor suggestion. overall lgtm

PR #1425, opened 2026-09-01
refactor(common): remove redundant GRC20 wrappers

notJoon:
## Description

`grc20reg` now provides `Transfer`, `TransferFrom`, and `Approve` functions with the same interface and behavior as the wrappers previously defined in `common`.

This PR removes those redundant wrappers and updates callers to invoke the `grc20reg` functions directly, reducing unnecessary indirection and gas overhead.

The `SafeGRC20XXX` helpers were also removed. The corresponding `grc20reg` functions already panic when an operation fails, so they provide the required failure safety without a separate safe wrapper layer.


notJoon:
temporary close this PR to avoid confusion

PR #1428, opened 2026-09-02
refactor(governance): index active proposal snapshots

junghoon-vans:
## Summary

Preserve proposal and voting history with one snapshot-ordered active index and admin-only single-proposal maintenance.

## Behavior

- `GetOldestActiveProposalSnapshotTime() (int64, bool, error)` checks only the first index entry. A stale head reports `proposalID: N`; missing records also fail closed.
- `RemoveInactiveProposalFromIndexByAdmin(cur, proposalID)` removes only that inactive proposal's index entry. It rejects unauthorized callers and missing or still-active proposals.
- Successful admin maintenance emits `RemoveInactiveProposalFromIndex` with `prevAddr`, `prevRealm`, and `proposalId`; existing lifecycle event payloads are unchanged.
- Creation inserts index entries; cancel and execute remove them immediately. Other inactive entries are removed explicitly in separate transactions before retrying snapshot cleanup.
- Snapshot cleanup never maintains the index implicitly and still protects active proposal snapshots. Proposal and vote records remain permanent.

## Minimal diff

Against main `ed4e8f52dae3ef073ad0b1660b88af7a77e4df3a`: **20 files, +378 / -72**.

- No new Gno source or scenario files. Maintenance lives in the existing lifecycle file; index-key encoding is private; the existing proxy tree factory is reused.
- Removed the extra implementation file, upgrade symlink, and two standalone scenario files. `proposal.gno` and the staker snapshot implementation match main exactly.
- Kept the original snapshot scenario output and appended concise maintenance coverage. Getter edge cases share one small table; independent-transaction coverage remains in the txtar fixture.
- Removed error-declaration alignment churn and redundant comments/assertions. No changelog changes.

## Verification

Verified commit: `0c9b4c4b45048212dc939f75e157ad5be0c172e3`.
Merged latest main `ed4e8f52dae3ef073ad0b1660b88af7a77e4df3a`; retained explicit snapshot-index maintenance and added `common/v1` to the transaction fixture for the common implementation split.
Gno runtime: `393b6f92ac151e31916c96fbe30f6bcdc2f1dc01`, built in isolated local directories.

Formatting: ran `gofumpt` on all 18 PR-changed Gno files only; no remaining formatting differences.

All seven affected package suites passed without updating goldens:

```sh
make test WORKDIR=tmp/verification PKG='gno.land/r/gnoswap/gov/governance ./gno.land/r/gnoswap/gov/governance/v1 ./gno.land/r/gnoswap/gov/staker/v1 ./gno.land/r/gnoswap/scenario/getter ./gno.land/r/gnoswap/scenario/gov/governance ./gno.land/r/gnoswap/scenario/gov/staker ./gno.land/r/gnoswap/scenario/upgrade'
```

The local-node transaction regression also passed:

```sh
python3 setup.py -w tmp/transactions --exclude-tests
go -C tmp/transactions/gno test ./gno.land/pkg/integration -run '^TestTestdata$/^gov_governance_remove_inactive_proposal_from_index$' -v -count=1 -timeout=5m
```

Coverage: maintenance event emission and proposal ID (regression fails before the fix and passes afterward); stale/missing/tied snapshot ordering; admin and active-proposal guards; cancellation and expiration; retained delegation and vote weight; separate ID2/ID1/ID3 maintenance commits, intervening cleanup rejection, retained archive, and final cleanup success.

## Deployment notes

Initial deployment only: no migration/backfill for pre-existing unindexed archives. Include the matching governance proxy/implementation and staker changes together. Existing event formats are unchanged; no external-network deployment was performed.

Each getter/maintenance call examines at most one proposal, not a constant amount of gas. B+Tree costs remain, and delegation-history deletion is not batched by this change.

### DIFF
--- a/contract/r/gnoswap/gov/governance/proposal.gno
+++ b/contract/r/gnoswap/gov/governance/proposal.gno
@@ -1,8 +1,7 @@
 package governance
 
 import (
-	"strconv"
-
+	"gno.land/p/gnoswap/utils"
 	bptree "gno.land/p/nt/bptree/v0"
 )
 
@@ -84,8 +83,6 @@ func (p *Proposal) SnapshotTime() int64 {
 	return p.snapshotTime
 }
 
-const activeProposalIndexKeyPadding = "00000000000000000000"
-
 // ActiveProposalSnapshotKey returns this proposal's chronological snapshot index key.
 func (p *Proposal) ActiveProposalSnapshotKey() string {
 	return activeProposalIndexKey(p.snapshotTime, p.id)
@@ -96,13 +93,18 @@ func (p *Proposal) ActiveProposalDeadlineKey(deadline int64) string {
 	return activeProposalIndexKey(deadline, p.id)
 }
 
+// activeProposalIndexKey encodes the ordering timestamp (snapshot time or
+// lifecycle deadline) and proposal ID as fixed-width big-endian integers, so
+// B+Tree byte-ordering matches numeric ordering without decimal zero-padding.
 func activeProposalIndexKey(timestamp int64, proposalID int64) string {
-	timestampKey := strconv.FormatInt(timestamp, 10)
-	proposalIDKey := strconv.FormatInt(proposalID, 10)
+	return utils.EncodeUint64(uint64(timestamp)) + utils.EncodeUint64(uint64(proposalID))
+}
 
-	return activeProposalIndexKeyPadding[:20-len(timestampKey)] + timestampKey +
-		":" +
-		activeProposalIndexKeyPadding[:20-len(proposalIDKey)] + proposalIDKey
+// ActiveProposalIndexKeyDeadline decodes the timestamp a deadline-index key
+// was built with, so callers can recover the lifecycle stage a stored entry
+// currently represents without recomputing and comparing candidate keys.
+func ActiveProposalIndexKeyDeadline(key string) int64 {
+	return int64(utils.DecodeUint64(key[:8]))
 }
 
 // Data returns the proposal data.
--- a/contract/r/gnoswap/gov/governance/proposal.gno
+++ b/contract/r/gnoswap/gov/governance/proposal.gno
@@ -1,6 +1,10 @@
 package governance
 
-import bptree "gno.land/p/nt/bptree/v0"
+import (
+	"strconv"
+
+	bptree "gno.land/p/nt/bptree/v0"
+)
 
 // Proposal represents a governance proposal with all its associated data and state.
 // This is the core structure that tracks proposal lifecycle from creation to execution.
@@ -78,6 +82,27 @@ func (p *Proposal) ConfigVersion() int64 {
 // SnapshotTime returns the snapshot time for voting weight lookup.
 func (p *Proposal) SnapshotTime() int64 {
 	return p.snapshotTime
+}
+
+const activeProposalIndexKeyPadding = "00000000000000000000"
+
+// ActiveProposalSnapshotKey returns this proposal's chronological snapshot index key.
+func (p *Proposal) ActiveProposalSnapshotKey() string {
+	return activeProposalIndexKey(p.snapshotTime, p.id)
+}
+
+// ActiveProposalDeadlineKey returns this proposal's chronological lifecycle deadline index key.
+func (p *Proposal) ActiveProposalDeadlineKey(deadline int64) string {
+	return activeProposalIndexKey(deadline, p.id)
+}
+
+func activeProposalIndexKey(timestamp int64, proposalID int64) string {
+	timestampKey := strconv.FormatInt(timestamp, 10)
+	proposalIDKey := strconv.FormatInt(proposalID, 10)
+
+	return activeProposalIndexKeyPadding[:20-len(timestampKey)] + timestampKey +
+		":" +
+		activeProposalIndexKeyPadding[:20-len(proposalIDKey)] + proposalIDKey
 }
 
 // Data returns the proposal data.
@@ -173,6 +198,11 @@ func (p *Proposal) Clone() *Proposal {
 }
 
 func NewProposalTree() *bptree.BPTree {
+	return bptree.NewBPTreeN(16)
+}
+
+// NewActiveProposalTree creates the snapshot-time ordered index of active proposals.
+func NewActiveProposalTree() *bptree.BPTree {
 	return bptree.NewBPTreeN(16)
 }
 
--- a/contract/r/gnoswap/gov/governance/v1/active_proposal_index.gno
+++ b/contract/r/gnoswap/gov/governance/v1/active_proposal_index.gno
@@ -1,0 +1,19 @@
+package governance
+
+import (
+	"time"
+
+	"gno.land/r/gnoswap/access"
+)
+
+// PruneInactiveProposals removes ended proposals from the active proposal index.
+// Proposal records and their voting information remain available as permanent history.
+// Only gov/staker may call this before it cleans delegation snapshots.
+func (gv *governanceV1) PruneInactiveProposals(_ int, rlm realm) {
+	access.AssertIsRlmCurrent(0, rlm)
+	access.AssertIsGovStaker(rlm.Previous().Address())
+
+	if err := gv.pruneInactiveProposals(0, rlm, time.Now().Unix()); err != nil {
+		panic(err)
+	}
+}
--- a/contract/r/gnoswap/gov/governance/v1/active_proposal_index.gno
+++ b/contract/r/gnoswap/gov/governance/v1/active_proposal_index.gno
@@ -6,6 +6,12 @@ import (
 	"gno.land/r/gnoswap/access"
 )
 
+// maxProposalsPrunedPerCall bounds how many terminal proposals a single
+// PruneInactiveProposals call removes from the active index. This keeps the
+// call's gas cost predictable even if cleanup is neglected for a long time;
+// gov/staker calls this again on its next cleanup to keep making progress.
+const maxProposalsPrunedPerCall = 200
+
 // PruneInactiveProposals removes ended proposals from the active proposal index.
 // Proposal records and their voting information remain available as permanent history.
 // Only gov/staker may call this before it cleans delegation snapshots.
@@ -13,7 +19,7 @@ func (gv *governanceV1) PruneInactiveProposals(_ int, rlm realm) {
 	access.AssertIsRlmCurrent(0, rlm)
 	access.AssertIsGovStaker(rlm.Previous().Address())
 
-	if err := gv.pruneInactiveProposals(0, rlm, time.Now().Unix()); err != nil {
+	if err := gv.pruneInactiveProposals(0, rlm, time.Now().Unix(), maxProposalsPrunedPerCall); err != nil {
 		panic(err)
 	}
 }
--- a/contract/r/gnoswap/gov/governance/v1/governance_execute.gno
+++ b/contract/r/gnoswap/gov/governance/v1/governance_execute.gno
@@ -68,7 +68,14 @@ func (gv *governanceV1) Execute(_ int, rlm realm, proposalID int64) int64 {
 		panic(err)
 	}
 
-	if err := gv.removeActiveProposal(0, rlm, proposal); err != nil {
+	// Execute only succeeds once voting has ended, but pruneInactiveProposals
+	// may or may not have already rescheduled this entry from VotingEndTime
+	// to ExpiredTime, so both candidate deadline keys must be attempted.
+	if err := gv.removeActiveProposal(
+		0, rlm, proposal,
+		proposal.Status().Schedule().VotingEndTime(),
+		proposal.Status().Schedule().ExpiredTime(),
+	); err != nil {
 		panic(err)
 	}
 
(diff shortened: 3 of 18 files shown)


contract/r/gnoswap/gov/governance/v1/instance.gno
```diff
@@ -84,6 +84,22 @@ func (g *governanceV1) addProposal(_ int, rlm realm, proposal *governance.Propos
 		return false
 	}
 
+	snapshotIndex := g.store.GetActiveProposalsBySnapshot()
+	deadlineIndex := g.store.GetActiveProposalsByDeadline()
+
+	snapshotIndex.Set(proposal.ActiveProposalSnapshotKey(), proposal.ID())
+	deadlineIndex.Set(
+		proposal.ActiveProposalDeadlineKey(proposal.Status().Schedule().VotingEndTime()),
```
jinoosss, line 97: If the proposal is feasible, the deadline must be set as the ExpiredTime.

contract/r/gnoswap/gov/governance/v1/instance.gno
```diff
@@ -93,6 +109,80 @@ func (g *governanceV1) addProposal(_ int, rlm realm, proposal *governance.Propos
 	return true
 }
 
+func (g *governanceV1) removeActiveProposal(_ int, rlm realm, proposal *governance.Proposal) error {
+	snapshotIndex := g.store.GetActiveProposalsBySnapshot()
+	deadlineIndex := g.store.GetActiveProposalsByDeadline()
+
+	snapshotIndex.Remove(proposal.ActiveProposalSnapshotKey())
+	deadlineIndex.Remove(proposal.ActiveProposalDeadlineKey(proposal.Status().Schedule().VotingEndTime()))
+	deadlineIndex.Remove(proposal.ActiveProposalDeadlineKey(proposal.Status().Schedule().ExpiredTime()))
```
jinoosss: Why are two values being removed?

contract/r/gnoswap/gov/governance/v1/instance.gno
```diff
@@ -93,6 +109,80 @@ func (g *governanceV1) addProposal(_ int, rlm realm, proposal *governance.Propos
...
+		proposal, exists := g.getProposal(proposalID)
+		if !exists {
+			deadlineKeysToRemove = append(deadlineKeysToRemove, key)
+			return false
+		}
+
+		deadline := proposal.Status().Schedule().VotingEndTime()
+		if key == proposal.ActiveProposalDeadlineKey(proposal.Status().Schedule().ExpiredTime()) {
+			deadline = proposal.Status().Schedule().ExpiredTime()
+		}
```
jinoosss: It seems like a good idea to clearly define the deadline criteria.
Also, since deadlines are commonly used, consider implementing a receiver function for the `proposal` struct.

contract/r/gnoswap/gov/governance/proposal.gno
```diff
@@ -80,6 +84,27 @@ func (p *Proposal) SnapshotTime() int64 {
...
+func (p *Proposal) ActiveProposalDeadlineKey(deadline int64) string {
+	return activeProposalIndexKey(deadline, p.id)
+}
+
+func activeProposalIndexKey(timestamp int64, proposalID int64) string {
+	timestampKey := strconv.FormatInt(timestamp, 10)
+	proposalIDKey := strconv.FormatInt(proposalID, 10)
+
+	return activeProposalIndexKeyPadding[:20-len(timestampKey)] + timestampKey +
+		":" +
```
jinoosss: For timestamp key conversion, please refer to the function below.
https://github.com/gnoswap-labs/gnoswap/blob/main/contract/p/gnoswap/utils/codec.gno#L21-L48

contract/r/gnoswap/gov/governance/v1/instance.gno
junghoon-vans: Kept VotingEndTime as the initial deadline for every proposal type on purpose: whether an executable proposal later needs the ExpiredTime deadline depends on whether it actually passes, which isn't known until voting ends. Setting ExpiredTime up front for executable proposals that get rejected would leave them un-prunable until an ExpiredTime that never mattered. pruneInactiveProposals already reschedules to ExpiredTime once a proposal is confirmed to have passed. Added a comment explaining this in 9ef380bf.

contract/r/gnoswap/gov/governance/v1/instance.gno
junghoon-vans: Good catch — fixed in 9ef380bf. removeActiveProposal now takes the specific deadline(s) the caller can have stored: Cancel only ever writes a VotingEndTime key (cancel requires upcoming status, before any reschedule), so it now removes just that one. Execute can happen after pruneInactiveProposals has or hasn't already rescheduled the entry, so it still removes both VotingEndTime and ExpiredTime candidates.

contract/r/gnoswap/gov/governance/v1/instance.gno
junghoon-vans: Addressed in 9ef380bf. Switched the active-index keys to a fixed-width big-endian uint64 encoding, so the deadline can be decoded directly from the key (ActiveProposalIndexKeyDeadline) instead of recomputing and string-comparing candidate keys here.

contract/r/gnoswap/gov/governance/proposal.gno
junghoon-vans: Switched to utils.EncodeUint64/DecodeUint64 in 9ef380bf, matching the raw-uint64-key approach. Removed the manual zero-padded decimal encoding entirely.

jinoosss (commented):
In my opinion, a single function to handle inactive proposals on a case-by-case basis should suffice. Since it is an administrative function, we can simply identify the problematic proposals and delete them.

contract/r/gnoswap/gov/governance/v1/active_proposal_index.gno
```diff
@@ -0,0 +1,25 @@
...
+// PruneInactiveProposals removes ended proposals from the active proposal index.
+// Proposal records and their voting information remain available as permanent history.
+// Only gov/staker may call this before it cleans delegation snapshots.
+func (gv *governanceV1) PruneInactiveProposals(_ int, rlm realm) {
+	access.AssertIsRlmCurrent(0, rlm)
+	access.AssertIsGovStaker(rlm.Previous().Address())
+
+	if err := gv.pruneInactiveProposals(0, rlm, time.Now().Unix(), maxProposalsPrunedPerCall); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
junghoon-vans: The batch-pruning API and implicit pruning during snapshot cleanup have been removed. Its replacement now emits `RemoveInactiveProposalFromIndex` after successful single-proposal admin maintenance, with `prevAddr`, `prevRealm`, and `proposalId` (1d4ec008). The real-node transaction regression checks the event and proposal ID; it failed before the fix and passes now. This addresses the observability concern without restoring batch pruning.

contract/r/gnoswap/gov/governance/v1/init.gno
```diff
@@ -66,6 +66,18 @@ func initStoreData(_ int, rlm realm, governanceStore governance.IGovernanceStore
 		}
 	}
 
+	if !governanceStore.HasActiveProposalsBySnapshotStoreKey() {
+		if err := governanceStore.SetActiveProposalsBySnapshot(0, rlm, governance.NewActiveProposalTree()); err != nil {
```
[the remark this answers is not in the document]
junghoon-vans: This is a valid concern for an upgrade with pre-existing proposal archives, but the confirmed scope is an initial deployment. Migration/backfill is intentionally out of scope; stores containing unindexed proposal archives are not supported. Made that deployment restriction explicit in `docs/gov.md` (1d4ec008). Resolving on that deployment scope, not claiming that legacy archives are migrated.

contract/r/scenario/upgrade/implements/v2_invalid/gov/governance/test_impl.gno
```diff
@@ -80,6 +80,10 @@ func (t *TestGovernance) Cancel(_ int, rlm realm, proposalId int64) int64 {
 	return t.instance.Cancel(0, rlm, proposalId)
 }
 
+func (t *TestGovernance) PruneInactiveProposals(_ int, rlm realm) {
+	t.instance.PruneInactiveProposals(0, rlm)
+}
```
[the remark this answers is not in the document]
junghoon-vans: Already addressed in the current implementation: `RemoveInactiveProposalFromIndexByAdmin` checks `isActive("RemoveInactiveProposalFromIndexByAdmin")` and panics before forwarding when disabled. `PruneInactiveProposals` and implicit pruning from snapshot cleanup no longer exist. The upgrade scenario suite passes on the current head.

contract/r/gnoswap/gov/governance/v1/active_proposal_index.gno
```diff
@@ -0,0 +1,26 @@
...
+	access.AssertIsAdmin(rlm.Previous().Address())
+
+	proposal, exists := gv.getProposal(proposalID)
+	if !exists {
+		panic(makeErrorWithDetails(errProposalNotFound, ufmt.Sprintf("proposalID: %d", proposalID)))
+	}
+	if NewProposalResolver(proposal).IsActive(time.Now().Unix()) {
+		panic(makeErrorWithDetails(errInvalidInput, ufmt.Sprintf("proposal %d is still active", proposalID)))
+	}
+	if err := gv.removeActiveProposal(0, rlm, proposal); err != nil {
```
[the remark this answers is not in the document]
junghoon-vans: Fixed in 1d4ec008. After the index write succeeds, the admin method emits `RemoveInactiveProposalFromIndex` with the standard `prevAddr`, `prevRealm`, and `proposalId` attributes. Existing lifecycle events are unchanged. Added an assertion to the existing real-node transaction fixture: missing emission fails before the fix, and the event with proposal ID 2 passes afterward. All seven affected package suites also pass.

contract/r/scenario/upgrade/implements/v3_valid/gov/governance/test_impl.gno
```diff
@@ -49,6 +49,10 @@ func (t *TestGovernance) Cancel(_ int, rlm realm, proposalId int64) int64 {
 	return t.instance.Cancel(0, rlm, proposalId)
 }
 
+func (t *TestGovernance) RemoveInactiveProposalFromIndexByAdmin(_ int, rlm realm, proposalID int64) {
```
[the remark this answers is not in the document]
junghoon-vans: Added the documentation comment for the exported wrapper in 1d4ec008. Ran `gofumpt` on all 18 PR-changed Gno files only, with no remaining formatting differences. The upgrade suite and the other six affected package suites pass.

contract/r/gnoswap/gov/governance/v1/governance_execute.gno
```diff
@@ -186,6 +193,32 @@ func (gv *governanceV1) Cancel(_ int, rlm realm, proposalID int64) int64 {
 	return proposal.ID()
 }
 
+// RemoveInactiveProposalFromIndexByAdmin removes one inactive proposal from the
+// snapshot index without deleting its proposal or voting history.
+func (gv *governanceV1) RemoveInactiveProposalFromIndexByAdmin(_ int, rlm realm, proposalID int64) {
+	access.AssertIsRlmCurrent(0, rlm)
+	access.AssertIsAdmin(rlm.Previous().Address())
```
jinoosss: There doesn't seem to be a need to restrict admin privileges.

junghoon-vans: [a0fb67b](https://github.com/gnoswap-labs/gnoswap/pull/1428/commits/a0fb67b24798f6302fcbe4de3c62198509ba543c)

contract/r/gnoswap/gov/governance/v1/instance.gno
```diff
@@ -93,6 +105,14 @@ func (g *governanceV1) addProposal(_ int, rlm realm, proposal *governance.Propos
 	return true
 }
 
+// removeActiveProposal removes a proposal from the active snapshot index.
+func (g *governanceV1) removeActiveProposal(_ int, rlm realm, proposal *governance.Proposal) error {
+	snapshotIndex := g.store.GetActiveProposalsBySnapshot()
```
jinoosss, line 110: It appears necessary to check whether the key exists (i.e., whether the index has already been removed).

contract/r/gnoswap/gov/governance/v1/governance_execute.gno
jinoosss: Check halt

junghoon-vans: [0c87011](https://github.com/gnoswap-labs/gnoswap/pull/1428/commits/0c870111b6628525a37cbec0e32abc0be50d9b3b)

contract/r/gnoswap/gov/governance/v1/instance.gno
junghoon-vans: [0c87011](https://github.com/gnoswap-labs/gnoswap/pull/1428/commits/0c870111b6628525a37cbec0e32abc0be50d9b3b)

jinoosss (approved):

PR #1429, opened 2026-09-02
fix(staker): decouple unstaking from reward payout

jinoosss:
## Problem

`UnStakeToken` called `CollectReward` inline, so every failure mode of the payout half also
blocked withdrawal:

- a drained staker GNS reserve made `gns.Transfer` abort ([staker.gno#L671](https://github.com/gnoswap-labs/gnoswap/blob/main/contract/r/gnoswap/staker/v1/staker.gno#L671))
- a reward token whose transfer reverts did the same for external rewards ([staker.gno#L561](https://github.com/gnoswap-labs/gnoswap/blob/main/contract/r/gnoswap/staker/v1/staker.gno#L561))

Either case left the position NFT locked in the staker. A panic crossing a realm boundary
cannot be recovered from in Gno, so the payout cannot simply be wrapped — the accounting and
the transfer have to be separated.

## Change

Reward handling is split into settlement and payout:

- **`settlePositionReward`** does the accounting only — stops accrual, advances the
  last-collect cursors, draws the amounts down from the incentives, and credits what is owed
  to a new claimable reward ledger. It performs no token transfer and no reward-token
  cross-call, so it cannot revert on anything a token realm or a short reserve does.
- **`payoutClaimableRewards`** drains the ledger one token at a time. A token the reserve
  cannot cover stays claimable and is reported through an `InsufficientClaimableReward`
  event instead of aborting the payout.
- **`CollectReward`** = settle + payout, so the collect path is unchanged for users.
- **`UnStakeToken`** settles only, and no longer calls `MintAndDistributeGns`. The NFT always
  comes back; rewards are withdrawn separately.

New entry points:

| Function | Purpose |
|---|---|
| `ClaimRewards(cur)` | Pays out the caller's settled-but-unpaid rewards |
| `ClaimRewardsFor(cur, receiver)` | Same payout for another receiver. Permissionless — funds only ever reach the ledger owner — so the community pool's share (internal penalties, unclaimable rewards) can be flushed after unstake-only flows |
| `GetClaimableRewardsOf(receiver)` / `GetClaimableReward(receiver, tokenPath)` | Views over the ledger |

Protocol fees follow the existing pending-fee idiom: `accrueStakingRewardFee` records the fee
at settlement, and `settleProtocolFee` forwards it during payout — only when the reserve
covers both the fee and the receiver's amount.

## Behaviour change

`UnStakeToken` no longer transfers rewards. It is a 1-tx → 2-tx change for anyone who relied
on unstaking to also pay out. Everything else about the amounts is identical.

## Verification

- `contract/r/gnoswap/staker`, `staker/v1`, `pool/v1`, `position/v1`, `router/v1`,
  `gov/staker/v1`, `gov/governance/v1`, `protocol_fee/v1`, `gnft` — all pass
- All `contract/r/scenario/*` suites pass
- The six scenario filetests that measured balance deltas across `UnStakeToken` now claim
  explicitly at the same point. **Their golden outputs are unchanged**, which shows the split
  preserves the amounts and only moves when they are paid.
- New coverage: `claimable_reward_test.gno` (ledger, payout, short-reserve skip) and
  `unstake_with_drained_reward_reserve_filetest.gno` — drains the staker's GNS reserve, then
  unstakes and gets the NFT back, with the reward settled as claimable and paid once the
  reserve is restored.

### DIFF
--- a/contract/r/gnoswap/staker/v1/claimable_reward.gno
+++ b/contract/r/gnoswap/staker/v1/claimable_reward.gno
@@ -1,0 +1,252 @@
+package staker
+
+import (
+	"chain"
+	"chain/runtime"
+	"time"
+
+	bptree "gno.land/p/nt/bptree/v0"
+	ufmt "gno.land/p/nt/ufmt/v0"
+
+	"gno.land/p/gnoswap/gnsmath"
+	prbac "gno.land/p/gnoswap/rbac"
+	"gno.land/p/gnoswap/utils"
+
+	"gno.land/r/gnoswap/access"
+	"gno.land/r/gnoswap/common"
+	en "gno.land/r/gnoswap/emission"
+	"gno.land/r/gnoswap/halt"
+	sr "gno.land/r/gnoswap/staker"
+)
+
+// claimableRewardKeySeparator separates the receiver address from the token path in a
+// ledger key. Neither an address nor a token path can contain it, so a key round-trips
+// unambiguously and a receiver's entries form a contiguous key range.
+const claimableRewardKeySeparator = "\x00"
+
+// claimableRewardRangeEnd terminates a receiver's key range. Iterate treats end as
+// exclusive, and "\x01" is the smallest string that sorts after every key prefixed with
+// the receiver plus the separator.
+const claimableRewardRangeEnd = "\x01"
+
+// ClaimableRewards is the ledger of settled-but-not-yet-paid reward amounts, keyed by
+// (receiver, token path).
+//
+// Reward settlement credits this ledger instead of transferring tokens, so the accounting
+// half of a collect can never be blocked by the payout half. Payout drains the ledger in a
+// separate step (CollectReward's tail, or ClaimRewards), and anything that cannot be paid
+// right now stays here until it can.
+type ClaimableRewards struct {
+	tree *bptree.BPTree
+}
+
+// NewClaimableRewards creates an empty claimable reward ledger.
+func NewClaimableRewards() *ClaimableRewards {
+	return &ClaimableRewards{
+		tree: sr.NewBPTreeN(16), // receiver|tokenPath -> int64
+	}
+}
+
+// claimableRewardEntry is one ledger row.
+type claimableRewardEntry struct {
+	tokenPath string
+	amount    int64
+}
+
+// claimableRewardKey builds the ledger key for a (receiver, token path) pair.
+func claimableRewardKey(receiver address, tokenPath string) string {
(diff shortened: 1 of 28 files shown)


contract/r/gnoswap/staker/v1/claimable_reward.gno
```diff
@@ -0,0 +1,252 @@
...
+
+			continue
+		}
+
+		// Effects before interactions: the row is cleared before anything leaves this
+		// realm, so a re-entrant claim cannot see it twice.
+		claimable.remove(receiver, entry.tokenPath)
+
+		pendingProtocolFee := s.store.GetPendingProtocolFees()[entry.tokenPath]
+		if pendingProtocolFee > 0 && balance >= gnsmath.SafeAddInt64(pendingProtocolFee, entry.amount) {
```
junghoon-vans, line 181: Blocker: no

This condition only retries the pending fee while an existing claimable row for the same token is being paid. If the reserve is in the range `entry.amount <= balance < entry.amount + pendingProtocolFee`, the user row is removed and the user amount is transferred, but the pending fee remains. A later `ClaimRewards` call has no row for this token, so this branch is never reached again; `accrueStakingRewardFee` only adds to the pending map. The protocol fee can therefore remain permanently stranded. Please add an independent pending-fee flush path (or iterate pending-fee tokens separately) and cover the underfunded-fee case with a test.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -445,11 +455,35 @@ func (s *stakerV1) CollectReward(_ int, rlm realm, positionId uint64) (string, s
 	caller := rlm.Previous().Address()
 	assertIsDepositor(s, caller, positionId)
 
+	en.MintAndDistributeGns(cross(rlm))
+
+	settlement := s.settlePositionReward(0, rlm, positionId)
+
+	communityPoolAddr := access.MustGetAddress(prbac.ROLE_COMMUNITY_POOL.String())
+	s.payoutClaimableRewards(0, rlm, settlement.owner)
```
junghoon-vans, line 463: Blocker: no

`CollectReward` ignores the result of `payoutClaimableRewards` and returns `settlement.*` below. When the reserve is short, payout intentionally skips the row and leaves it in the ledger, but this method still reports the settled amount as if it were paid. That changes the observable return semantics and can make callers credit a balance the user has not received. Please return the actual payout result, or rename/document these values as settled rather than paid, and add a drained-reserve test for the return values.

PR #1432, opened 2026-09-03
feat(staker): defer unstaked reward collection to an exit checkpoint

jinoosss:
## Problem

`UnStakeToken` collected the position inline, so the whole reward calculation had to fit in the
withdrawal transaction. That calculation is not bounded: it grows with the unclaimed window, the
emission-rate segments crossed, and the number of external incentives on the pool. When it stops
fitting, the position cannot be withdrawn at all — and splitting the payout off the collect does
not help, because the cost is in the calculation, not the transfer.

Measured on this branch's stress scenario (24 never-collected incentives on one pool), the
inline equivalent costs **242.1M gas in the single mandatory withdrawal transaction**, growing
~9.9M per incentive.

## Change

Unstaking records an **exit checkpoint** instead of collecting.

`UnStakeToken` now does only what withdrawal actually requires: update the pool and tick
accounting, hand the NFT back, and write a checkpoint. No reward is calculated and none is paid.
On the same stress scenario it costs **17.3M gas** (~235k per pending incentive of pure
bookkeeping), and each later per-source collect is individually bounded.

### The checkpoint pins what a later collect must reproduce

`UnstakedPosition` (in `/r/gnoswap/staker`) holds the deposit, the exit timestamp, and the
pieces of state a later collect would otherwise re-derive from a pool that has moved on:

| Pinned | Why |
|---|---|
| `exitTick` | The unstake itself may append a historical tick entry at the exit timestamp |
| `lowerTick` / `upperTick` | The unstake may prune them from the pool; holding the objects keeps their outside-accumulation history reachable |
| `lowerOutsideAcc` / `upperOutsideAcc` | Entries before the exit can no longer change, but the one *at* the exit timestamp can be overwritten by a tick cross later in the same block |
| `tier` / `tierRatio` / `tierCount` | `appendInternalRewardTailSegments` reads these live, so a tier change after the exit would re-rate the already-closed window |
| `unstakingFee` | The window closed under the rate in force at exit; collect is permissionless, so a later rate change must not apply retroactively |

`PoolResolver` reads through that pin (`GetOrNewTick`, `CurrentTick`, the outside
accumulations and the internal reward-rate schedule) only on the checkpoint path; the staked
path is untouched.

The pool teardown in `applyUnStake` is unchanged — including the tick pruning. The checkpoint
simply keeps a reference to the tick objects it needs.

### Collecting is unchanged for callers, and per source

`CollectReward`, `CollectEmissionReward` and `CollectExternalIncentiveReward` all take a
position that was unstaked without collecting, as well as a staked one. Each collects only its
own source; the checkpoint tracks what is left (`emissionCollected`, `pendingIncentiveIds`) and
is dropped once nothing is. So a position with many incentives can be drained one transaction
per source, which is the point of the change.

A collect on a checkpoint is **permissionless**: it can only ever pay `deposit.Owner()`, and an
open entry point is what keeps the incentive guard below from being stuck behind an absent
staker.

A checkpoint's window is final. A zero user reward is left pending only while the position is
staked and still accruing; on a checkpoint the delivery still pays the penalty, advances the
cursor and marks the source collected.

### Guards

- `StakeToken` refuses a position that still has a checkpoint (`[GNOSWAP-STAKER-024]`).
  Collecting it there would smuggle the same unbounded calculation into staking, surfacing as
  out-of-gas rather than an error that names the fix; leaving the checkpoint would let the next
  unstake overwrite it and drop what it owes.
- `EndExternalIncentive` refuses while an unstaked position still owes a reward from **that**
  incentive. That share has not been drawn down from the incentive yet, so refunding first
  would pay it to the creator. The guard is keyed by incentive id, never by pool — a pool-wide
  key would let one blocked incentive take every other incentive in the pool down with it.
  Anyone can clear the block by collecting the position.
- An incentive's reward amount only ever decreases, so a checkpoint owed more than the
  incentive still holds can never be paid. Such a debt is **forfeited** on collect
  (`ForfeitUncollectedIncentiveReward` event), releasing both the position's re-staking and the
  incentive's refund; keeping it pending would lock both forever with no escape path.
- A checkpoint delivery refuses when its per-source cursor already reached the window's close.
  Collect computes amounts once and then transfers per source, and checkpoint collects are
  permissionless, so a re-entering call between transfers must not get a precomputed amount
  replayed.
- Pool-global accumulators (`processUnClaimableReward`, reward-cache materialization) always
  run at the current block time, never at the checkpoint's exit time: rewinding them would
  redirect other positions' rewards through the unclaimable path.

### Observability

`HasUnstakedPosition`, `GetUnstakedPositionExitTime`, `GetUnstakedPositionPendingIncentives`
and `GetUncollectedIncentiveCount` expose what a checkpoint still owes, so an integrator can
tell why a stake or an incentive refund is refused and which collects clear it.

## Behaviour change

Withdrawing is now `UnStakeToken` plus a collect, instead of one call. A position must be
collected before it can be staked again.

## Verification

- `staker`, `staker/v1`, `pool/v1`, `position/v1`, `router/v1`, `protocol_fee/v1`,
  `gov/staker/v1`, `gov/governance/v1`, `launchpad/v1`, `gnft`, `emission` and all
  `scenario/*` suites pass. Fuzz and integration suites were left to the author.
- The scenario filetests that measured reward deltas across `UnStakeToken` now collect
  explicitly at the same point. **Their golden outputs are unchanged**, which is what shows the
  deferral preserves the amounts and only moves when they are paid.

New scenario filetests cover the deferral itself:

| Filetest | Asserts |
|---|---|
| `unstake_collect_later` | A collect deferred past tick pruning, tick re-creation, a tier-layout change and 500 blocks pays exactly what an immediate one did. Removing either pin makes it fail — both were checked. |
| `unstake_collect_per_source` | Emission and incentive collect independently, the checkpoint clears only when both are done, and staking is refused until then. |
| `end_external_incentive_needs_collect` | The refund is refused while a position is uncollected, a third party can clear it, and the reward still goes to the staker. |
| `end_external_incentive_per_incentive` | One pending incentive blocks only its own refund; an unrelated incentive on the same pool can still end. |
| `checkpoint_collect_keeps_unclaimable` | A checkpoint collect at a past exit time leaves the pool-global unclaimable accounting of other positions untouched. |
| `checkpoint_collect_delivers_penalty` | A checkpoint whose user reward rounds to zero still delivers the penalty and completes, instead of staying pending forever. |
| `unstake_many_incentives_stress` | With 24 never-collected incentives: unstake stays a bounded bookkeeping write, emission plus 24 per-source collects drain every source, the checkpoint clears, and the position re-stakes. |

## Notes for review

- The pool-tier pin was found by testing, not by reading: a `SetPoolTier` between the unstake
  and the collect changed the deferred amount by exactly the tier-ratio factor
  (`2,675,513` → `1,783,675` per-second rate on the same window). `unstake_collect_later` now
  covers it.
- `UnstakedPosition.pendingIncentiveIds` is allocated inside `/r/gnoswap/staker`, not by the
  caller: a map built in the implementation realm is readonly tainted when the domain type
  mutates it.
- Checkpoints and the uncollected-incentive counts are stored in BP-trees under two new store
  keys, following the existing store conventions; value types live in the domain package.
- The deferred collect depends on one invariant, now recorded in `docs/staker.md`: every path
  that changes the tier layout must materialize the reward cache of all tiered pools first, at
  the current time (`changeTier` does this via `cacheReward`).

## Summary by CodeRabbit

* **New Features**
  * Unstaking now records an exit checkpoint, allowing rewards to be collected afterward.
  * Emission and external incentive rewards can be collected separately from unstaking, including permissionless collection for checkpointed positions.
  * Added visibility into unstaked positions, exit times, pending incentives, and uncollected rewards.
  * Administrators can deny external reward tokens from new incentives, while default tokens remain protected.

* **Bug Fixes**
  * Prevented staking or incentive closure when rewards remain uncollected.
  * Unpayable checkpoint rewards are safely forfeited, and insufficient balances no longer advance collection.

* **Documentation**
  * Updated staking guidance, usage examples, and failure scenarios.


contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -472,15 +478,106 @@ func (s *stakerV1) newCollectContext(_ int, rlm realm, positionId uint64, curren
...
+	s.getDeposits().set(self.positionId, self.deposit)
+}
+
+// markEmissionCollected records the emission collect on the checkpoint.
+func (s *stakerV1) markEmissionCollected(target *collectTarget) {
+	if !target.isCheckpoint() {
+		return
+	}
+
+	target.checkpoint.MarkEmissionCollected()
```
junghoon-vans, line 503: **Blocking:** `markEmissionCollected()` is unconditional, but `deliverEmissionReward` deliberately does not advance the internal cursor when `reward.Internal == 0` (including the `reward.InternalPenalty > 0` case). On an exit checkpoint this marks the emission source as settled and can drop the checkpoint even though the reward/penalty was not delivered, so a later collect cannot recover it. Please return an explicit settlement/cursor-advanced result from delivery and mark the source collected only after successful settlement; add a checkpoint regression for `reward == 0, penalty > 0`.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -472,15 +478,106 @@ func (s *stakerV1) newCollectContext(_ int, rlm realm, positionId uint64, curren
...
+
+// markIncentiveCollected records the incentive collect on the checkpoint. An incentive that
+// could not be funded stays pending, the same way it stays collectible while staked; delivery
+// reports that as an empty reward token against a non-zero amount.
+func (s *stakerV1) markIncentiveCollected(_ int, rlm realm, target *collectTarget, incentiveId string, rewardAmount int64, rewardToken string) {
+	if !target.isCheckpoint() || !target.checkpoint.HasPendingIncentiveId(incentiveId) {
+		return
+	}
+
+	if rewardToken == "" && rewardAmount > 0 {
```
junghoon-vans: **Blocking:** This condition only keeps the checkpoint pending when `rewardAmount > 0`. `deliverExternalIncentiveReward` returns an empty token and does not advance the incentive cursor whenever `rewardAmount == 0`, including when `externalPenalty > 0`. This branch then marks the incentive collected anyway; if it is the last pending source, the checkpoint/counter are cleared and the still-owed accrual can later be refunded by `EndExternalIncentive`. Please base pending removal on an explicit successful settlement/cursor-advancement result and cover the zero-user-reward/positive-penalty case.

contract/r/gnoswap/staker/v1/unstaked_position.gno
```diff
@@ -0,0 +1,157 @@
...
+	tier, tierRatio, tierCount, err := s.getPoolTier().tierContextOf(poolPath)
+	if err != nil {
+		panic(err)
+	}
+
+	position := sr.NewUnstakedPosition(
+		deposit,
+		exitTime,
+		poolResolver.CurrentTick(exitTime),
+		poolResolver.GetOrNewTick(deposit.TickLower()),
```
junghoon-vans: **Major / blocking for exit-state integrity:** These expressions store the live `Tick` pointers returned by the pool directly in the checkpoint. `applyUnStake` and later swap/cross paths can mutate the same objects; when a boundary is still shared, a same-timestamp or ordering-sensitive write can be included by `CurrentOutsideAccumulation(exitTime)`, so the checkpoint is not an immutable exit-time snapshot. The current pruning/recreation scenario does not cover a still-shared boundary tick. Please deep-copy the tick state (or persist an immutable exit-time snapshot) and add that shared-boundary regression.

contract/r/gnoswap/staker/v1/assert.gno
```diff
@@ -65,22 +65,46 @@ func assertIsDepositor(s *stakerV1, caller address, positionId uint64) {
...
+		panic(makeErrorWithDetails(
+			errUncollectedPosition,
+			ufmt.Sprintf("positionId(%d) still owes rewards from its previous stake; collect it first", positionId),
+		))
+	}
+}
+
+// assertHasNoUnsettledIncentive ensures no unstaked position of the pool still owes an
+// incentive reward, which has not been drawn down from the incentive yet.
+func assertHasNoUnsettledIncentive(s *stakerV1, poolPath string) {
```
junghoon-vans: **Blocking:** `unsettledIncentiveCounts` is keyed only by `poolPath`, but `EndExternalIncentive` receives a specific `incentiveId`. A checkpoint with an unsettled reward from incentive A increments the pool-wide count; that count then rejects ending/refunding an unrelated incentive B in the same pool, even after B itself is fully settled or began after the checkpoint exited. If A cannot be funded, this can indefinitely lock refunds for every other incentive in the pool. Track unsettled checkpoints by `(poolPath, incentiveId)` and have `EndExternalIncentive` check only the target incentive; add a two-incentive regression proving A pending blocks A only, while B can end.

jinoosss: 587eccabf63df866d55aa52aa487abbeb5dd8f3b

contract/r/gnoswap/staker/v1/unstaked_position.gno
```diff
@@ -0,0 +1,178 @@
...
+// The count is per incentive, not per pool: one checkpoint that can never be funded would
+// otherwise lock the refunds of every other incentive in the same pool.
+func (s *stakerV1) uncollectedIncentiveCountOf(incentiveId string) int64 {
+	return s.store.GetUncollectedIncentiveCounts()[incentiveId]
+}
+
+// addUncollectedIncentiveCounts applies one delta per incentive in a single store write.
+func (s *stakerV1) addUncollectedIncentiveCounts(_ int, rlm realm, deltas map[string]int64) {
+	counts := make(map[string]int64)
+	for incentiveId, count := range s.store.GetUncollectedIncentiveCounts() {
```
junghoon-vans: **Major scalability issue:** Although the guard is now keyed by incentive ID, every increment/decrement copies the entire global map and writes it back. With K distinct incentives, every checkpoint creation/source collection costs O(K) even when touching one ID, so unrelated incentives can reintroduce the same gas/storage growth. Please use per-key persistent storage, a BPTree, or another bounded update representation, and add a multi-incentive scale test.

contract/r/gnoswap/staker/v1/calculate_pool_position_reward.gno
```diff
@@ -96,15 +112,15 @@ type positionRewardContext struct {
...
+	deposit := param.deposit()
 	poolPath := deposit.TargetPoolPath()
 
 	pool, poolExisted := param.Pools.Get(poolPath)
 	if !poolExisted {
 		// Read-only: use an ephemeral pool; persistence is deferred to updatePositionReward.
 		pool = sr.NewPool(poolPath, param.CurrentTime)
 	}
-	poolResolver := NewPoolResolver(pool)
+	poolResolver := newPoolResolverWithExit(pool, param.Exit)
```
junghoon-vans, line 123: **Blocking scalability gap:** Adding an exit-specific resolver does not bound the work performed by the checkpoint path. External collection still discovers every incentive in the pool before computing one requested source, `UnStakeToken` still enumerates incentives up to exit, and internal collection still walks historical reward-cache entries. The change moves the worst-case cost out of withdrawal but does not prevent enough incentives/history from making unstake or the first collect exceed gas. Please add incremental discovery/reward cursors or materialize a bounded snapshot, plus a stress regression.

jinoosss: 8a2d1dc9fa445720580a30ac96f1ba93552c2bc6

contract/r/gnoswap/staker/v1/unstaked_position.gno
junghoon-vans: Confirmed fixed — `uncollectedIncentiveCounts` is now backed by `*bptree.BPTree` with per-key `Get`/`Set` (store.gno:422-441), not a full-map copy. Resolving.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -467,20 +470,136 @@ func (s *stakerV1) newCollectContext(_ int, rlm realm, positionId uint64, curren
...
+// decreases, so waiting cannot help, and a checkpoint pending it forever would lock both this
+// position's re-staking and the incentive's own refund.
+func (s *stakerV1) markIncentiveCollected(_ int, rlm realm, target *collectTarget, incentiveId string, rewardAmount, rewardPenalty int64, rewardToken string) {
+	if !target.isCheckpoint() || !target.checkpoint.HasPendingIncentiveId(incentiveId) {
+		return
+	}
+
+	if rewardToken == "" && (rewardAmount > 0 || rewardPenalty > 0) {
+		incentiveResolver := NewExternalIncentiveResolver(s.getExternalIncentives().get(incentiveId))
+		if incentiveResolver.RewardAmount() >= gnsmath.SafeAddInt64(rewardAmount, rewardPenalty) {
```
junghoon-vans, line 519: **Blocking:** This 'stays pending' check (`RewardAmount() >= rewardAmount+rewardPenalty`) is unreachable from any real caller. `CollectExternalIncentiveReward`/`CollectReward` only reach this line after `deliverExternalIncentiveReward` already skipped delivery because `RewardAmount() < total` on the exact same incentive state (line ~837-855) — so this line re-evaluates the negation of a condition that was just true, and it can never hold in production. The only test that reaches the pending branch (`TestMarkIncentiveCollectedForfeitsUncoverableIncentive`) calls this function directly with a hand-picked (rewardAmount, rewardPenalty) pair no real delivery outcome would produce; no scenario test exercises this decision through `CollectExternalIncentiveReward`.

Since `ExternalIncentive.RewardAmount()` only decreases (no top-up path), an underfunded claim genuinely cannot be recovered later, so this isn't a fund-loss bug — but the code and its doc comment describe a two-outcome design ('stays pending' vs forfeit) the implementation never delivers: every underfunded checkpoint collect forfeits unconditionally on the first, permissionless attempt. Please either drop the dead pending branch and update the doc comment to say collection is always terminal here, or add a scenario test proving a real 'stays pending' path is reachable if one is actually intended.

contract/r/gnoswap/staker/v1/calculate_pool_position_reward.gno
```diff
@@ -409,19 +428,23 @@ func (self *PoolResolver) resolveInternalRewardSegments(poolTier *PoolTier, pool
 // appendInternalRewardTailSegments appends the schedule for the tail [startTime, endTime], where no
 // persisted cache entry exists beyond startTime. Over this span tier/count are constant, so the rate
 // changes only at halving boundaries.
-func appendInternalRewardTailSegments(segments []internalRewardSegment, poolTier *PoolTier, poolPath string, startTime, endTime, baseReward int64) ([]internalRewardSegment, error) {
-	tier := poolTier.CurrentTier(poolPath)
+func appendInternalRewardTailSegments(segments []internalRewardSegment, poolTier *PoolTier, poolPath string, startTime, endTime, baseReward int64, exit *sr.UnstakedPosition) ([]internalRewardSegment, error) {
+	// A checkpoint reads the tier context it exited under: this window closed at the exit.
+	tier, ratio, count, err := poolTier.tierContextOf(poolPath)
```
junghoon-vans, line 433: **Non-blocking / perf:** `poolTier.tierContextOf(poolPath)` runs unconditionally even when `exit != nil`, but its result is immediately discarded and overwritten by the exit-pinned values two lines below. This isn't currently exploitable — `tierContextOf` can't return a non-nil error given today's tier-range guards (`assertIsValidPoolTier` keeps `CurrentTier` in `{0,1,2,3}`) — but every checkpoint collect still pays for a live current-tier lookup (`CurrentTier` + `tierRatio.Get`) it doesn't need, and it couples a checkpoint's result to an unrelated live-pool-tier failure mode that a future tier-config change could turn into a real outage for this permissionless path.

Consider skipping the lookup entirely when `exit != nil`:

```go
var tier, ratio, count uint64
var err error
if exit != nil {
    tier, ratio, count = exit.Tier(), exit.TierRatio(), exit.TierCount()
} else {
    tier, ratio, count, err = poolTier.tierContextOf(poolPath)
    if err != nil {
        return nil, err
    }
}
```

This also makes the checkpoint path's independence from live pool tier state explicit, matching the comment above it.

contract/r/gnoswap/staker/v1/external_token_list.gno
```diff
@@ -92,3 +92,62 @@ func checkCannotRemoveDefaultToken(tokenPath string) error {
...
+//
+// Only the admin or governance can set the flag.
+func (s *stakerV1) SetDeniedRewardToken(_ int, rlm realm, tokenPath string, denied bool) {
+	access.AssertIsRlmCurrent(0, rlm)
+
+	halt.AssertIsNotHaltedStaker()
+
+	previousRealm := rlm.Previous()
+	caller := previousRealm.Address()
+	access.AssertIsAdminOrGovernance(caller)
```
junghoon-vans, line 120: Please wire this operation into the governance parameter registry as well. The permission check allows governance, but `gov/governance/v1/parameter_registry.gno` has no handler for `SetDeniedRewardToken`, so the current governance implementation cannot execute this switch through a parameter-change proposal. Admin calls work, but the documented governance path is unavailable. Please add the handler with token-path/boolean validation and cover execution through a governance proposal.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -930,8 +1188,8 @@ func (s *stakerV1) UnStakeToken(_ int, rlm realm, positionId uint64) string { //
 	// unStaked status
 	poolPath := deposit.TargetPoolPath()
 
-	// claim All Rewards
-	s.CollectReward(0, rlm, positionId)
+	// record the exit checkpoint; collecting is left to the Collect* entry points
+	checkpoint := s.recordUnstakedPosition(0, rlm, positionId, deposit, time.Now().Unix())
```
junghoon-vans, line 1197: Please update the reward quote getters alongside this lifecycle change. Unstaking now leaves a payable checkpoint and removes the active deposit, but `CollectableEmissionReward` and `CollectableExternalIncentiveReward` in `v1/getter.gno:277-306` still return early when `getDeposit(positionId)` fails. Consequently, a position can successfully collect its exit rewards while both quote APIs report that its deposit does not exist. The read-only calculation should resolve the checkpoint deposit, exit timestamp and pinned state too; removing only the deposit check would still leave it calculating against the wrong time. Please cover both quotes after unstaking and before settlement.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -467,20 +471,180 @@ func (s *stakerV1) newCollectContext(_ int, rlm realm, positionId uint64, curren
...
+		"CollectUnstakedPositionSource",
+		"prevAddr", previousRealm.Address().String(),
+		"prevRealm", previousRealm.PkgPath(),
+		"positionId", utils.FormatUint(target.positionId),
+		"poolPath", target.deposit.TargetPoolPath(),
+		"source", source,
+		"incentiveId", incentiveId,
+		"exitTime", utils.FormatInt(target.checkpoint.ExitTime()),
+		"emissionCollected", utils.FormatBool(target.checkpoint.EmissionCollected()),
+		"pendingIncentiveCount", utils.FormatInt(int64(len(target.checkpoint.PendingIncentiveIdList()))),
```
junghoon-vans: Could we obtain this count without materializing the ID list? `PendingIncentiveIdList()` allocates a slice and walks every remaining entry in `pendingIncentiveIds`, even though this event only needs the count. Since this helper runs after each source settles, draining N incentives individually adds O(N²) cumulative traversal just for event metadata, working against the per-source collection design. Please expose a constant-time count accessor backed by `len(pendingIncentiveIds)` and use it here.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -707,12 +902,61 @@ func (s *stakerV1) deliverExternalIncentiveReward(
...
+			"incentiveId", incentiveId,
+			"rewardToken", rewardToken,
+			"requiredBalance", utils.FormatInt(requiredBalance),
+			"stakerBalance", utils.FormatInt(common.BalanceOf(rewardToken, stakerAddr)),
+			"currentTime", utils.FormatInt(ctx.currentTime),
+			"currentHeight", utils.FormatInt(ctx.blockHeight),
+		)
+
+		if ctx.checkpoint != nil {
+			return "", 0, 0, externalDeliveryUnpayable
```
junghoon-vans: Is it intentional that a temporary ledger balance shortfall is treated as `Unpayable` for a checkpoint, while the same condition remains `Deferred` for a staked position? A closed accrual window fixes the amount owed, but does not prevent the token balance from being replenished later. Since checkpoint collection is permissionless, is the intended policy to let any caller relinquish the pending claim here to unblock re-staking/refunds, or should a recoverable balance shortfall remain deferred with an explicit owner-controlled waiver? Could you clarify the intended tradeoff and the terminal state/event semantics?

junghoon-vans:
The follow-up changes look good; I’ve resolved all remaining review threads. Adding regression tests for SetDeniedRewardToken execution through a governance proposal and delayed post-unstake reward quotes (including a non-zero emission reward) would round this out.

junghoon-vans (approved):

PR #1434, opened 2026-09-04
feat(router): refund route-token leftovers to the caller in-swap

jinoosss:
## What changed

The router now refunds any route-token surplus it accrues during a swap back to the swap caller **within the same transaction**.

- **`contract/r/gnoswap/router/v1/refund.gno` (new)**
  - `captureFundsSnapshot` records, per route token, the router's balance and pending protocol fee before the swap moves any funds.
  - `refundLeftovers` runs after the output payout and transfers each token's surplus over the snapshot to the caller, emitting a `SwapRouteRefund` event per refunded token.
- **`exact_in.gno` / `exact_out.gno`**: both swap paths (covering all four entry points) take the snapshot before `commonSwapRoute` and refund after the output transfer.
- **`docs/router.md`**: documents the leftover-refund behavior.

## Why

In a multi-hop swap, each hop's output is delivered to the router and fed into the next hop. When an intermediate hop only partially fills (thin liquidity: the pool price reaches its limit before consuming the full amount), the unconsumed intermediate tokens were left stranded in the router with no way back to the user. This mirrors the intent of Uniswap's `refundETH`/`sweepToken`, adapted to GnoVM constraints (see security notes below).

## Security considerations

- **Delta-only refunds**: only the surplus accrued during the current call is refunded. Funds the router held before the swap are never touched, so pre-existing balances cannot be drained by a later caller.
- **Pending protocol fee exclusion**: when the protocol fee realm is halted mid-swap, the swap fee stays in the router as a pending fee. The refund computation subtracts the signed pending-fee delta, so those fees are excluded exactly (and a settlement of old pending fees during the swap nets out correctly).
- **CEI**: the refund runs after all state updates and payouts, as the final interaction of the call.
- **No permissionless sweep**: Uniswap's claim-anyone model relies on multicall atomicity, which does not exist on GnoVM. Refunds go only to the swap caller, in-swap.
- **Fail-closed arithmetic**: all delta math uses `gnsmath.SafeSubInt64`, so any overflow aborts the transaction.

## Tests

- **Unit** (`refund_test.gno`): route-token collection/dedup; delta-only refund with pre-existing balance preserved; pending-protocol-fee exclusion (1000 accrued, 400 pending → 600 refunded); no-op when nothing accrued; an end-to-end multi-hop swap against a thin narrow-range pool asserting the leftover comes back to the caller and the router balance is unchanged.
- **Scenario filetests** (`contract/r/scenario/router/`), amount-focused:
  - `exact_in_multi_swap_route_partial_fill_refund_filetest.gno`: 10,000,000 BAR in → 9,609,680 BAZ leftover refunded, 99,999 QUX out, router deltas 0/0/0.
  - `..._with_pending_protocol_fee_filetest.gno`: same route with a 1% swap fee and the protocol fee realm halted → gross output 99,999 splits into 99,000 to the swapper + 999 kept in the router matching the recorded pending fee exactly; the 9,609,680 BAZ refund is unaffected.
- Full `router/v1` package suite and the full `scenario/router` filetest suite pass; existing golden outputs are unchanged.

### DIFF
--- a/contract/r/gnoswap/router/v1/exact_in.gno
+++ b/contract/r/gnoswap/router/v1/exact_in.gno
@@ -170,6 +170,8 @@ func (r *routerV1) exactInSwapRoute(
 	params SwapRouteParams,
 	referrer string,
 ) (int64, int64) {
+	fundsSnapshot := r.captureFundsSnapshot(params.routeArr)
+
 	inputAmount, outputAmount, err := r.commonSwapRoute(0, rlm, params)
 	if err != nil {
 		panic(err)
@@ -180,9 +182,12 @@ func (r *routerV1) exactInSwapRoute(
 
 	common.SafeGRC20Transfer(0, rlm, params.outputToken, caller, outputAmount)
 
+	// refund any route-token surplus (e.g. a partially filled intermediate hop)
+	// within the same transaction
+	r.refundLeftovers(0, rlm, fundsSnapshot, caller)
+
 	// handle referral registration
 	actualReferrer := referral.TryRegister(cross(rlm), caller, referrer)
-
 
 	eventAttrs := append([]string{
 		"prevAddr", caller.String(),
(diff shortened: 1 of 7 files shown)


jinoosss:
Close PR because partial execution is not allowed.

PR #1435, opened 2026-09-04
fix(router): revert exact-in swaps that cannot consume the full input

jinoosss:
## Summary

An exact-in swap whose requested input cannot be fully consumed (a partial fill caused by exhausted pool liquidity) previously settled silently for the smaller amount. This PR makes the router **revert** in that case, while keeping intentional partial fills available through an explicit `sqrtPriceLimitX96`.

## Motivation

- **User intent**: an exact-in swap is a commitment to trade the full requested input. Settling for an arbitrary smaller fill can slip past a loose `amountOutMin` and produce a trade the user never asked for.
- **Fund safety (multi-hop)**: intermediate hops pay from the router's own balance. If an intermediate hop partially fills, the unconsumed remainder of the previous hop's output is stranded in the router with **no recovery path** (the router deliberately has no sweep function and must hold no balance after a swap).

### Router (`contract/r/gnoswap/router/v1`)

- `finalizeSwap` now validates `resultAmountIn == amountSpecified` for `ExactIn` swaps (`SwapValidator.exactInAmount`) and aborts with `[GNOSWAP-ROUTER-020]` on mismatch.
  - Exemption: when the caller sets `sqrtPriceLimitX96`, partial consumption up to that price is the intended semantics (mirrors the existing `ExactOut` exemption).
- New `assertHopFullyConsumed` / `validateHopFullyConsumed` guards (`assert.gno`): every exact-in hop of `multiSwap` and `executeCollectedSwaps` must consume its specified input entirely, preventing funds from being stranded in the router.
- `DrySwapRoute` mirrors both the route-level and hop-level guards, so dry runs predict the revert instead of returning amounts the real swap would reject.

### Tests

- Unit tests: `TestSwapValidator_ExactInAmount`, `TestAssertHopFullyConsumed`, and updated `ValidateSwapResults` cases (partial fill and over-consumption now fail).
- New scenario `exact_in_partial_fill_revert_filetest.gno` covering the full contract:
  1. `DrySwapRoute` reports the partial fill as an error without touching state.
  2. The same swap **succeeds** with an explicit price limit (opt-in partial fill).
  3. The unbounded swap **aborts** with `[GNOSWAP-ROUTER-020]`.
- Existing scenarios that intentionally drive pools to terminal ticks (`scenario/router` terminal-tick and inactive-liquidity tests, `scenario/staker` in-range-change tests) now use `ExactInSingleSwapRoute` with an explicit price limit — the sanctioned way to perform bounded partial swaps. Swap results and tick expectations are unchanged.

## Behavior

| Case | Before | After |
| --- | --- | --- |
| ExactIn, no price limit, full fill | succeeds | succeeds (unchanged) |
| ExactIn, no price limit, partial fill | settles for smaller amount | **reverts** (`GNOSWAP-ROUTER-020`) |
| ExactIn, explicit `sqrtPriceLimitX96`, partial fill | succeeds | succeeds (unchanged) |
| Multi-hop, intermediate hop partial fill | remainder stranded in router | **reverts** |
| `DrySwapRoute` on a partial fill | returns partial amounts | returns amounts + error |

## Testing

```
gno test ./gno.land/r/gnoswap/router/v1      ok
gno test ./gno.land/r/gnoswap/router         ok
gno test ./gno.land/r/gnoswap/scenario/router  ok
gno test ./gno.land/r/gnoswap/scenario/staker  ok
gno test ./gno.land/r/gnoswap/scenario/getter  ok
gno test ./gno.land/r/gnoswap/scenario/upgrade ok
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Exact-input swaps without a price limit now require the consumed amount to exactly match the requested amount.
  * Partial fills and over-consumption are rejected with insufficient-liquidity errors when full consumption is required.
  * Multi-hop swaps validate full consumption for applicable exact-input hops.
  * Dry-run validation now reports the same failures as live execution.

* **Tests**
  * Added coverage for exact matches, partial fills, over-consumption, and multi-hop validation errors.


junghoon-vans:
DrySwapRoute currently promises to preserve simulated amounts on post-simulation failures, and this PR describes partial fills as returning amounts plus an error. That remains true for single-hop paths because ValidateSwapResults returns an error directly. In the new multi-hop path, however, multiDrySwap calls assertHopFullyConsumed, which panics. DrySwapRoute recovers the panic but overwrites both quote amounts with "0". Thus a multi-hop exact-in partial fill reports "0, 0, err" rather than the simulated partial quote, unlike the single-hop path. Please propagate validateHopFullyConsumed error through the dry path and preserve the computed quote values, or update the documented API contract.

junghoon-vans (approved):

PR #1452, opened 2026-09-14
perf(launchpad): store deposit payloads as values

junghoon-vans:
## Summary
- Store Launchpad `Deposit` payloads as values instead of pointers.
- Return public and internal deposit getters by value and remove `Deposit.Clone`.
- Pass deposits by value through withdrawal, reward collection, and tier accounting; explicitly persist the updated withdrawal value.
- Move the pointer-producing `newDeposit` fixture helper into `_helper_test.gno`.

## Decision
Proceed with the fresh-deployment-only value-storage change for this measured workload: both aggregate execution gas and net storage decrease. This is not deployment approval or a claim that every operation becomes cheaper; read-heavy workloads still require separate cost assessment.

## Context and changes
Each participant has one Deposit payload in the B+ tree. Value storage removes pointer serialization overhead. Withdrawal helpers return an updated local value plus the original withdrawal amount, and the caller writes that value back to the deposits tree. Read-only reward and tier helpers take values; setters retain pointer receivers only to modify local values.

Proxy interfaces, mocks, test callers, and upgrade scenario implementations match the value-return API. The existing public withdrawal lifecycle test checks persisted zero amount, exact withdrawal height/time, returned amount, and unchanged caller-held values.

## Performance evidence
**Full active metric suite: 86/86 fixtures, 169/169 metric rows per revision, no embedded errors.** This is not a Launchpad-only filtered run. The existing archived `.gnoa` fixture remains excluded by the upstream runner.

Candidate minus baseline. Storage is the sum of net per-realm storage diffs during measured callbacks, not peak footprint or total writes. Totals sum independent benchmark scenarios, not a production transaction mix.

| Scope | Baseline gas | Candidate gas | Gas delta | Gas delta % | Net storage delta (B) | CPU cycles delta |
|---|---:|---:|---:|---:|---:|---:|
| Entire active metric suite | 4,705,253,794 | 4,704,622,889 | -630,905 | -0.0134% | -40,492 | -77,126 |
| Launchpad lifecycle subset (23 rows) | 1,494,266,852 | 1,493,870,048 | -396,804 | -0.0266% | -40,016 | -73,982 |
| All other metrics (146 rows) | 3,210,986,942 | 3,210,752,841 | -234,101 | -0.0073% | -476 | -3,144 |

Total net storage: **911,623 → 871,131 B** (−4.4417%). Total CPU cycles: **4,254,508,198 → 4,254,431,072** (−0.0018%).

33 rows change; all are Launchpad-related. The remaining 136 rows are identical in gas, storage, and cycles. The lifecycle subset still has regressions: reward claims add **192,676 gas (+0.1018%)**, and retained getters add **27,900 gas (+2.5349%)**. The suite aggregate is not a claim that every operation gets cheaper.

- [Complete 169-row metric comparison](https://github.com/gnoswap-labs/gnoswap-performance-tracker/blob/5f2cd2d132429f8e6880275f826c77bda38d308f/reports/metric/compares/diff_e3bf690_ed4e8f5.md)
- [Runtime, fixture hashes, complete-suite coverage, totals, and state snapshots](https://github.com/gnoswap-labs/gnoswap-performance-tracker/blob/5f2cd2d132429f8e6880275f826c77bda38d308f/reports/metric/compares/state_e3bf690_ed4e8f5.json)

## Measurement revisions
- Tracker rebased onto main `babaa9e328c4fbb2cc5f7d3fbba0ac80509fa18b`.
- Baseline contract: `ed4e8f52dae3ef073ad0b1660b88af7a77e4df3a` — the implementation PR's pre-change merge base.
- Candidate contract: `e3bf69050a8ea46cce93b5fba3523b8d83ebf36f` — the final head of Contract #1452, including review changes.
- Metric-enabled Gno: `a021c9ee3ccfc0ec74350614d0dca9c3e1ade504` (`gas-9c8eb132`). Both sides use this same pinned runtime and identical fixture sources.

Contract #1452 is now merged. This comparison intentionally retains its pre-change baseline and final head; comparing two current-main snapshots would not isolate the change. Earlier PoC and single-fixture reports used different runtime/fixture inputs and are historical, not a like-for-like time series.

## Reproduction and verification
From this tracker branch with its pinned submodules initialized, run **without `METRIC_FILES`**:

```sh
make compare-metric-force e3bf69050a8ea46cce93b5fba3523b8d83ebf36f ed4e8f52dae3ef073ad0b1660b88af7a77e4df3a
```

The unrestricted command completed for both revisions. Temporary raw-output capture verified every active fixture produced metrics, no output contained `// Error:`, and every raw metric matched its generated report row. This matters because `-update-golden-tests` can otherwise bless errors while returning success. Both captures matched the committed fixture sources; the 15 Launchpad lifecycle STATE snapshots were identical. The lifecycle fixture covers three projects with 64 + 32 + 0 deposits and 16-record batches. Temporary capture instrumentation was removed before committing.

## Verification
- `make test PKG=gno.land/r/gnoswap/launchpad/v1`
- `make test PKG=gno.land/r/gnoswap/launchpad`
- `make test PKG=gno.land/r/gnoswap/scenario/upgrade`
- Full tracker suite above: both revisions, 86 fixtures and 169 metric rows each; 15 identical Launchpad state snapshots.

## Related PRs
Companion benchmark: https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/62

## Compatibility
This PR targets a new Launchpad deployment. Existing pointer-backed deposit payloads are not decoded or migrated. Event payloads and public event schemas are unchanged; no batch, indexer, API, or frontend companion change is required.

## Deployment Notes
Fresh deployments only: do not activate this implementation over an existing pointer-backed deposits tree. Merge this Contract PR before deploying the new Launchpad contract. Tracker #62 provides evidence only and has no deployment dependency. Re-measure if runtime-affecting changes are added after the measured candidate.

## Cost assumptions
For the 23-row Launchpad lifecycle subset only, at 100 ugnot/B and 1 ugnot/1,000 gas, the simple same-rate equivalent is −4.001600000 GNOT storage plus −0.000396804 GNOT execution gas = **−4.001996804 GNOT**. This is not an actual wallet debit or lifetime-profit estimate; identical fixed GasFee means zero cash gas delta. Deployment cost, transaction overhead, price changes, and capital time value are excluded.

## Summary by CodeRabbit

* **Changes**
  * Deposit records are now returned and processed as values rather than pointers, providing more predictable handling across launchpad operations.
  * Missing or invalid deposit lookups now return a default empty deposit alongside the error instead of a null result.
  * Deposit creation now uses the `MakeDeposit` API, while the previous constructor and cloning capability are no longer available.
  * Stored withdrawal updates use independent deposit values, preventing unintended changes from local mutations.

* **Tests**
  * Added coverage for value-based construction, default deposits, and deposit retrieval.


contract/r/gnoswap/launchpad/v1/launchpad_deposit.gno
```diff
@@ -153,9 +153,10 @@ func (lp *launchpadV1) depositGns(
 		projectTier.EndTime(),
 	)
```
jinoosss: How about defining a function that creates a value object?

junghoon-vans: Implemented in d34bced4. Added `launchpad.MakeDeposit`, which constructs a `Deposit` value, and use it for persisted deposits and value getter fallbacks. `NewDeposit` remains as the pointer-producing wrapper for mutable callers.

contract/r/gnoswap/launchpad/v1/deposit.gno
```diff
@@ -21,7 +21,7 @@ func newDeposit(
 	createdTime int64,
 	endTime int64,
 ) *launchpad.Deposit {
-	return launchpad.NewDeposit(
```
jinoosss, line 24: Is this necessary?

contract/r/gnoswap/launchpad/v1/launchpad_withdraw.gno
jinoosss: Why use pointer deposit object `*launchpad.Deposit` ?

jinoosss (commented):
I would appreciate it if you could upload the performance report after revising the comments.

contract/r/gnoswap/launchpad/v1/deposit.gno
junghoon-vans: [e3bf690](https://github.com/gnoswap-labs/gnoswap/pull/1452/commits/e3bf69050a8ea46cce93b5fba3523b8d83ebf36f)

contract/r/gnoswap/launchpad/v1/launchpad_withdraw.gno
junghoon-vans: [e3bf690](https://github.com/gnoswap-labs/gnoswap/pull/1452/commits/e3bf69050a8ea46cce93b5fba3523b8d83ebf36f)

jinoosss (approved):

PR #1453, opened 2026-09-14
feat(r/common): GRC20 token management with versioned implementation

jinoosss:
## Description

Makes `r/gnoswap/common` version-agnostic with respect to the GRC20 package generation, so that future GRC20 generations (`p/nt/grc20/v1`, `r/nt/grc20reg/v1`, ...) can be supported by activating a new `common/vN` implementation instead of redeploying core realms.

Context: upstream gno moved `p/demo/tokens/grc20` and `r/demo/defi/grc20reg` under `/v0` (gnolang/gno#6161, folded into #6162) explicitly so that successor generations can be published. `common` is deployed once at a permanent path and cannot be redeployed, so it must not be bound to a single generation. GNS/xGNS stay on the generation they were deployed with; v0 and v1 tokens are treated as distinct tokens.

### `r/gnoswap/common` becomes a proxy realm

Follows the same layout, naming and permission handling as the other proxy realms (`protocol_fee`, `pool`, ...):

| File | Content |
|---|---|
| `types.gno` | `ICommon`: the implementation interface |
| `state.gno` | `versionManager`, `implementation`, `getImplementation`, `updateImplementation` (no domain store) |
| `upgrade.gno` | `RegisterInitializer`, `UpgradeImpl` (admin or governance), `GetImplementationPackagePath` |
| `grc20reg_helper.gno` | Public helpers now validate the token key and delegate to the implementation |

`ICommon` exposes an operation surface only, no token types:

```go
type ICommon interface {
	IsRegistered(tokenKey string) bool
	TotalSupply(tokenKey string) int64
	BalanceOf(tokenKey string, addr address) int64
	Allowance(tokenKey string, owner, spender address) int64
	Transfer(_ int, rlm realm, tokenKey string, to address, amount int64) error
	TransferFrom(_ int, rlm realm, tokenKey string, from, to address, amount int64) error
	Approve(_ int, rlm realm, tokenKey string, spender address, amount int64) error
}
```

Public helper signatures (`BalanceOf`, `SafeGRC20Transfer`, `IsRegistered`, ...) are unchanged, so callers need no changes.

**Removed:** `common.GetToken` and `common.GetTokenTeller`. They returned concrete `*grc20.Token` / `grc20.Teller` values and were the only way a concrete GRC20 type could leak into core realms. The single production caller (`pool/v1/transfer.gno`) now uses `common.BalanceOf`.

**Added:** token key validation in the proxy. Empty keys and keys containing `:` are rejected before resolution, since keys are embedded verbatim in pool paths (`<key0>:<key1>:<fee>`).

### `r/gnoswap/common/v1`

First implementation. Resolves keys through `gno.land/r/nt/grc20reg/v0` and executes each operation via `RealmTeller(0, rlm)`. This is the only place in GnoSwap, besides the token realms themselves, that imports a concrete GRC20 package.

The `rlm` threading is preserved: proxy -> implementation -> `RealmTeller` are all same-realm borrows, so `rlm.IsCurrent()` holds and the teller actor is the calling realm, exactly as before.

## Adding a future GRC20 generation

1. Deploy `r/gnoswap/common/v2` implementing `ICommon` against every registry generation (v0 first). It must panic if more than one registry claims the same key.
2. Its `init` registers via `common.RegisterInitializer(cross(cur), ...)`.
3. Admin or governance calls `common.UpgradeImpl("gno.land/r/gnoswap/common/v2")`.

Core realms and their KV state are untouched by the switch.

## Summary by CodeRabbit

* **New Features**
  * Added a versioned Common package with upgrade and rollback support.
  * Added token registration validation for balances, transfers, approvals, and allowances.
  * Added support for tokens across multiple registry generations, prioritizing the original registry for duplicate keys.
  * Added protection against unsupported native-coin transfers.
  * Added governance controls for Common implementation upgrades.

* **Bug Fixes**
  * Improved validation and error messages for invalid or unregistered token keys.

* **Tests**
  * Expanded coverage for upgrades, permissions, cross-generation tokens, and pool continuity.


contract/r/gnoswap/common/_mock_native_coin_test.gno
```diff
@@ -0,0 +1,13 @@
+package common
+
+import (
+	"chain/runtime/unsafe"
+	"errors"
+)
+
+func (m *MockGrc20regCommon) ValidateNotHandleNativeCoin() error {
+	if len(unsafe.OriginSend()) > 0 {
+		return errors.New("[GNOSWAP-COMMON-002] handle native coin is not allowed")
```
notJoon, line 10: how about using pre-defined error messages here? 
```suggestion
		return errors.New(errNotHandleNativeCoin)
```

notJoon (commented):
confirmed common package dose not introducing any logical changes.

junghoon-vans (approved):

PR #1454, opened 2026-09-14
refactor(governance): store proposal voting info as values

junghoon-vans:
## Summary
- Store per-proposal voting records as `governance.VotingInfo` values instead of pointers.
- Remove `VotingInfo.Clone` and `VotingInfoResolver`; construct immutable unvoted/voted values with `NewVotingInfo` and `NewVotedVotingInfo`.
- Return typed voting values from proposal lookup and vote lifecycle helpers, while preserving read-only value-backed getter entries.

## Decision
- Proceed with the value-backed representation. The paired 128-voter lifecycle measurement preserves all observed lifecycle states and reduces both storage and execution gas in the measured scope.
- This is not deployment approval. Existing persisted pointer-backed voting records need a migration plan before this representation can be used against legacy realm state.

## Context
- Proposal voting-info entries are written once per voter and are retained in the per-proposal tree.
- The pointer-backed representation required cloning at realm boundaries. Values remove that clone path and make stored-entry ownership explicit.

## Changes
- Store `VotingInfo` values in proposal voting-info trees.
- Replace pointer mutation with `NewVotedVotingInfo` after duplicate-vote validation.
- Remove the obsolete voting-info resolver and pointer-based test fixtures.
- Keep public read APIs value-backed and return `VotingInfo` directly from the internal lookup.

## Performance Evidence
Measured in [Tracker #67](https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/67). Delta is candidate minus baseline; negative values are improvements. Storage is the tracker's net per-realm storage diff during each workload, not peak storage footprint or total writes.

| Workload | Baseline | Candidate | Delta |
|---|---:|---:|---:|
| 128 distinct voter lifecycles | 257,150 B | 204,002 B | -53,148 B |
| 128 distinct voter lifecycles | 159,862,108 gas | 159,646,006 gas | -216,102 gas |
| Full-tree read | 100,167 gas / 0 B | 100,167 gas / 0 B | unchanged |

- Same inputs: yes.
- Same observable states: yes, across five lifecycle checkpoints.
- Modeled combined delta: `-5.315016 GNOT` (`-5.314800 GNOT` storage; approximately `-0.000216 GNOT` execution gas).
- The measurement targets contract candidate `21189b72f697deaa95272efcd26b4075e700ad83`; it validates the representation change. Any later head-only change must be remeasured before being treated as a new performance result.

## Reproduction
- Baseline contract: `b12f0f8760a3e88e9539b0c8cfd32b38a5297c68`
- Measured candidate: `21189b72f697deaa95272efcd26b4075e700ad83`
- Tracker: [#67](https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/67)
- Pinned Gno runtime: `bac78a97b20e12ba471bdd73a41af31bd94bdc12`

```sh
python3 run_case.py voting-info --skip-unit
```

## Verification
- Tracker #67: baseline capture, candidate capture, comparison, and assessment passed.
- `make test PKG=gno.land/r/gnoswap/gov/governance/v1`
- `make test PKG=gno.land/r/gnoswap/scenario/getter`
- `git diff --check`

## Compatibility
- Event payloads and public event schemas are unchanged; no batch, indexer, API, or frontend companion change is required.
- Existing proposal voting-info trees with pointer payloads are incompatible with this value-only representation. Do not deploy against legacy state without an explicit migration.

## Cost Assumptions
- Tracker #67 uses the Pearl observation rate of 1,000 gas = 0.000001 GNOT and 1 B = 0.000100 GNOT.
- Storage pricing models fresh-state deposit/refund effects. Values are modeled estimates, not wallet debits.
- Deployment/migration cost, transaction overhead, and capital time value are excluded.


contract/r/gnoswap/gov/governance/voting_info.gno
```diff
@@ -18,20 +18,32 @@ type VotingInfo struct {
...
 
+// NewVotedVotingInfo creates a voting information value with a recorded vote.
+func NewVotedVotingInfo(availableVoteWeight int64, votedYes bool, votedWeight, votedHeight, votedAt int64) VotingInfo {
+	return VotingInfo{
+		availableVoteWeight: availableVoteWeight,
+		votedWeight:         votedWeight,
+		votedHeight:         votedHeight,
+		votedAt:             votedAt,
+		votedYes:            votedYes,
+		voted:               true,
```
jinoosss, line 43: Is `voted` requried ?

jinoosss (approved):

notJoon (approved):

contract/r/gnoswap/gov/governance/voting_info.gno
junghoon-vans: Yes. I recommend keeping `voted`: `votedYes == false` is shared by unvoted records and recorded no votes. Inferring the state from `votedWeight > 0` would make positive weight a permanent persisted-state invariant; a zero-weight recorded value from a future migration or fixture could otherwise bypass duplicate-vote protection. The explicit flag keeps the state unambiguous without relying on sentinel field values.

PR #1456, opened 2026-09-14
perf(staker): store external incentive rate by value

junghoon-vans:
## Summary
- Store `ExternalIncentive.rewardPerSecondX128` as `u256.Uint` rather than `*u256.Uint`.
- Return the rate by value through the staker getter API and remove redundant rate clones.
- Return `GetPoolIncentives` entries as value snapshots, eliminating `ExternalIncentive.Clone()` and pointer allocation on read-only tree access.

## Decision
- Proceed with this fresh-deployment-only contract change: the selected external-incentive-rate-only representation had a lower measured storage cash flow than the baseline in the separated workload.
- This is a Contract PR decision, not deployment approval. Correctness, performance remeasurement, deployment cost, and compatibility remain separate gates.

## Context
- `u256.Uint` is a fixed four-word value. Persisting the rate by value removes the pointer-backed representation without encoding it as a binary string.
- `ExternalIncentive` now contains only value fields. A read-only tree can safely return `*incentive` as an `ExternalIncentive` value snapshot instead of allocating a cloned pointer.
- The PR intentionally excludes `Deposit.liquidity`, `Tick.stakedLiquidityGross`, and `Tick.stakedLiquidityDelta`.

## Changes
- Change `ExternalIncentive.rewardPerSecondX128` to `u256.Uint`.
- Return the rate by value from the model, implementation, public getter, and test upgrade implementations.
- Replace pointer-cloned external-incentive tree entries with value snapshots and remove `ExternalIncentive.Clone()`.
- Pass local value addresses only to the existing pointer-based uint256 arithmetic helpers.
- Cover zero value, setter-input mutation, getter-result mutation, and tree-entry value isolation.

## Performance Evidence
The selection measurement is recorded in [gnoswap-performance-tracker PR #61](https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/61). This is an explicit cross-repository URL; it does not refer to an issue or PR in this repository.

The tracker candidate was `0bee3ad1b436a2d858f69d730d79b0cfc6be9558`; this PR subsequently removes the no-longer-needed presence flag, converts getters to values, and eliminates pointer-cloned tree reads for a fresh deployment. Re-run the tracker comparison against this PR head before treating the numbers below as final branch-specific evidence.

| Workload | Storage cash-flow Δ | Execution-gas equivalent Δ | Net estimated Δ |
|---|---:|---:|---:|
| 33 deposits fully unstaked; signed tick crossings; 111 external reward collections; incentive end | -0.036400 GNOT | +0.000884 GNOT | -0.035516 GNOT |

Delta is candidate minus baseline. Storage and gas use the tracker’s per-realm deposit/refund model at 100 ugnot/B and 1 ugnot/1,000 gas; they are same-rate estimates, not wallet debits. With the same valid fixed `GasFee`, the monetary gas delta is zero. Migration and deployment costs are excluded.

## Reproduction
- Baseline contract: `b12f0f8760a3e88e9539b0c8cfd32b38a5297c68`
- Selection-measurement candidate: `0bee3ad1b436a2d858f69d730d79b0cfc6be9558`
- Current Contract PR head: `3d7b7886`
- Tracker workflow and fixture: [gnoswap-performance-tracker PR #61](https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/61)

## Verification
- `make test PKG=gno.land/r/gnoswap/staker`
- `gofumpt -d` on changed files
- `git diff --check origin/main...HEAD`

## Compatibility
- Fresh deployments only. No typed-state migration is included or required for this new contract.
- Event payloads and public event schemas are unchanged; no batch, indexer, API, or frontend companion change is required.

## Cost Assumptions
- 1,000 gas = 1 ugnot and 1 byte = 100 ugnot for the selection measurement.
- Workload rows must not be generalized into a lifetime or protocol-wide estimate; recurring collections and gas-price changes require their own joint analysis.


jinoosss:
It would be worth taking a fundamental look at the Uint type change.

PR #1459, opened 2026-09-14
perf(staker): use keyed incentive cancellation buckets

junghoon-vans:
## Summary
- Remove the full start-time bucket scan and slice copy from incentive cancellation.
- Store each bucket as an incentive-ID keyed B+ tree using the existing tree implementation.

## Context
Cancelling one pending incentive previously rebuilt the entire slice of IDs sharing its start timestamp. The work and temporary allocation grew linearly with the number of sibling incentives. This addresses the cancellation-cost observation in I-03; it does not claim a permanent fund lock, since post-end refunds remain available.

Compared against `main` at `0d1ef0c961d465e4a3e38653272e4197aa914bdf`; candidate `0a50d22a9268096355c8f8313b59f3c4e795fe00`.

## Changes
- Create a persistent keyed bucket when its first incentive is registered; reuse it for subsequent additions and removals.
- Delete only the selected ID and remove the timestamp entry when the bucket becomes empty.
- Preserve inclusive time-range discovery and callback early termination across nested buckets.
- Add regression coverage for sibling preservation across tree splits/deletions, missing-ID removal, bucket recreation, and early termination.
- Document the storage layout and traversal behavior.

## Performance Tracker Results
Full reports: https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/70

Negative gas deltas are improvements.

| Workload | Baseline gas | Candidate gas | Gas delta |
| --- | ---: | ---: | ---: |
| Add, empty bucket | 84,351 | 107,805 | +27.81% |
| Add, 100 siblings | 112,249 | 138,886 | +23.73% |
| Remove, 1 incentive | 160,052 | 213,506 | +33.40% |
| Remove, 10 incentives | 165,813 | 140,075 | -15.52% |
| Remove, 50 incentives | 356,888 | 167,982 | -52.93% |
| Remove, 100 incentives | 595,119 | 245,504 | -58.75% |
| Full iteration, 100 incentives | 275,642 | 1,287,720 | +367.17% |
| Early stop, 100 incentives | 110,716 | 138,102 | +24.74% |

The keyed bucket improves removal only when roughly 10 or more incentives share a start timestamp. Addition and traversal regress, including a 367.17% full-iteration increase at 100 incentives. The measurements therefore do not show a net benefit outside removal and do not support landing the current design as-is.

The focused index scenarios report a 0 B net storage diff; this is not a claim that the persistent layouts have identical storage footprints.

Exact tracker command:

```sh
make compare-metric-force 0a50d22a9268096355c8f8313b59f3c4e795fe00 0d1ef0c961d465e4a3e38653272e4197aa914bdf
```

## Verification
Passed locally:
- `make test PKG=gno.land/r/gnoswap/staker`
- `make compare-metric-force 0a50d22a9268096355c8f8313b59f3c4e795fe00 0d1ef0c961d465e4a3e38653272e4197aa914bdf` in the performance tracker

Coverage includes cancellation authorization/time boundaries, creator refunds, sibling preservation, end-of-incentive refunds, reward accounting, and same-start bucket add/remove/traversal costs. No live-chain gas-limit reproduction or deployment was performed.

## Compatibility
- Public function signatures and event schemas are unchanged.
- Within the same timestamp, discovery now visits IDs in key order rather than insertion order.
- No companion repository changes are required for this internal index change.

## Deployment Notes
- Intended for a new deployment only. No migration or legacy slice compatibility is included.
- Do not apply this storage layout to an existing deployment containing slice-valued buckets.
- Bucket creation has tree overhead; the measured benefit is limited to removal in larger same-start buckets.

## Rollback
Before deployment, revert this commit if needed. After tree-valued state exists, the previous slice-based implementation is not a compatible in-place rollback.


jinoosss:
It seems the cost could be higher.

junghoon-vans:
@jinoosss I updated the PR description with the performance-tracker results: https://github.com/gnoswap-labs/gnoswap-performance-tracker/pull/70

Your concern was correct. The keyed bucket improves removal for larger same-start buckets, but the benchmark shows no benefit outside removal: insertion regresses by up to 27.81%, early-stop traversal by up to 25.59%, and full iteration at 100 incentives by 367.17%. The current design does not look worthwhile as-is.

PR #1464, opened 2026-09-14
feat(protocol-fee): update pending protocol fees per token instead of copying the map

jinoosss:
## Summary

Resolves **GSW-MCP-M01**: the router, staker and pool each keep a `pendingProtocolFees` map for fees that could not be forwarded to `protocol_fee` (collection halted or transfer failed). Every add and every settlement copied the whole map into a fresh one, so the cost of a swap, reward collection or fee withdrawal grew with the number of distinct tokens accumulated during a halt, and recovery cost O(N) per token on top of that.

This PR keeps the single map and its store key, but adds per-token accessors so both paths touch one entry.

## Changes

- **Store layer** (`router`, `staker`, `pool` `store.gno` / `types.gno`, mocks)
  - Add `GetPendingProtocolFee(tokenPath)`, `SetPendingProtocolFee(tokenPath, amount)` and `RemovePendingProtocolFee(tokenPath)` that read, update and delete a single map entry in place.
  - `SetPendingProtocolFees` now copies the given map inside the store realm. In gnovm an object keeps the `PkgID` of the realm that allocated it even after being persisted in another realm's storage, so a map created by the `v1` implementation stays readonly for the proxy store and any in-place mutation panics with `cannot delete from readonly tainted map`. Allocating the copy in the store realm makes the per-token accessors valid.
- **Implementation realms** (`*/v1/protocol_fee*.gno`, `*/v1/init.gno`)
  - `addPendingProtocolFee` and `settleProtocolFee` use the per-token accessors; the copy loops are gone.
  - `initStoreData` re-stores the existing map once so a map persisted by an earlier implementation becomes owned by the store realm. Keys and values are unchanged.
- **Tests**
  - Seed the mock stores per token instead of with a map literal allocated in the test realm, which would hit the same readonly check inside the mock realm.

## Testing

Run against `gnoswap-labs/gno` master in an isolated checkout:

| Package | Result |
|---|---|
| `r/gnoswap/{pool,router,staker}/v1` | ok |
| `r/gnoswap/{pool,router,staker}` | ok |
| `r/scenario/{halt,position,protocol_fee,upgrade}` | ok |
| `r/gnoswap/test/fuzz` | compiles (full fuzz run skipped locally) |

## Notes

- `GetPendingProtocolFees()` (the public view) and `SetPendingProtocolFees()` are kept for compatibility; only the internal fee paths changed.
- No storage layout or key change; the one-time re-store in `initStoreData` is the only migration-like step and it is idempotent.

## Summary by CodeRabbit

* **New Features**
  * Added per-token pending protocol fee access and management across pool, router, and staking operations.
  * Pending fees can now be retrieved, updated, or removed individually without replacing the full fee collection.
  * Added authorization checks for fee updates and removals.

* **Bug Fixes**
  * Improved preservation of existing pending fees during initialization and storage updates.

* **Tests**
  * Expanded coverage for per-token fee handling, authorization failures, updates, and settlement cleanup.


junghoon-vans:
One point to consider while keeping the per-token optimization: the new SetPendingProtocolFee and RemovePendingProtocolFee methods in the router, pool, and staker stores check rlm.IsCurrent(), but no longer pass through the write-authorization check in kvStore.Set/Delete. IsCurrent validates the realm token for the current frame; it does not establish write permission for this store.

Could these methods preserve the existing KV-store writer policy before mutating the map in place? This would retain the performance improvement without making the per-token API enforce a different authorization policy from SetPendingProtocolFees. A store-level regression test covering authorized and unauthorized callers would help keep the two paths aligned.

This is an advisory note about preserving the storage authorization contract, not a claim of a demonstrated externally reachable exploit.

Reference: https://github.com/gnoswap-labs/gnoswap/blob/c17268d69dd44062ad3d6a5cb8e4d8ddd4b4c0c9/contract/r/gnoswap/router/store.gno#L91-L108 (the pool and staker stores have the same pattern).

junghoon-vans (approved):

PR #1465, opened 2026-09-14
refactor(gov/staker): settle protocol fee rewards on collect, not on stake change

jinoosss:
## Problem

`Undelegate` (and every other stake change on gov/staker) folded the pending protocol fees into the per-token accumulators inline, and then walked every token that ever collected a fee to refresh the staker's reward debt. Both costs are unbounded:

- the walk grows with the number of fee tokens ever seen (`H`), which anyone can raise by registering a token, creating a pool and swapping once;
- the fold grows with the number of tokens that collected a fee since the last stake change (`P`), which hits whoever touches gov/staker next.

Measured on the previous design (test harness gas meter, 3B block ceiling):

| path | H=10 | H=1000 | per token |
|---|---|---|---|
| stake change, accumulators unchanged | 128k | 128k | 0 |
| stake change, one token changed | 1.06M | 55.2M | ~55k |
| folding H pending tokens at once | 7.4M | 666M | ~600k |

At a 100M gas-wanted, ~160 pending tokens block every `Delegate` / `Undelegate` / `CollectReward` until the pending set is drained token by token. This is GSW-MCP-H01.

## Change

Stake changes no longer touch any accumulator. Fees are folded and settled per token, on collect.

### protocol_fee: accrual epochs and buckets

- `AdvanceAccrualEpoch` (gov/staker only) closes the epoch in force and returns the new one. gov/staker calls it on every stake change.
- `AddToProtocolFee` buckets the gov/staker share under `(tokenPath, epoch)` instead of appending to a flat pending list.
- `ConsumeAccrualBuckets(tokenPath, limit)` (gov/staker only) returns and clears up to `limit` of the oldest buckets of one token.
- `GetAccrualEpoch`, `GetAccrualPendingTokens`, `GetAccrualBuckets` expose the same without clearing.
- `reservedTokens` and `accrualPendingTokens` are tree-backed sets, so settling one token no longer rewrites the whole list.
- Removed: `ConsumeAccrualPendingProtocolFees`, `ConsumeAccrualPendingProtocolFeeByTokenPath`, `GetAccrualPendingProtocolFees`.

### gov/staker: epoch history instead of reward debt

- `ProtocolFeeRewardManager` keeps `currentEpoch`, `totalStakedHistory` (epoch → total stake in force) and one `ProtocolFeeTokenAccumulator` per token, which records the accumulated fee per stake after each folded epoch.
- A stake change only records the new total for the new epoch and appends one `ProtocolFeeStakeEvent` (epoch, stake after) to the staker. **Its cost is independent of `H` and `P`.**
- Folding a token divides each bucket by `GetTotalStakedAmountAt(epoch)`, the total that was live when the fee arrived, so a fee that arrived before a stake change is never redistributed by it. A bucket that arrives while nothing is staked is dropped, as before.
- A staker's reward for a token is settled from their own events: for each closed segment, `stake × (A(nextEvent.epoch − 1) − A(segment start))`, plus the open segment up to the current accumulator. An event may only be settled once the epoch before it is fully folded (`foldedEpoch`), so a partial fold never mis-attributes a fee across a stake change. The sub-unit remainder stays in `earnedX128` and tops up the next payout.
- `CollectProtocolFeeReward(tokenPath)` is the bounded path: at most `maxAccrualBucketsPerCollect` buckets and `maxStakeEventsPerCollect` events per call, the rest is paid by the next call. `CollectReward` still folds and settles every known token in one call, as before.
- Queries (`GetClaimableReward*`) project the pending buckets in memory and report exactly what a collect would pay.

Measured on this branch:

| path | H=10 | H=1000 |
|---|---|---|
| `removeStake` (Undelegate core) | 238,706 | 238,706 |
| single-token collect | 378k | 418k |

### Storage

This assumes a fresh deployment: `ProtocolFeeRewardManager`, `ProtocolFeeRewardState` and the protocol_fee store keys changed shape and no migration is included. A staker's event list grows with their own stake changes; each collect settles it incrementally and bounded.

## Behaviour changes

- `Delegate` / `Undelegate` / `Redelegate` / `SetAmountByProjectWallet` no longer emit `UpdateProtocolFeeRewardAccumulation`; they emit `UpdateProtocolFeeStakeAccrual` (epoch, total stake). `UpdateProtocolFeeRewardAccumulation` is emitted per folded token on collect and now carries `foldedEpoch`.
- `GetProtocolFeeAmount(tokenPath)` advances on collect rather than on stake change; `GetProtocolFeeAccumulatedTimestamp` moves on both.
- Payouts carry the sub-unit remainder, so a few golden outputs moved by one unit in the staker's favour (still within what gov/staker received). Reviewed per file:
  - `scenario/launchpad`: three files, `1499 → 1500`, `998 → 999`, `2498 → 2499`.
  - `scenario/gov/staker/zero_stake_window_reward_discard`: the "cursor advanced on delegate" line became "still pending until collect / per-stake stays zero".

## Verification

- `protocol_fee`, `protocol_fee/v1`, `gov/staker`, `gov/staker/v1`, `launchpad`, `launchpad/v1`, `gov/governance/v1`, `router/v1`, `pool/v1`, `staker/v1`, `position/v1`, `emission` and every `scenario/*` package pass.
- Unit tests rewritten for the new model: epoch/total history, accumulator history and floor lookups, `applyAccrualBuckets` (zero stake, partial fold, malformed input), exact attribution across stake changes, bounded settle, fold gating, remainder carry, projection equals collect, `MaxInt64` amounts.
- Twelve new scenario filetests under `contract/r/scenario/gov/staker/protocol_fee_*` compare amounts through golden output. Every expected value was computed by hand first (the derivation is in each file's header) and matched on the first run; the only correction was epoch numbering in `protocol_fee_many_stake_changes` (an undelegate across two delegation records opens one epoch per record).

| filetest | asserts |
|---|---|
| `epoch_attribution` | fees across three stake distributions, collected at the end, are attributed to the distribution live when they arrived |
| `undelegate_without_collect` | undelegate folds nothing; the share earned before leaving is paid later; the zero-stake window is dropped |
| `per_token_collect` | collecting one token leaves the others untouched |
| `same_block_order` | attribution follows transaction order within a block |
| `redelegate` | a redelegate (two epochs) keeps the share |
| `late_token_late_joiner` | a late token and a late joiner earn only from when they appear |
| `rounding_carry` | 1/3 shares floor per collect and the remainder is carried |
| `devops_split` | only the gov/staker share is bucketed |
| `collect_twice_same_second` | the all-token collect is gated per second, the single-token collect is not |
| `launchpad_project_share` | a project wallet shares epochs with delegators |
| `many_stake_changes` | six stake changes settled by one collect, query and single-token collect agree |
| `three_stakers_interleaved` | interleaved joins, exits and collect times sum to the fees |

- Docker integration tests were not run here; no txtar references the removed APIs.

## Docs

`docs/gov.md` (new gov/staker section), `docs/protocol_fee.md`, `protocol_fee` READMEs and `AGENTS.md` pitfalls updated.

## Summary by CodeRabbit

* **New Features**
  * Protocol fees are now organized into accrual epochs and per-token buckets, enabling older fees to be collected in order.
  * Added bounded bucket collection and accrual status visibility, including the current epoch and pending tokens.
  * Stake changes record settlement events, improving reward attribution across changing stake levels.
  * Added validation for accrual epochs, bucket limits, ordering, and malformed data.

* **Documentation**
  * Expanded guidance for gov staking, fee folding, collection limits, and common settlement pitfalls.


contract/r/gnoswap/gov/staker/v1/protocol_fee_reward_manager.gno
```diff
@@ -4,358 +4,280 @@ import (
...
 	}
 
-	resolvedState := NewProtocolFeeRewardStateResolver(rewardState)
-
-	currentTotal := self.GetTotalStakedAmount()
-	if currentTotal > math.MaxInt64-amount {
-		return errors.New(errTotalStakedAmountOverflow)
+	if len(pendingEpochs) == 0 {
+		return view
 	}
```
junghoon-vans, line 206: Please advance the projected folded epoch even when there are no pending buckets. This early return preserves the accumulator's old `foldedEpoch`, whereas an actual collect calls `applyAccrualBuckets(..., exhausted=true)` and advances it to `currentEpoch - 1` even for an empty batch. After fees have been folded and the staker subsequently changes their stake, the query can therefore stop at an unsettled event and omit rewards that collection would pay. A temporary consistency test on this commit returned `query=0 collect=100`. The public claimable queries should project the same empty-bucket advancement as collection, while keeping the stored-accumulator view used by settlement distinct from that projection.

contract/r/gnoswap/gov/staker/v1/protocol_fee_reward_state.gno
```diff
@@ -4,370 +4,269 @@ import (
...
-		// keeping cost proportional to the tokens that actually received fees (gas report issue 1 / §1.2).
-		if rewardDebtDeltaX128.IsZero() {
-			rewardAmounts[token] = 0
-			continue
+	for index := eventCount - 1; index >= 0; index-- {
+		event, ok := p.GetStakeEvent(index)
+		if !ok {
+			break
+		}
+		if event.GetEpoch() <= view.firstEpoch {
```
junghoon-vans: Please bound the initial event lookup as well as `settleEvents`. On the first collection of a token, this loop walks backward through the staker's history until it reaches `firstEpoch`; it runs before `maxStakeEventsPerCollect` is applied and has no limit of its own. Consequently, the initial collection cost still grows with all stake changes since the token's first fee, despite the documented per-call event bound. A binary search over the monotonically ordered event epochs, or an incremental initialization cursor with bounded work, would avoid the full-history scan. This is a static complexity finding; I have not measured a gas-limit failure.

contract/r/gnoswap/gov/staker/v1/protocol_fee_reward_manager.gno
jinoosss: Fixed in 467f342d5. `accumulatorView` now takes `pendingComplete`: the claimable queries pass the whole pending set (which may be empty) and the view advances `foldedEpoch` to `currentEpoch - 1` regardless, mirroring the empty exhaustive fold a collect performs. Settlement keeps using the stored accumulator view (`pendingComplete = false`), so a bounded partial fold still cannot settle an event past what is actually folded. Added a unit test that reproduces the `query < collect` case (bounded fold consuming exactly its limit, then stake changes without fees) and asserts both paths now pay the same amount.

contract/r/gnoswap/gov/staker/v1/protocol_fee_reward_state.gno
jinoosss: Fixed in 128c20b4e. Stake events are appended in epoch order, so `loadTokenSettlement` now locates the last event at or before `firstEpoch` by binary search over the event index (O(log n) tree reads) instead of scanning backward through the history. The initial lookup no longer grows with the number of stake changes since the token's first fee, and `settleEvents` remains bounded by `maxStakeEventsPerCollect` as before. Added a test with a 41-event history and duplicate epochs at the boundary to pin the cursor and the picked-up stake.

junghoon-vans: Checked the latest head, 128c20b4eb2b1c3263d605f629851cfa9d59fa21: loadTokenSettlement now uses binary search over the ordered stake-event epochs instead of the backward full-history scan. This addresses the initialization scan raised in this thread: the initial lookup takes logarithmically many event reads before the bounded settlement loop. This follow-up is based on source inspection; I have not rerun tests or measured gas on this new head.

junghoon-vans (approved):
Reviewed the mitigation on 128c20b4eb2b1c3263d605f629851cfa9d59fa21. Stake changes no longer fold or iterate over fee tokens, reserved/pending token operations are keyed, and initial per-token settlement now uses binary search. The gov/staker/v1 and protocol_fee/v1 package tests pass on this head. Approval covers the contract changes under the documented fresh-deployment assumption; live deployment and migration were not verified.

PR #1472, opened 2026-09-15
fix(staker): defer internal rewards during emission halt

junghoon-vans:
## Summary
- Defer internal GNS reward settlement while emission is halted.
- Keep external incentive collection available during the halt.

## Problem
`MintAndDistributeGns` returns `active = false` without minting or distributing GNS during an emission halt. The staker ignored that result, continued calculating internal rewards, and paid them from its existing GNS reserve. This made claims order-dependent and could deplete reserves needed by other stakers.

## Changes
- Make `CollectEmissionReward` return zero without changing internal reward state when emission is inactive.
- Add an external-only calculation path for `CollectReward` during an emission halt.
- Preserve the internal reward cursor, collected amount, reward cache, checkpoint state, and staker GNS reserve until emission resumes.
- Continue discovering, paying, and checkpointing external incentives during the halt.
- Replace the previous no-panic check with a halt-to-resume accounting regression scenario.

## Behavior Changes
- During an emission halt, internal rewards continue accruing but are not paid from existing reserves.
- After emission resumes, deferred internal rewards remain collectible.
- External incentive behavior is unchanged.

## Verification
- `make test PKG=gno.land/r/gnoswap/staker/v1 RUN=TestCollectReward_EmissionHaltDefersInternalReward`
- `make test PKG=gno.land/r/gnoswap/staker/v1 RUN=TestCollect`

## Compatibility
- No public function signatures or event payloads change.
- No companion backend, indexer, database, or frontend changes are required.

## Deployment Notes
- The updated staker implementation must be deployed for the behavior to take effect. Existing internal reward cursors remain compatible.

## Summary by CodeRabbit

- **New Features**
  - External incentive rewards can now be calculated across all applicable incentives for a position.

- **Bug Fixes**
  - During an emission halt, external incentives remain collectible while internal GNS rewards are deferred.
  - Emission-only collection performs no action until emissions resume.
  - Deferred internal rewards remain available after the halt is cleared.
  - Positions can be unstaked during an emission halt without losing deferred rewards.

- **Documentation**
  - Clarified reward collection behavior and emission-status handling during halts.


notJoon (approved):

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -817,9 +832,12 @@ func (s *stakerV1) CollectReward(_ int, rlm realm, positionId uint64) (string, s
...
 	}
 
-	internalRewardToUser, internalRewardPenalty := s.deliverEmissionReward(0, rlm, ctx, reward)
-
-	s.markEmissionCollected(0, rlm, target)
+	internalRewardToUser := int64(0)
+	internalRewardPenalty := int64(0)
+	if emissionActive {
+		internalRewardToUser, internalRewardPenalty = s.deliverEmissionReward(0, rlm, ctx, reward)
+		s.markEmissionCollected(0, rlm, target)
```
jinoosss, line 868: **Restaking is locked for the whole halt when the position was unstaked during it.**

Flow under an emission-only halt (`SetOperationStatus(OpTypeEmission, true)`, staker/withdraw open):

1. `UnStakeToken` records the exit checkpoint with `emissionCollected = false`.
2. `CollectReward` on the checkpoint settles externals and marks them, but `markEmissionCollected` is skipped here because it sits inside `if emissionActive`.
3. `FullyCollected()` never becomes true, the checkpoint is never dropped, and `StakeToken` panics at `assertHasNoExitCheckpoint` with "collect it first" even though collecting cannot clear it.

This also hits positions in tier-0 (external-only) pools whose internal reward is exactly zero: no GNS is owed, yet the checkpoint still needs the emission source marked.

Suggestion: for a checkpoint target, still run the pure internal calculation while halted (it does not mint or transfer), and if `reward.Internal == 0 && reward.InternalPenalty == 0` call `markEmissionCollected` so the checkpoint can be cleared. Tiered pools would remain deferred until resume, which is unavoidable, but zero-owed checkpoints should not block restaking.

jinoosss (commented):
Side-effect review of the emission-halt deferral. The core accounting is sound: `gns.MintGns` catches up from `lastMintedTimestamp+1`, and `ChangeDistributionPct` is blocked while emission is halted, so deferred internal rewards are fully backed once emission resumes. Three follow-ups inline.

docs/staker.md
```diff
@@ -47,6 +47,11 @@ provided.
 - A checkpoint delivery refuses when its cursor already reached the window's close. Collect computes amounts once and then transfers per source, and checkpoint collects are permissionless, so a re-entering call between transfers must not get a precomputed amount replayed.
 - **Invariant the deferred collect depends on**: every path that changes the tier layout must materialize the reward cache of all tiered pools first, at the current time (`changeTier` does this via `cacheReward`). A tier change that skips it would let `resolveInternalRewardSegments` re-rate a checkpoint's closed window with the new layout.
 
+### Emission Halt
+- `MintAndDistributeGns` returning `active = false` defers internal reward calculation, cache updates, checkpoint settlement, and GNS transfers.
+- Internal reward cursors remain unchanged during the halt, so the accrued interval is payable after emission resumes.
```
jinoosss, line 52: Please also document the restaking consequence: a position unstaked during an emission halt keeps its exit checkpoint until emission resumes, so `StakeToken` on that position fails until a post-resume collect clears the emission source. `EndExternalIncentive` is not affected (it only checks the pending-incentive count, which is still decremented while halted).

It would also help to make the `assertHasNoExitCheckpoint` error message hint at this, since "collect it first" is misleading while collecting cannot clear the checkpoint.

contract/r/gnoswap/staker/v1/staker.gno
```diff
@@ -781,19 +787,28 @@ func (s *stakerV1) CollectReward(_ int, rlm realm, positionId uint64) (string, s
...
+	if emissionActive {
+		var err error
+		rewards, rewardUpdate, err = s.calculatePositionReward(rewardParam)
+		if err != nil {
+			panic(err)
+		}
+		s.updatePositionReward(rewardParam, rewardUpdate)
+	} else {
+		rewards, rewardUpdate = s.calculateExternalPositionRewards(rewardParam)
+		s.updateExternalPositionReward(rewardParam, rewardUpdate)
```
jinoosss, line 836: **`ctx.persistDeposit(s)` is no longer reached on the halted path.**

Before this change the deposit write-back happened inside `deliverEmissionReward`. In the `!emissionActive` branch that call is skipped, and nothing after the external delivery loop persists the deposit, while `CollectExternalIncentiveReward` still persists explicitly. Pointer mutations on the tree-held `*sr.Deposit` most likely survive anyway, but the two entry points are now inconsistent and the new tests collect externals only once, so the cursor persistence is not exercised.

Suggestions:
- add `ctx.persistDeposit(s)` after the external loop when `!emissionActive`.
- add a filetest: halt -> `UnStakeToken` -> `CollectReward` -> `StakeToken` (expect failure) -> resume -> `CollectReward` -> `StakeToken` (expect success).
- add a filetest that collects external rewards twice in a row during the halt and asserts the second collect pays nothing.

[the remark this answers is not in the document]
jinoosss: Declining this suggestion. Three points:

1. **No such repository rule exists.** Neither `AGENTS.md`, `docs/`, nor any CodeRabbit configuration in this repo requires an event for every state change. Events here are emitted for user-visible or accounting-relevant transitions (stake, unstake, collect, incentive lifecycle), not for internal bookkeeping.

2. **This is not a gap introduced by the halt path.** `applyExternalIncentiveIndex` has always run on the normal `CollectReward` and `CollectExternalIncentiveReward` paths without an event when no reward was delivered. The `!emissionActive` branch calls the same function with the same behavior.

3. **The mutated state is a scan cache, not accounting state.** `newExternalIncentiveIds` and `LastExternalIncentiveUpdatedAt` only record which incentives have already been scanned. `ExternalRewardLastCollectTime` falls back to `StakeTime` for ids not yet persisted, so the reward calculation is identical whether or not the index is written (see the comment on `discoverExternalIncentives`). The cursor that affects amounts is the per-incentive `lastCollectTime`, which advances only on delivery and is already covered by the per-incentive `CollectReward` event. An indexer reconstructs incentive membership from `StakeToken` + `CreateExternalIncentive`/`EndExternalIncentive` and cursors from the delivery events.

Adding `UpdateExternalIncentiveIndex` would fire on nearly every collect (the cursor advances whenever at least one second has passed), mostly with an empty `newIncentiveIds`, and forces `cur realm` into functions that were pure state appliers. The event will be removed, keeping the PR's "no event payload changes" compatibility claim accurate.

jinoosss (approved):

PR #1473, opened 2026-09-15
fix(governance): allow halt-only proposals while governance is halted for self-recovery

jinoosss:
## Summary

Once the `governance` operation is halted (e.g. `COMPLETE`), governance could not recover on its own because `ProposeParameterChange` / `Vote` / `Execute` were all blocked by the governance halt guard, so only admin could unhalt the system. This PR lets governance propose, vote on, and execute halt-only proposals while halted.

Admin authority over halt levels is unchanged: admin or governance can still set any level and toggle any operation, including `withdraw`.

### `r/gnoswap/gov/governance/v1`
- Added a halt-recovery exemption: a `ParameterChange` proposal whose executions **all** target `gno.land/r/gnoswap/halt/v1` can be proposed, voted on, and executed while the `governance` op is halted.
- Applied to `ProposeParameterChange`, `Vote`, and `Execute`. `ProposeText`, `ProposeCommunityPoolSpend`, `Cancel`, `RemoveInactiveProposalFromIndex`, and `Reconfigure` keep the existing halt guard.
- Mixed proposals (halt + any other target), text proposals, and nonexistent proposal IDs are still blocked while governance is halted.
- Voting period, quorum, and timelock rules apply unchanged to recovery proposals.

### `r/gnoswap/halt/v1`
- README: note that governance can recover from `COMPLETE` on its own via halt-only proposals.

### Unit (`gov/governance/v1`)
- Propose: halt-only proposal allowed under `COMPLETE`; mixed / non-halt blocked.
- Vote: halt-only (single and multi-message) votable under `COMPLETE`; mixed / non-halt / nonexistent blocked.
- Execute: round trip that unhalts governance and withdrawals; multi-message halt-only; `COMPLETE` → `EMERGENCY`; mixed / nonexistent / text blocked; timelock still enforced; `Cancel` stays blocked.

### Scenario (`r/scenario/halt`)
- `complete_mode_governance_self_recovery_filetest.gno`: governance sets `COMPLETE`, text / non-halt / mixed proposals are blocked, a halt-only recovery proposal is proposed → voted → executed under `COMPLETE`, the system returns to `NONE`, and withdrawals and ordinary proposals work again.

## Verification

`gno test` (against `gnoswap-labs/gno` master) passes for `halt/v1`, `gov/governance/v1`, and `scenario/halt`.

## Notes

- The PR history contains an intermediate commit that restricted `SAFE_MODE` / `COMPLETE` / `withdraw` halts to governance and a follow-up revert. The net diff against `main` does not change admin authority.


contract/r/gnoswap/gov/governance/v1/governance_vote.gno
```diff
@@ -42,7 +41,7 @@ import (
 func (gv *governanceV1) Vote(_ int, rlm realm, proposalID int64, yes bool) string {
 	access.AssertIsRlmCurrent(0, rlm)
 
-	halt.AssertIsNotHaltedGovernance()
+	assertIsNotHaltedGovernanceForProposal(gv, proposalID)
```
junghoon-vans, line 44: Please add a success case that calls `Vote` for a halt-only parameter-change proposal while governance is halted. This branch is required to recover from `COMPLETE`, but the current halted-state case only verifies that a text proposal is rejected. Please assert that the vote succeeds and records the voter’s weight, while retaining the existing non-halt rejection case.

junghoon-vans (approved):

notJoon (approved):

